PR #1, opened 2023-08-13
Extended GroundItem.java to allow interactions 

Jrod7938:


Jrod7938:
example GroundItem.interact(Item: "box trap", Action: "lay", Range: 7)

PR #29, opened 2023-09-05
Paydirt script and some pathfinding stuff

Sunderw3k:
Need to test if teleports still work in pathfinding. I dont think there was a check to not teleport if useTeleport was false so I added that. Overall pathfinding was the biggest change even tho i didnt touch the logic.


chsami:
Amen

CzajnikOfficial:
Amen

PR #32, opened 2023-09-08
Feat: Sand Miner plugin

Kstar-rs:
Mines sand at the desert quarry, deposits in the grinder.
Does not support equipment management or waterskins in current version.
Users are expected to be ready at the quarry.

Motivation for this script: need to farm 50k buckets on my ironman 😵

image


chsami:
Thanks for the contribution

PR #41, opened 2023-09-11
Extending crafting script

Kstar-rs:
Added new glassblowing script
Moved scripts to sub dir to keep root package tidy, expandable in future
Added Activity enum to expand on later
Updated config to be able to select Activity
Updated config to select "Random AFK" (so far only used in glassblowing script)
Each script now has its own config section in the plugin, expandable in future


chsami:
Thanks for the contribution <3

PR #73, opened 2023-09-19
add ability to define how far to stray from starting position

jkryspin:


chsami:
You should write a method for it in Rs2GameObject

We already have a method called 

`interact(int id, String action, int distance)`

You should create a new one with a string name instead of int id.

Something like this:

`interact(String name, String action, int distance)`

This will allow us to do the same in other scripts.

Thanks!

jkryspin:
@chsami It doesn't really make sense to use distance here, since this change prevents straying from an origin location, not from the users current position.

I did it this way because when chopping trees with a low distance the player would stray far away from the origin where the script is started. I could pass in WorldLocation aka the origin and distance but that seems like too much logic to expose via a shared method

chsami:
> @chsami It doesn't really make sense to use distance here, since this change prevents straying from an origin location, not from the users current position.
> 
> I did it this way because when chopping trees with a low distance the player would stray far away from the origin where the script is started. I could pass in WorldLocation aka the origin and distance but that seems like too much logic to expose via a shared method

Oh okay that make sense

PR #78, opened 2023-09-20
External plugin fix

Sunderw3k:
Damn you Adam. That's fair tho


chsami:
Already fixed in development

PR #82, opened 2023-09-20
Improvements to Fishing Script

jkryspin:
Classify Lobster Locations as Harpoonable (which the are, for example Karamja spot 1522)
Also the logic was a bit wonky before -- passing an id which could result in a null NPC to interact, now that is explicitly checked


chsami:
Please make a PR into development, we don't do PR's directly into main anymore

PR #97, opened 2023-09-24
Deprecate Inventory in favor of Rs2Inventory

GriffinBoris:


chsami:
I'll be rewriting the Inventory class itself, so this will be redudant.

PR #116, opened 2023-10-03
Griffin/auto pilot

GriffinBoris:


runelite-client/src/main/java/net/runelite/client/plugins/griffinplugins/util/helpers/WorldHelper.kt
```diff
@@ -0,0 +1,75 @@
...
+
+            val worldsToHopTo = worlds.worlds.filterNotNull()
+            if (worldsToHopTo.isEmpty()) {
+                return false
+            }
+
+            val worldIds: MutableList<Int> = mutableListOf()
+
+            if (!isMembers) {
+                worldIds.addAll(worldsToHopTo.filter { !it.types.contains(WorldType.MEMBERS) }.map { it.id })
```
chsami: We have a method in Login.java that gets you a random world. getRandomWorld(boolean isMembers)

It's better to use that as it filters, total level worlds, pvp and some other stuff

chsami (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/griffinplugins/util/helpers/WorldHelper.kt
GriffinBoris: Oh awesome! Thank you sir I will make that change!

chsami (approved):

PR #160, opened 2023-11-08
fix for get item

GriffinBoris:


DetectiveCode (approved):
Looks good. Just curious though, what's the result of running code on the client thread?

PR #172, opened 2024-02-28
Add knife_log method. Also add cut-eat for any 3t method

ckilmer:
Update to Dan's plugin. Allows for additional option to give free cooking xp. Also added Knife log 3t method as well.


chsami:
Looks good to me! Thanks for your contribution!

PR #194, opened 2024-04-22
change: prayer level to active rigour - Vorkath

gabrielteodoroo:


gabrielteodoroo:
rigour require 74 prayer level to active

PR #195, opened 2024-04-23
BanksBankStander

Bankjs:
This pull request introduces the BanksBankStander script, designed to automate simple bank skilling activities in Old School RuneScape. The script allows players to efficiently perform tasks such as fetching items from the bank and combining them, enhancing the skilling experience.

Changes:

Added the BanksBankStander package containing the script implementation.
Implemented the script logic for fetching items from the bank, combining them, and handling sleep durations for synchronization.
Modified the plugin descriptor to include the developer name for attribution.
Benefits:

Offers an automated solution for repetitive bank skilling tasks, saving time and effort for players.
Enhances the skilling experience by streamlining the process of gathering and combining items.
Testing:

Tested the script functionality to ensure correct behavior in various scenarios, including starting with items in the inventory and starting without items.
Verified the synchronization and sleep durations to prevent script errors or inefficiencies.
Additional Notes:

The script is designed to be easily customizable and extensible for future updates or specific user requirements.
Feedback and suggestions for improvements are welcome to further enhance the script's functionality and usability.


chsami:
Script looks neat, goodjob. My only concern is the obscurity of finding the item ids, which might be pretty annoying. Would it be possible to allow ids and names.

Could we change the firstItemId and secondItemId to type string. This allows for names or numbers to be put in the config field.

We could then parse the config field in our script by writing an extra method that checks if the input is a number

```
public static Integer TryParseInt(String someText) {
   try {
      return Integer.parseInt(someText);
   } catch (NumberFormatException ex) {
      return null;
   }
}
// This is just pseudo code
 Integer firstItemId = TryParseInt(config.firstItemId());
  if (firstItemId == null) {
     // FirstItemId is text
      Rs2Inventory.hasItemAmount(config.firstItemId(), config.firstItemIdAmount());
  } else {
      //firstItemId is a number
      Rs2Inventory.hasItemAmount(firstItemId, config.firstItemIdAmount());
  }
```

This allows for names or ids to be put it and more flexibility to the end user.

Bankjs:
Added flexibility to item identification in fetch and combine methods

- Modified the fetchItems() and combineItems() methods to handle cases where item identifiers can be either item IDs or names.
- Implemented logic to check if an item identifier is null and handle it appropriately in both methods.
- Updated the hasItems() method to accommodate the changes made in fetchItems() and combineItems().

This update allows users to input either item IDs or names for both items, providing more flexibility in item identification.

chsami:
Please take out the code for the bankscript plugin. I want to discuss the bank pin plugin with you before implementing this.

We will add that feature in a seperate PR. Please also change the title to only include the bankstanding plugin.

Thanks!

PR #204, opened 2024-05-02
JAD Prayer update

UKMonster:


chsami (commented):
Your plugin got an error when building your PR

 PluginManagerTest.testLoadPlugins:170 expected:<183> but was:<182>

Might be because PrayerPlugin already exists, not sure

PR #206, opened 2024-05-04
Fixed logic and massive performance to GiantsFoundryScript

See1Duck:


chsami:
oh this is tough.... I had just reworked the giants foundry script a day ago ... Thanks for the PR tho.

I'm willing to merge it as a seperate plugin under your name, if you're OK with that. These plugins could live side by side

PR #243, opened 2024-06-01
Added import for util.widget.Rs2Widget

xicez:
Was running into build issues without the import


chsami:
Already done, thanks!

PR #245, opened 2024-06-01
Added Fairy Rings & Gliders. Added Rs2Inventory.useItemOnNpc()

Bankjs:
Rs2Inventory.useItemOnNpc(itemId, NPC)

This additional method allows the user to use an item on a specific NPC rather than a generalised NPC id. This is crucial to be able to use items like Ice Coolers and the Rock Hammer on Lizards and Gargoyles respectively.

The current Method which takes itemId and NpcId as params causes issues because the NPC effectively dies at 0 hitpoints and the result is that the character tries to use the item on a different NPC.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/inventory/Rs2Inventory.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/inventory/Rs2Inventory.java
@@ -446,7 +446,7 @@ public class Rs2Inventory {
     /**
      * Drop all items that fall under the gpValue
      *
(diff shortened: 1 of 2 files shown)


chsami:
Hey @Bankjs, Looks really good. 

My only concern was that the Rs2Walker will become a giant class with all the logic for teleports and other utility stuff.

Is is possible to move the logic for the glider and walker to the Transport.java class?

Your code would be something like:

```
if (b.isGnomeGlider()) 
{
    b.useGnomeGlider();
}
```

This would keep the logic in the transport class if we ever would need this somewhere else.

Bankjs:
Closing due to issues

PR #253, opened 2024-06-04
BankShopper supports multiple items + logout when out of supplies/coi…

chsami:
1. BankShopper supports allows multiple items to be sold/bought;  These items are comma seperated in the itemName(s) config.
2. A new config has been added to allow the player to logout when out of supplies.
3. When buying the Script listens to "you don't have enough coins." to detect out of supplies. If not it will stop the script and check the config logout to logout the player.
4. When selling items the script checks if you have the items in your inventory. If not it will stop the script and check the config logout to logout the player.
5. Aded an extra check for the Rs2Shop.isOpen() to make sure the interface is visible.
6. Selling items will now check the minimum amount config item in the shop to detect when to hop.
7. Script now supports banking when buying items.
8. When buying items the script will track the gp of bought items when you bank.
9. Bumped version to "1.2.0"
10. Added an overlay to display the amount of gp for the bought items


Bankjs (approved):
Looks good to me.

Changes are valid and offer good utility.

PR #257, opened 2024-06-09
added isInArea(WorldPoints) and isInArea(WorldPoint, int). 

ObserveObserver:
will check if players in specified area. added 2, one for two world points (NW and SE corners) and one for the center area + range.


chsami:
Oh nice method! Thanks!

PR #258, opened 2024-06-11
Updated Woodcutting Script & Add Methods to RS2Player (#1)

gmason0:
I have created an update to the Woodcutting script with the following features:
* Added new options when Inventory is full, Banking or Firemake
* Adjusted Dropping logic to only drop the specific log for the tree you are cutting to prevent other items in your inventory from being dropped
* Added configuration options for where to walk back when banking

During the time making this, I have also made some new utility methods inside of RS2Player:
* isStandingOnGameObject() - returns boolean if player is standing on GameObject
* isStandingOnGroundItem() - returns boolean if player is standing on GroundItem
* getAnimation() - returns int for current animationid
* getPoseAnimation() - returns in for current poseanimationid

Let me know your thoughts!


chsami:
I'll probably change some of the code to banking and walking back, as i made a general method to bank & walk back to original location, as this is used in many scripts.

The rest looks good to me.

PR #267, opened 2024-06-12
Rs2Inventory.useItemOnNpc(itemId, NPC)

Bankjs:
Rs2Inventory.useItemOnNpc(itemId, NPC)

This additional method allows the user to use an item on a specific NPC rather than a generalised NPC id. This is crucial to be able to use items like Ice Coolers and the Rock Hammer on Lizards and Gargoyles respectively.

The current Method which takes itemId and NpcId as params causes issues because the NPC effectively dies at 0 hitpoints and the result is that the character tries to use the item on a different NPC.

Since NPC are unique, this is necessary.


chsami:
It seems that your commit history is messed up, because many of the changes are already in the main branch.

PR #279, opened 2024-06-16
Implemented Lunar Tanner in Microbot

feveromo:
- Added functionality for tanning various types of dragonhides using the Lunar Spellbook.
- Handles banking, withdraw/deposit hides, cast tan leather, and calculating profit.
- This plugin generates a lot of gp and xp, so be careful not to run it too long.


chsami:
Please makr a PR to development. This way we can test the script before merging it to the main client

PR #283, opened 2024-06-18
Fix AutoWoodcuttingScript & other API changes

gmason0:
I have implemented a fix & tested the AutoWoodcuttingScript in a few different spot. Between the hasLineOfSight check & this added reset boolean, things seem to work properly with all three reset options.


runelite-client/src/main/java/net/runelite/client/plugins/microbot/woodcutting/AutoWoodcuttingScript.java
```diff
@@ -136,11 +142,11 @@ private boolean isFiremake() {
     private void walkBack(AutoWoodcuttingConfig config) {
         if (config.walkBack().equals(WoodcuttingWalkBack.INITIAL_LOCATION)) {
             Rs2Walker.walkTo(new WorldPoint(initialPlayerLocation.getX() - Random.random(-1, 1), initialPlayerLocation.getY() - Random.random(-1, 1), initialPlayerLocation.getPlane()));
-            sleepUntil(() -> Rs2Player.getWorldLocation().distanceTo(initialPlayerLocation) <= 4);
+            sleepUntilOnClientThread(() -> Rs2Player.getWorldLocation().distanceTo(initialPlayerLocation) <= 4);
```
chsami: Why are you sleeping on the client thread?

runelite-client/src/main/java/net/runelite/client/plugins/microbot/woodcutting/AutoWoodcuttingScript.java
```diff
@@ -54,32 +56,35 @@ public boolean run(AutoWoodcuttingConfig config) {
...
+                        isResetting = false;
                         break;
                     case BANK:
                         List<String> itemNames = Arrays.stream(config.itemsToBank().split(",")).map(String::toLowerCase).collect(Collectors.toList());
 
-                        if (Rs2Inventory.isFull()) {
-                            if (!Rs2Bank.bankItemsAndWalkBackToOriginalPosition(itemNames, initialPlayerLocation))
-                                return;
-                        }
+                        if(!Rs2Bank.walkToAndBankItems(itemNames))
```
chsami: What was the issue with bankItemsAndWalkBackToOriginalPosition?

runelite-client/src/main/java/net/runelite/client/plugins/microbot/woodcutting/AutoWoodcuttingScript.java
chsami, line 1: @g-mason0 We should create a getter for returnPoint that has a config as a parameter and does the check

private static WorldPoint calculateReturnPoint(AutoWoodcuttingConfig config) {
    if (config.walkBack().equals(WoodcuttingWalkBack.LAST_LOCATION)) {
        return returnPoint
    } else {
        return initialLocation;
    }
}

We can then remove the if statement in walkback and the resetInventory method and replace it with the method.

Let me know if you need more information.

chsami (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/woodcutting/AutoWoodcuttingScript.java
gmason0: I have reverted these changes & went back to use bankItemsAndWalkBackToOriginalPosition

runelite-client/src/main/java/net/runelite/client/plugins/microbot/woodcutting/AutoWoodcuttingScript.java
gmason0: Not sure why I changed this, I have swapped it back.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/woodcutting/AutoWoodcuttingScript.java
gmason0: This has been implemented & tested in commit a094135

chsami (approved):

PR #292, opened 2024-06-22
Chompy

belsan:
Same pull-request as before, just now for the development branch.


runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/inventory/Rs2Inventory.java
```diff
@@ -1622,9 +1622,13 @@ private static void invokeMenu(Rs2Item rs2Item, String action) {
                     break;
                 }
             }
-            if((Objects.equals(actions[1], "Wear") && actions[0]==null) || action.equalsIgnoreCase("drop") || action.equalsIgnoreCase("empty") || action.equalsIgnoreCase("check")){
+            if ((Objects.equals(actions[1], "Wear") && actions[0] == null) || action.equalsIgnoreCase("drop") || action.equalsIgnoreCase("empty") || action.equalsIgnoreCase("check")) {
                 identifier++;
             }
+            if (action.equalsIgnoreCase("drop") && rs2Item.id == 2875){ // bloated toad
```
chsami: can you try and merge the latest main into your branch? This has a generic way of handling inventory actions, meaning that this code would not be necessary. You should test your script with the new generic invokeMenu. Let me know if it works,.

belsan: It does not work, the identifier is correct, but somehow the item ID gets set to -1:
MenuEntryImpl(getOption=Use, getTarget=Bloated toad, getIdentifier=2, getType=CC_OP, getParam0=7, getParam1=9764864, getItemId=-1, isForceLeftClick=true, getWorldViewId=-1, isDeprioritized=false)

chsami: Make sure to clean your .m2 folder that contains maven packages.  clean maven package and run install again.

The reason I say this, is that there was an update in the runelite client-patch which comes from the maven repository. This used an oudated version for some people. Using the latest should fill in the itemId.

chsami (approved):

PR #297, opened 2024-06-25
Utility Class Changes and New Scripts

gmason0:
I have been cooking on this one for a few days now & I wanted to get these changes into upstream before more changes are done to the utility classes to prevent merge conflicts.

Utility Class Changes:

1. Rs2Walker Changes - I added the 'Pick-Lock' menu option in-order to get into some of the locations where the Nature Rune Chests are located.
2. Rs2GameObject Changes  - I added a similar method of findObject but using the ObjectID instead of a string, again Nature rune chests are labeled as 'chest'.
3. Rs2Bank Changes - I was running into issues with my inventory not being completely full & the script going to bank. I have adjusted this to rely off of empty slots instead of isInventoryFull, there is a helper method that will do 0 empty slots to accommodate for previous integrations. 
4. Rs2NPC Changes - I have implemented a method to find the NPC nearest to the player.

Plugin & Script additions: 
1. I have created a script to spin flax at a few preset spinning wheel locations, I am having an issue with the Lumbridge Castle location. I believe this is related to the Rs2Bank logic with using distanceTo2D. I tried changing things to use distanceTo & it was trying to take me somewhere completely different. Not sure where to start with this... So I left it out for now.
2. I created an auto looter plugin with a few different functions. This can pick flax at a few locations, to supplement the Flax Spinning script, it can loot nature rune chests & it also has the functions from the AIO fighter to pick up items based off a comma-separated list or GE Price.
3. I have re-worked the event dismiss plugin to check which random event NPC spawns & allow you to talk to them. This could be useful if you would like to grab the XP lamps from the Genie & count check, but dismiss others.

Pottery Plugin is still in the works, I need to finalize one more location before getting this merged upstream.


gmason0:
Just to add, I am having three weird issues with the walker in this.

1. For the Flax Spinning Script under the Crafting Plugin, when using the Lumbridge location, it can't find the bank because it is right above the player, i have just commented it out for now.
2. For the Nature Rune Chest Script, when using the East ardougne location, it can't seem to get into the room. I tried adding the transports to the .tsv file which seems to actually get the pathing to go into the room, but it wants to click going up the stairs before going through the door.
3. For the Flax picking script, I am experiencing an issue where using the Lands End location & it goes to return to bank, it can't open the chest nearby, the scripts starts returning null when calling Rs2Bank.openBank()

Any pointers on this would be greatly appericated!

gmason0:
[Video of Walker issue with East Ardougne Nature Rune Chest](https://azpublicfiles.blob.core.windows.net/uploads/java_dvCKOSncYw.mp4)

Is there some sort of priority or order I need to do to make sure it does the door first then the stairs?

gmason0:
I have added a few changes to the ShortestPathPlugin & Rs2Walker to use a local variable called "reachedDistance" to ensure that "reachedDistance" is respected when using walkTo(worldPoint, distance).

I also added another fix mentioned but another member in the discord to fix interacting with stairs or other objects.

chsami:
Interaction with staircase and bank should be fixed. So please remove any code you ajdusted for that. It makes the PR a bit easier to review

gmason0:
I have removed my modifications to the walker to make the staircases to work & also my adjustments to reached distance. 

I can confirm the issue with Rs2Walker.WalkTo(WorldPoint, distance) not working properly is related to the onGameTickEvent [here](https://github.com/chsami/microbot/blob/main/runelite-client/src/main/java/net/runelite/client/plugins/microbot/shortestpath/ShortestPathPlugin.java#L253), I am just not sure the best way to implement a fix to pass the distance from Rs2Walker utility class to the ShortestPathPlugin.

In turn of removing changes, I have added some more 😊

1. I added utility function for SuperHeat, I will be using this in my next plugin - Auto Smelting ( may also add another for enchanting jewelry )
2. I added a utility method into Rs2Settings for disabling level up notifications, similar to the toggle roofs.
3. I have put my pottery plugin in here - it is mostly complete, I haven't had time to finish up the humidify spell function, so I have hidden the option to use it at the moment. It works in its current state all the locations.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/npc/Rs2Npc.java
```diff
@@ -154,6 +155,16 @@ public static Optional<NPC> getNpc(int id, List<Integer> excludedIndexes) {
                         value.getLocalLocation().distanceTo(Microbot.getClient().getLocalPlayer().getLocalLocation())));
     }
 
+    public static Optional<NPC> getNPCNearestToPlayer() {
```
chsami: Does the following method not qualify?

`    public static Stream<NPC> getNpcs() {
        Stream<NPC> npcs = Microbot.getClientThread().runOnClientThread(() -> Microbot.getClient().getNpcs().stream()
                .filter(x -> x != null && x.getName() != null && !x.isDead())
                .sorted(Comparator.comparingInt(value -> value.getLocalLocation()
                        .distanceTo(Microbot.getClient().getLocalPlayer().getLocalLocation()))));`

chsami: The getNpcs() also sorts from closest npc to further away. So if you get the first element of the stream, should always have the closest npc

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/walker/Rs2Walker.java
```diff
@@ -296,11 +296,19 @@ private static boolean handleDoors(WorldPoint currentWorldPoint, WorldPoint next
         if (wallObject != null) {
             ObjectComposition objectComposition = Rs2GameObject.getObjectComposition(wallObject.getId());
 
-            for (var action : objectComposition.getActions()) {
-                if (action != null && action.contains("Pay-toll")) {
-                    Rs2GameObject.interact(wallObject, action);
-                    Rs2Player.waitForWalking();
-                    return true;
+            for (var action : objectComposition.getActions()){
+                if(action != null){
```
chsami: if (action != null && (action.contains("Pay-toll") || action.contains("Pick-lock"))

Seems cleaner and you will avoid duplicate code

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/npc/Rs2Npc.java
gmason0: I should stop working on this stuff so late. I have simplified the logic for this in commit c6202bb

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/walker/Rs2Walker.java
gmason0: I have also simplified the logic here on commit d9572de

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/settings/Rs2Settings.java
```diff
@@ -49,6 +50,38 @@ public static void hideRoofs() {
...
+    public static boolean disableLevelUpNotifications() {
+        if(isLevelUpNotificationsEnabled()){
+            Rs2Tab.switchToSettingsTab();
+            Rs2Widget.clickWidget(7602208);
+            final boolean isSettingsInterfaceVisible = Rs2Widget.getWidget(8781825) != null;
+            sleepUntilOnClientThread(() -> isSettingsInterfaceVisible);
+            if(isSettingsInterfaceVisible){
+                Rs2Widget.clickWidget(8781834);
+                sleep(600);
+                Rs2Keyboard.keyPress('l');
```
chsami: You can use Rs2Keyboard.typeString to type it in one word

chsami (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/settings/Rs2Settings.java
gmason0: should be good now in commit a005a47

chsami (approved):

PR #307, opened 2024-06-28
Improvements for MQuestScript and different helper changes

Alexander1609:
Changes for MQuestScript:
* Re-added the click selection 1 part, if the choices from questhelper are not complete or the "Yes" to start the quest is not included
* Added sleeps to prevent spamming
* Added support for multi-NPC steps
* Added support for using items on NPCs if the widget is displayed by the questhelper
* Rs2Walker target will now be cleared if an Object/NPC is clicked to prevent the walker from aborting the action
* Added support for ground objects
* Added object action validation when interacting with objects that have multiple actions available, as the questhelper step description will most likely contain the correct action to be used 

Different helper changes:
* WorldPointUtil: Fix or the distanceBetween function to respect planes (hard distance of 1000 to make the pathfinder not ignore the tile if it would just return max value), added isPointInPolygon function
* Rs2Walker: Added walkTo and walkMiniMap functions for areas (choosing a random point within), added walk-through action for some tents in the desert, removed hasLineOfSight check for wall- and ground-objects as some transport never pass the line-of-sight check
* Rs2Tile: Fixed the isWalkable function (to not block tiles that are walkable and just have the n/s/w/e flag), added a getWalkableTiles function independent of the player position
* Rs2Npc: Added function to get all npcs by an id
* Rs2Inventory: Fixed the containsAll function to actually check all passed ids, added confirmation for destroy dialogs to the invokeMenu function
* Rs2Widget: Bugfix for the isWidgetVisible function if the widget is not present at all
* Rs2Dialogue: Added multiple widgets to isInDialogue check, and added a function to check if the continue widget is displayed
* Rs2Camera: Bugfix to not use the right key when lowering the pitch
* Rs2Bank: Bugfix for walkToBankAndUseBank as 2d distance calculation leads to the walker stopping on a lower/higher plane
* Rs2GameObject: Extension for clickObject function to use the first not-null action if passing no specific action or the specified action is not found (for some objects the default index would be 1 or 2), added getTileObjects to get all non-gameobject objects by id similar to getGameObjects
* QuestHelper: Added some getters, small fixes in quests to be better automated by quest plugin
* Puzzlesolver: Made VarrockMuseumAnswer accessable from outside the puzzlesolver plugin
* Added quest related transports, and made the waterfall quest transports usable during waterfall quest

### DIFF
--- a/runelite-client/src/main/resources/net/runelite/client/plugins/microbot/shortestpath/transports.tsv
+++ b/runelite-client/src/main/resources/net/runelite/client/plugins/microbot/shortestpath/transports.tsv
@@ -1338,10 +1338,10 @@
 2557 9844 0	2557 3444 0	Climb-up Ladder 17387				2	
 							
 # Tree Gnome Village Dungeon (during Waterfall quest)							
-2533 3156 0	2533 9556 0	Climb-down Ladder 5250			Waterfall Quest		
-2534 3155 0	2534 9555 0	Climb-down Ladder 5250			Waterfall Quest		
-2532 3155 0	2532 9555 0	Climb-down Ladder 5250			Waterfall Quest		
-2533 9556 0	2533 3156 0	Climb-up Ladder 17387			Waterfall Quest		
+2533 3156 0	2533 9556 0	Climb-down Ladder 5250
+2534 3155 0	2534 9555 0	Climb-down Ladder 5250
+2532 3155 0	2532 9555 0	Climb-down Ladder 5250
+2533 9556 0	2533 3156 0	Climb-up Ladder 17387
 							
 # Tree Gnome Village Dungeon (after Waterfall quest)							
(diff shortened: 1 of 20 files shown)


runelite-client/src/main/resources/net/runelite/client/plugins/microbot/shortestpath/transports.tsv
```diff
@@ -1338,10 +1338,10 @@
 2557 9844 0	2557 3444 0	Climb-up Ladder 17387				2	
 							
 # Tree Gnome Village Dungeon (during Waterfall quest)							
-2533 3156 0	2533 9556 0	Climb-down Ladder 5250			Waterfall Quest
```
chsami, line 1341: why did you remove the waterfall quest?

chsami (approved):

PR #315, opened 2024-07-04
Update for ShortestPathPlugin

Alexander1609:
Updated the ShortestPathPlugin to the newest version, supporting item teleports.
* Added an option to set the distance increase with consumable teleports (default 200 tiles) to prevent unnessecary usage
* Tested so far: Amulets (Glory, Games), tablets and Ardy cape
* Other items would need more testing


Alexander1609:
FIXED Ive noticed an issue with multi-step agility transports (stepping stones in particular) coming from this update.. Will take a look at this.

bttq-dll:
@Alexander1609 could you add support for lumbridge basement for nearest bank?

Alexander1609:
> @Alexander1609 could you add support for lumbridge basement for nearest bank?

@bttqjs Thats not really related to this PR. Iirc the lumb basement chest is already added as a bank location, but the problem is the 2D distance calculation in getNearestBank leads to it not being used.

But yea, especially with the Item teleport i added here i already thought of using the pathfinder to choose the best bank. Sadly, this already takes 10+ seconds to calculate on a good computer..

In my opinion there are two good ways to handle this: Either we come up with a better approximation for the distance or use some sort of precalculated map. The second option would have the downside to require remapping If a new bank is added.

chsami:
I'll close this PR as I included most of these changes in the new-webwalker branch.

PR #317, opened 2024-07-06
Rs2Inventory:

See1Duck:
ADD: CombineClosest
REMOVE: removed various redundant find functions

Normal Combine Method
Combine normal

Combine Closest method
Combine closest (1)

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/inventory/Rs2Inventory.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/inventory/Rs2Inventory.java
@@ -1835,6 +1835,7 @@ public class Rs2Inventory {
         }
 
         if (!action.isEmpty()) {
+            assert inventoryWidgets != null;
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/inventory/Rs2Inventory.java
```diff
@@ -1289,13 +1371,13 @@ public static boolean isSlotFull(int slot) {
      * @param rs2Item The item to get the bounds for.
      * @return The bounding rectangle for the item's slot, or null if the item is not found.
      */
-    public static java.awt.Rectangle itemBounds(Rs2Item rs2Item) {
+    public static Rectangle itemBounds(Rs2Item rs2Item) {
         Widget inventory = getInventory();
 
         if (inventory == null) return null;
```
chsami, line 1377: What happens when null is returned, will the invoke crash?

See1Duck: Shouldn't theoretically be possible based on the other checks before, but i pushed changes to handle null values just in case

chsami (approved):

PR #320, opened 2024-07-08
New plugin: Mage Training Arena

Alexander1609:
Adds a new plugin to play all minigames in the MTA fully automated.

- 2 Modes: Buying rewards / infinite
- Moves to the next room if the point goal is reached (if not buying items it will still set the reward price as a goal and repeat)
- Adaptive spell choice
   - Uses Bones to Peaches over Bones to Bananas if unlocked
   - Uses the highest level enchantment spell unlocked (except for T7 as it is not worth)
- Supports multi purchase rewards (higher level wands)
- Stave configuration (mud staff and one staff for air/earth/fire is required)
- Healing with created fruits
- Estimations: 70-80k xp/h and 500k+ gp/h (depending on reward selected)

Requirements:
- 33 Magic & 30 Attack
- Mud + elemental staves
- Enough runes (https://oldschool.runescape.wiki/w/Mage_Training_Arena#Rewards)

Minor changes:
* Added support for the "new" enchant window in Rs2Magic (similar change is probably required for bolt enchants)
* Changed the sprite ids for all enchantment spells


runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/magic/Rs2Magic.java
```diff
@@ -29,6 +31,18 @@ public static boolean canCast(MagicAction magicSpell) {
             Rs2Tab.switchToMagicTab();
             sleep(150, 300);
         }
+
+        if (magicSpell.getName().toLowerCase().contains("enchant")){
+            if (Rs2Widget.findWidget(magicSpell.getName(), Arrays.stream(Rs2Widget.getWidget(218, 0).getStaticChildren()).collect(Collectors.toList())) == null) {
```
chsami, line 36: Why was this custom code needed?

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/magic/Rs2Magic.java
```diff
@@ -143,6 +156,10 @@ public static void superHeat(Rs2Item item, int sleepMin, int sleepMax) {
             sleep(50, 150);
             return Rs2Tab.getCurrentTab() == InterfaceTab.MAGIC;
         });
+        if (Rs2Widget.isWidgetVisible(218, 4) && Arrays.stream(Rs2Widget.getWidget(218, 4).getActions()).anyMatch(x -> x.equalsIgnoreCase("back"))){
```
chsami, line 159: Is the MTA a custom interface for casting spells?

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/magic/Rs2Magic.java
Alexander1609: Because the enchantment spells are grouped within a new UI (even outside of the MTA): 
image
image

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/magic/Rs2Magic.java
Alexander1609: And similar to the code above to switch to the enchantment UI, this part switches back (because it stays open after casting an echant, and would block other spell casts)

PR #325, opened 2024-07-09
Update Thieving to 1.5.6

zebdo:
added config options
added a few checks
added banking for shadow veil reqs


zebdo:
ooo maybe that canCast edit is a bit goofy, whole Rs2Magic class needs a bit of a clean up hella wonky

PR #333, opened 2024-07-13
Devlopment

rastinder:
added itemQuantity function that returns exact amount of an item
added check if collect button is present only then it will be pressed


chsami:
Please make a pr into development instead of main

chsami (approved):

PR #354, opened 2024-08-01
Add rogues den plugin

ckilmer:
Tested for short while, looks like about 75% success rate on the course, but script doesn't really get stuck for more than a few seconds so it is self correcting. Most users probably won't be running for more than 90 minutes anyways


See1Duck:
Change the branch from main to development

chsami (approved):

PR #371, opened 2024-08-24
Added AutoEatPlugin

T3rr0or:


yace0:
you should add auto pot to this too

T3rr0or:
What pots should u suggest. I would assume this will be another seperate plugin.

chsami:
You can use AutoFighter with the auto eat option if you want this feature. We want to avoid to many plugins, this takes longer to load the client and makes it harder to find the plugin you need. Thanks for taking the time to make a PR. Feel free to join our discord and have a chat.

I'm also aware that S-1D has a QOL plugin, maybe we can add all those features in there.

PR #377, opened 2024-09-05
ShootingStar Plugin + Implement Antiban Features for plugins

gmason0:
I want to apologize in advance for adjusting various things in this PR, I tried to focus this as best I could but I am just happy to get this merged!

Implemented Antiban Features on the following plugins:
1. Nate's AutoFishing Plugin
2. AutoCooking Plugin
3. AutoLooter Plugin
4. AutoWoodcutting Plugin

ShootingStar Plugin -

[Showcase](https://azpublicfiles.blob.core.windows.net/uploads/java_NrtJdLa6eT.mp4)

The ShootingStar plugin is a star finder/auto star miner plugin. It leverages public data about scouted crashed stars & it will match those locations with some static game static & some checks around the player to confirm build a panel that the user can select stars from, walk to & mine them.

This plugin has two modes:
1. Manual Mode - This allows for single trip mining sessions, you will be directed to the bank & then you will be able to select a star in the panel to star heading to this star & mining it
2. Use Nearest High Tier Star - This is a fully automatic mode. This will check the distance of stars from an acceptable tier range & map them based on distance, then select the closest star that is the highest tier.

This plugin supports inventory setups, by default it uses an inventory setup name called "Shooting Star" but this is configurable in the plugin settings. It has been tested for many edge cases & when not using inventory setups the only inventory requirement is a pickaxe.

Other changes:
AutoCookingScript: Removed Pizza's ( going to merge this function into a baking script that is apart of the cooking plugin )
AutoFishingScript: Added Cave Eels
Rs2Player: adjusted waitForXPDrop methods to rely on each other & added shortestDistanceTo to check distance using ShortestPathPlugin (this method can take some time to return)


gmason0:
I apologize for the last second changes, I had found that when there is a lot of stars in your acceptable tier range the Rs2Player.shortestDistanceTo seems to take a bit to perform that many pathing calculations at once - so I added a check to see if the distance already exists before calculating.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/walker/Rs2Walker.java
```diff
@@ -801,4 +801,19 @@ public void setStart(WorldPoint start) {
...
     }
+
+    /**
+     * Checks the distance between startpoint and endpoint using ShortestPath
+     * 
+     * @param startpoint
+     * @param endpoint
+     * @return distance
+     */
+    public static int shortestDistanceTo(WorldPoint startpoint, WorldPoint endpoint) {
```
chsami: shortestDistanceTo is not a clear method name. I would suggest getDistanceBetween(WorldPoint startPoint, WorldPoint endPoint)

chsami (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/walker/Rs2Walker.java
gmason0: This has been updated!

gmason0: One last change :( I didn't change the Rs2Player method

chsami (approved):

chsami:
Thank you so much for your significant contribution! I really appreciate the plugin and the effort you put into it. A special thanks as well for providing a video demonstration and a thorough explanation—it was very helpful!

PR #379, opened 2024-09-08
NeverLog Addition

Zukish:
A plugin to extend the logout timer to 6 hours, rather than the default 5 minutes or extended 25 minutes.

This is a simple toggle plugin, either on for extended logout duration or off to default back to 5 mins.


chsami:
@See1Duck @Zukish 

I think this is a great addon to be moved under QOL plugins. We're trying to avoid having a big list of plugins.

What do you guys think?

See1Duck:
> @See1Duck @Zukish 
> 
> I think this is a great addon to be moved under QOL plugins. We're trying to avoid having a big list of plugins.
> 
> What do you guys think?

Yeah I can definitely add this to QoL

Zukish:
Closing due to merge into QoL :)

PR #385, opened 2024-09-15
Added option to select bones to bury

mysnphit:
Simple configuration option added to AIO Fighter


chsami:
whats the usecase for this?

mysnphit:
I used this to automatically pickup and bury the big bones obtained from the giant frogs in the Lumbridge Swamps for example. Before these changes, only regular bones would be picked up.

chsami:
The code should do a .contains("bones") on the items; meaning it should loot all the "big bones", "wyvern bones", "bones" and so on. If this is broken, we should fix that instead of creating a custom config with a comma seperated list.

PR #394, opened 2024-09-28
Development

See1Duck:
# **Changelog**

### **Upkeep**
- **Auto Run**: Introduced a boolean setting to automatically enable running when the player's energy is sufficient. 
- **Auto Stamina**: Added a boolean configuration to automatically consume Stamina Potions when available.
- **Stamina Threshold**: Implemented a customizable run energy threshold setting (1-99) to determine when a Stamina Potion should be consumed. The default value is set at 50%.
### **Overlay**
- **Changed**: The Wintertodt overlay has been separated from the main overlay.
### **Misc**
- **Changed**: The `[WintertodtScript]` class has been moved to its own package.

### **Looting**
- **Added**: Forced looting of items option, this will make the loot script loot regardless if the player is in combat or not.

### **Combat**
- **Improved**: The script will now prioritize attacking the NPC that is attacking the player.
- **Improved**: The script has been reworked for better performance and efficiency by removing redundant code and improving the overall structure.

### **Prayer Flicking**
- **Minor Fix**: The script will now wait to do the first flick until the player is in combat and the NPC hits the player.

## **[Amethyst Plugin]**
- **Improved**: selection of ore vein.

## **Class: [Rs2GroundItem]**
- **Major Fix**: Looting nolonger getting stuck in `pauseAllScripts` mode.
- **Minor Fix**: `LootingParamaters` for looting by value now has a default minimum Quantity change from 0 to 1.
- **Added**: 
`
waitForGroundItemDespawn(Runnable actionWhileWaiting,GroundItem groundItem)
` 
method that waits for the ground item to despawn while performing an action. (The action should be an interaction with the ground item)
<p> This method proves to be more reliable than `Rs2Inventory.waitForInventoryChanges` as it could cause endless loops of trying to loot the same despawned item if the items was looted by another player 
or if the player has a Open Herb Sack, Gem Bag or Seed Box etc... and the item was deposited directly into one of those containers bypassing the inventory, resulting in no inventory change.
<p> This method won't be plagued by the same issues as it monitors the ground item itself for despawn/change.

## **Class: [Microbot]**
- **Added**: `useStaminaPotsIfNeeded` variable to determine if the script should consume Stamina Potions when available.
- **Added**: `runEnergyThreshold` variable to determine the run energy threshold at which the script should consume Stamina Potions.
  
## **Misc**
- **Removed**: Auto Spec enabling has been removed from `[Script]` Class parent due to if one or more scripts its enabled this will cause multiple calls to spec, causing issues and spamming interactions. This feature is now available in the QoL Plugin or should be implemented in your individual scripts.
- **Added**: A check to `[Script]` Class parent if the `clickHereToPlayButton` already has been clicked, this will prevent the scripts with a short sleep schedule from spamming the button.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/Script.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/Script.java
@@ -30,6 +30,7 @@ public abstract class Script implements IScript {
     public ScheduledFuture<?> mainScheduledFuture;
     public static boolean hasLeveledUp = false;
     public static boolean useStaminaPotsIfNeeded = true;
+    public static boolean clickedPlayButton = false;
 
     public boolean isRunning() {
         return mainScheduledFuture != null && !mainScheduledFuture.isDone();
@@ -55,6 +56,7 @@ public abstract class Script implements IScript {
             System.out.println(e.getMessage());
         }
     }
+
     public boolean sleepUntil(BooleanSupplier awaitedCondition) {
         return sleepUntil(awaitedCondition, 5000);
     }
@@ -109,7 +111,7 @@ public abstract class Script implements IScript {
 
     public boolean run() {
         hasLeveledUp = false;
-        Microbot.getSpecialAttackConfigs().useSpecWeapon();
+        //Microbot.getSpecialAttackConfigs().useSpecWeapon();
 
         if (Microbot.pauseAllScripts)
             return false;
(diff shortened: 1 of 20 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/microbot/Script.java
```diff
@@ -122,16 +124,18 @@ public boolean run() {
                 Rs2Keyboard.keyPress(KeyEvent.VK_SPACE);
             }
             Widget clickHereToPlayButton = Rs2Widget.getWidget(24772680); //on login screen
-            if (clickHereToPlayButton != null && !Microbot.getClientThread().runOnClientThread(clickHereToPlayButton::isHidden)) {
+            if (clickHereToPlayButton != null && !Microbot.getClientThread().runOnClientThread(clickHereToPlayButton::isHidden) && !clickedPlayButton) {
                 Rs2Widget.clickWidget(clickHereToPlayButton.getId());
+                clickedPlayButton = true;
```
chsami: I rather add a sleepuntil with a timeout of maybe max 5 sec, before continue. To make sure the interface is gone before clicking again. IN this case if for some weird reason the interface was not closed, the bot will get stuck because the clickedPlayButton will be set on true even tho the interface is maybe still available.

chsami (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/Script.java
See1Duck: Updated the code to use sleepUntil instead. I also use a synchronized block to prevent multiple plugins from clicking the play button at the same time

chsami (approved):

PR #406, opened 2024-10-11
Add Alternative WalkTo Interaction

YifanLi5:
When using the original walkTo(WorldPoint target, int distance) method, there’s a noticeable delay between when the walking finishes and when the script resumes. This often results in the player stopping a few tiles short of a GameObject, pausing briefly before interacting with it.

This pull request introduces a new walkTo-based function that:

- Blocks execution until the walking task completes.
- Automatically attempts to resolve and interact with a GameObject based on a given condition when the player gets near the target.
- 
The goal is preform a more seamless transition from walking to interacting, all within a single thread so there doesn't need to be any synchronization.

Since the original walkTo is non-blocking, the code that invokes it doesn't know when the main walking loop finishes. This uncertainty likely causes delays, as the invoking code has to wait until the player stops moving.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/callback/ClientThread.java
+++ b/runelite-client/src/main/java/net/runelite/client/callback/ClientThread.java
@@ -69,6 +69,19 @@ public class ClientThread
 		scheduledFuture = scheduledExecutorService.submit(() -> method.call());
 	}
 
+	/***
+	 * Runs a Callable<Boolean> on a seperate thread. Similar to runOnSeperateThread except this
+	 * method blocks until the callable returns.
+	 * @param method Callable to submit
+	 * @return the result of the callable
+	 */
+	@SneakyThrows
+	public boolean runBlockingCallable(Callable<Boolean> method) {
+		if (scheduledFuture != null && !scheduledFuture.isDone()) return false;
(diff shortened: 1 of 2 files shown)


YifanLi5:
Would also need to

- Add a new parameter to the function to specify the desired GameObject's action.
- Include a separate assertion to verify if the GameObject action was successfully performed.

chsami:
You can increase the finish distance for your walkTo method

`walkTo(WorldPoint target, int distance)`

By default the distance is the one defined in your config for the web walker plugin.

image


A quick banking example on how to interact with an object while you are still running:

```
if (!Rs2Bank.isOpen()) {
    boolean reachedLocation = Rs2Walker.walkTo(BankLocation.FALADOR_WEST.getWorldPoint(), 12);

    if (reachedLocation) {
        if (Rs2GameObject.interact(24101, "bank")) {
            sleepUntil(Rs2Bank::isOpen);
        }
    }
}
```

Result:

https://github.com/user-attachments/assets/c180534f-6e5f-49e9-a95e-dac6b6dc2a0b



Let me know if this satisfies your requirement.

Thanks!

chsami:
No reaction, closing the PR

PR #408, opened 2024-10-11
Add Developer Overlay to ShootingStarPlugin + Fletch Arrow Shaft

gmason0:
While I am working on my revenant plugin, there has been a couple commits I would like to just get merged upstream..

# ShootingStar
- Add config option for show/hide overlay
- Added developer overlay to show more information about current star & runtime
- Add config option to force breaks at bank

# AutoWoodcutting
- Added Reset Option to fletch logs into arrow shafts

# Utility classes
- TimeUtils: change getFormattedDurationBetween to properly format negative duration


runelite-client/src/main/java/net/runelite/client/plugins/microbot/woodcutting/AutoWoodcuttingScript.java
```diff
@@ -74,7 +76,7 @@ public boolean run(AutoWoodcuttingConfig config) {
                     return;
                 }
 
-                if (Rs2Player.isMoving() || Rs2Player.isAnimating() || Microbot.pauseAllScripts || Rs2AntibanSettings.actionCooldownActive)
```
chsami, line 77: Is there a reason why this has been removed here?

gmason0: Duplicate check, if you look there is an if statement right after this one for that condition

chsami (approved):

PR #417, opened 2024-10-16
Small Tutorial Island Fixes

gmason0:
I have created two small fixes for the tutorial island script I had found when creating my test accounts..

- I have fixed the issue with it clicking the top left corner / not properly finding the set name button by using the Widget IDs..
- I have also re-order the logic to be more robust by finding existing names & removing them OR invalid names (which were handed before..)
- I have also fixed a part with the account manager where you can be stuck in a loop when talking about worlds..


gmason0:
I was asked via Discord to make the following changes:

- Randomize which ore is mined in the mining guide

chsami (approved):

PR #418, opened 2024-10-17
Update MicroAgilityConfig.java

abdstg:
add config agility boostable


chsami:
What does this config do? You just added a config without functionality?

chsami:
Try to merge the two pr's into one and make it to development instead of main

PR #421, opened 2024-10-18
Agility plugin improvements

Lori-rs:
Added wyrm, ape atoll and shayzien courses, as well as enabled the default antiban settings.

The runelite worldmap plugin doesn't have all of the agility courses, so I created a new list and amended all references to the new enum file.


Lori-rs:
Oops, wrong branch. Closing and submitting to development.

PR #422, opened 2024-10-18
Agility plugin improvements

Lori-rs:
Added wyrm, ape atoll and shayzien courses, as well as enabled the default antiban settings.

The runelite worldmap plugin doesn't have all of the agility courses, so I created a new list and amended all references to the new enum file.


Lori-rs:
Some obstacles in the wyrm courses don't work unless the player is adjacent to them and I can't figure out why. 
TIGHTROPE_55180, TIGHTROPE_55184, TIGHTROPE_55194, ROPE_55186

I think I lack the experience to fix the issue, whatever it is.

chsami (approved):
Thanks for the contribution! <3

PR #426, opened 2024-10-19
bird hunter plugin

changednamedes:
Bird Hunting Plugin mplementation

Description:

Implements a bird hunting script with dynamic trap handling.

Automatically sets a hunting area based on player location and configurable radius.

Manages trap interactions (setting, collecting, and replacing traps).

Automatically walks back to the center of the hunting area if outside.

Includes snare inventory checks based on player level.

AntiBan implementation


chsami (approved):

chsami:
Thanks for contributing <3

PR #427, opened 2024-10-19
Tutorial Island 

gmason0:
In this iteration of this script I have made the following adjustments:

1. Added Random Chance to change gender from current gender
2. Added Random Chance to change pronouns from current pronouns

I also added a helper method into Rs2UIHelper to strip the col tags from text, just figured this could be useful elsewhere..

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/tutorialisland/TutorialIslandScript.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/tutorialisland/TutorialIslandScript.java
@@ -392,7 +392,7 @@ public class TutorialIslandScript extends Script {
             WorldPoint targetPoint = (npc != null) ? npc.getWorldLocation() : worldPoint;
             int distance = Rs2Player.distanceTo(targetPoint);
 
-            if (distance >= 8) {
+            if (distance > 8) {
                 Rs2Walker.walkTo(targetPoint, 8);
             } else {
                 if (Rs2Npc.interact(npc, "Talk-to")) {
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/tutorialisland/TutorialIslandScript.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/tutorialisland/TutorialIslandScript.java
@@ -389,8 +389,11 @@ public class TutorialIslandScript extends Script {
 
         if (Microbot.getVarbitPlayerValue(281) == 610 || Microbot.getVarbitPlayerValue(281) == 620) {
             WorldPoint worldPoint = new WorldPoint(3141, 3088, 0);
-            if (Rs2Player.distanceTo(worldPoint) >= 4) {
-                Rs2Walker.walkTo(worldPoint, 4);
+            WorldPoint targetPoint = (npc != null) ? npc.getWorldLocation() : worldPoint;
+            int distance = Rs2Player.distanceTo(targetPoint);
+
+            if (distance >= 4) {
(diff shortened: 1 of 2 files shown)


gmason0:
1 last change, I had forgotten my adjustments for the mage guide!

runelite-client/src/main/java/net/runelite/client/plugins/microbot/tutorialisland/TutorialIslandScript.java
```diff
@@ -337,8 +389,11 @@ public void MageGuide() {
 
         if (Microbot.getVarbitPlayerValue(281) == 610 || Microbot.getVarbitPlayerValue(281) == 620) {
             WorldPoint worldPoint = new WorldPoint(3141, 3088, 0);
-            if (Rs2Player.distanceTo(worldPoint) >= 4) {
-                Rs2Walker.walkTo(worldPoint, 4);
+            WorldPoint targetPoint = (npc != null) ? npc.getWorldLocation() : worldPoint;
+            int distance = Rs2Player.distanceTo(targetPoint);
+
+            if (distance >= 4) {
```
chsami: this should be greater than 4.  because the webwalker will not execute if the distance is 4.
 
This is the check of the web webwalker:
Rs2Player.getWorldLocation().distanceTo(target) <= distance

chsami (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/tutorialisland/TutorialIslandScript.java
chsami: or better would be to 

if (Rs2Walker.walkTo(targetPoint, 4) && Rs2Npc.interact(npc, "Talk-to")) {
    sleepUntil(Rs2Dialogue::isInDialogue);
}

gmason0: I just increased the range to satisfy the walker requirements, this way it also clicks on the NPC before being so close to the mage guide

runelite-client/src/main/java/net/runelite/client/plugins/microbot/tutorialisland/TutorialIslandScript.java
```diff
@@ -337,8 +389,11 @@ public void MageGuide() {
 
         if (Microbot.getVarbitPlayerValue(281) == 610 || Microbot.getVarbitPlayerValue(281) == 620) {
             WorldPoint worldPoint = new WorldPoint(3141, 3088, 0);
-            if (Rs2Player.distanceTo(worldPoint) >= 4) {
-                Rs2Walker.walkTo(worldPoint, 4);
+            WorldPoint targetPoint = (npc != null) ? npc.getWorldLocation() : worldPoint;
+            int distance = Rs2Player.distanceTo(targetPoint);
+
+            if (distance >= 8) {
```
chsami: this has to be > 8, otherwise it can get stuck in an infinite loop, because the walker will not walk if it is equal to 8

chsami (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/tutorialisland/TutorialIslandScript.java
gmason0: I am sorry, I guess I misunderstood what you were saying before.. Its been changed

chsami (approved):

PR #428, opened 2024-10-21
Updated Scurrius

changednamedes:
Added more config options for food and prayer

Options to select which food, with quantity to withdraw - same with prayer

Fixed the prayers to be used correctly and disable/enable correctly to preserve prayer

fixed dodging of the dangerous tiles as this wasn't working

Added better checking for states, as well as checks for being in the room etc

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/scurrius/ScurriusConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/scurrius/ScurriusConfig.java
@@ -1,18 +1,161 @@
 package net.runelite.client.plugins.microbot.scurrius;
 
+import lombok.Getter;
+import net.runelite.api.ItemID;
 import net.runelite.client.config.Config;
 import net.runelite.client.config.ConfigGroup;
 import net.runelite.client.config.ConfigItem;
+import net.runelite.client.config.Range;
 
 @ConfigGroup("Scurrius")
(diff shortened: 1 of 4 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/microbot/scurrius/ScurriusConfig.java
```diff
@@ -1,18 +1,161 @@
...
 public interface ScurriusConfig extends Config {
     @ConfigItem(
-            keyName = "Prioritize",
+            keyName = "foodSelection",
+            name = "Select Food",
+            description = "Select the type of food you want to use",
+            position = 0
+    )
+    default FoodSelection foodSelection() {
+        return FoodSelection.SHARK;
```
gmason0: Why not use Rs2Food here? It will include alot more options to allow the user to pick whatever food they want..

changednamedes: Forgot all about it and I changed the food a while ago hah. Updated now to use Rs2Food. There is alot of options now though.

FunkyFishie:
is it possible that you change the room to the private instance? and not the FFA?

changednamedes:
> is it possible that you change the room to the private instance? and not the FFA?

Updated now to allow public and private

chsami (approved):

PR #431, opened 2024-10-24
Changed BankStander

lVidStorm:
somebody was asking for me to PR the version of bankstander that I was using. tried not to be too sloppy with it, fixed some things up before getting to this point. heard we want to reduce plugins, so I was a bit lost on giving bank full credit or saying it's my plugin. this version is likely mostly my code, but he provided the foundation for it. let me know if I should give him full credit, or flip how I decided to give credit.


runelite-client/src/main/java/net/runelite/client/plugins/microbot/bankjs/BanksBankStander/BanksBankStanderConfig.java
chsami, line 1: It's not allowed to just remove the script from someone elses folder, either ask permission or add onto the script and list your credits at the top of the files.

chsami (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/bankjs/BanksBankStander/BanksBankStanderConfig.java
lVidStorm: Fair. Since I'm unable to get replies from Bank I'll have to go with the latter option.

PR #432, opened 2024-10-24
adding herbrun script + name descriptor

yairtu:
Herb Run script


.run/Runelite.run.xml
chsami, line 1: Can you undo this change?

chsami (changes requested):
can you undo your change runelite.run.xml

chsami (approved):

PR #441, opened 2024-10-29
found some things that were named incorrectly

lVidStorm:
hopefully this changes the issue


chsami:
Ok thanks I'll test this out later

chsami:
image


I pressed escape and it keeps spamming microbot.showMessage, this is kind aannoying cuz it won't let me stop the plugin, might be better to just return the error once or just the Microbot.log to avoid blocking the user from input

lVidStorm:
I have to restructure the logic to parse int parameters, this could take a little bit. will reopen when finished.

PR #445, opened 2024-10-29
Walker Changes

gmason0:
I am going to use this PR to fix more walker issues throughout the evening ♥ - feel free to merge whenever

Fixes: 
1. Fix Destination Worldpoint for Farming Guild on Skills Necklace

### DIFF
--- a/runelite-client/src/main/resources/net/runelite/client/plugins/microbot/shortestpath/teleportation_items.tsv
+++ b/runelite-client/src/main/resources/net/runelite/client/plugins/microbot/shortestpath/teleportation_items.tsv
@@ -48,7 +48,8 @@
 2933 3295 0			11105;11107;11109;11111;11968;11970		4	3. Skills necklace: Crafting Guild	T	30
 3144 3438 0			11105;11107;11109;11111;11968;11970		4	4. Skills necklace: Cooks' Guild	T	30
 1662 3505 0			11105;11107;11109;11111;11968;11970		4	5. Skills necklace: Woodcutting Guild	T	30
-1248 3725 0			11105;11107;11109;11111;11968;11970		4	6. Skills necklace: Farming Guild	T	30
+1248 3725 0		45 Farming	11105;11107;11109;11111;11968;11970		4	6. Skills necklace: Farming Guild	T	30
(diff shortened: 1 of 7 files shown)


runelite-client/src/main/resources/net/runelite/client/plugins/microbot/shortestpath/teleportation_items.tsv
```diff
@@ -48,6 +48,7 @@
 2933 3295 0			11105;11107;11109;11111;11968;11970		4	3. Skills necklace: Crafting Guild	T	30
 3144 3438 0			11105;11107;11109;11111;11968;11970		4	4. Skills necklace: Cooks' Guild	T	30
 1662 3505 0			11105;11107;11109;11111;11968;11970		4	5. Skills necklace: Woodcutting Guild	T	30
+1248 3725 0		45 Farming	11105;11107;11109;11111;11968;11970		4	6. Skills necklace: Farming Guild	T	30
```
chsami, line 51: The line beneath this one already defines the farming guild

chsami (changes requested):

runelite-client/src/main/resources/net/runelite/client/plugins/microbot/shortestpath/teleportation_items.tsv
gmason0: I added a second option here because of the issue seen in the web walker chat. 

When the walker would take a transport using the skills necklace to farming guild (using the old, the one below) it would at the start point of the path after teleporting outside of the the farming guild, which is not ideal for a player above 45 farming that gets placed inside the farming guild, it will send the player outside of the farming guild & then they will walk back in

The second one, the one i added accounts for when a player is above 45 farming & it would set the start point to inside the farming guild..

I can say, i tried this both on my main & an account that was running 1 farming & got my desired results of choosing the 45 farming for the one on my main, making the start point after taking the transport inside the farming guild and on the level 1 farming it would set the start point outside of the farming guild..

Hope that provides the clarity around this adjustment..

chsami: Okay thats clear, thanks for explaining it.

chsami (approved):

PR #447, opened 2024-10-30
GiantsFoundary Fixes

YifanLi5:
Add check if bank contains enough of the selected bars. Stop script if insufficient materials. 

Prevent script idling when at a foundry station. This can happen if the player's lvls smithing up and if the random event plugin activates. 
doAction variable needs to be set back to true.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/giantsfoundry/GiantsFoundryScript.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/giantsfoundry/GiantsFoundryScript.java
@@ -63,6 +63,7 @@ public class GiantsFoundryScript extends Script {
                 } else {
                     if (weapon != null) {
                         handleGameLoop();
+
                     } else {
                         getCommission();
                         selectMould();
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/microbot/giantsfoundry/GiantsFoundryScript.java
```diff
@@ -245,6 +255,7 @@ public void handleGameLoop() {
                         || Rs2Player.getWorldLocation().equals(new WorldPoint(3371, 11498, 0));
                 if (!doAction && isAtLavaTile) return;
                 Rs2GameObject.interact(LAVA_POOL, "Heat-preform");
+                Rs2Player.waitForAnimation();
```
chsami: Do we really have to wait for animation before continuing?

I say this because at the bottom of this section there is:

 sleepUntil(() -> GiantsFoundryState.heatingCoolingState.getRemainingDuration() <= 1);

YifanLi5: Tested it with them removed, it worked. So they've been removed.

chsami (approved):

PR #449, opened 2024-10-30
Automatically farms blue dragons in Taverley Dungeon

changednamedes:
Automatically farms blue dragons in Taverley Dungeon, collects bones and banks/restocks.

Option to enable scales to be looted too.

You must have a falador teleport in your inventor, runes or teletab and also the same amount of food that you have in your config.

This uses a safe spot, so you should need anymore more than 1-3 food, used when running there normally.

If you are less than 70 agility, you need a dusty key

The plugin will bank and restock the food

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/zerozero/bluedragons/BlueDragonsConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/zerozero/bluedragons/BlueDragonsConfig.java
@@ -1,0 +1,95 @@
+package net.runelite.client.plugins.microbot.zerozero.bluedragons;
+
+import net.runelite.client.config.*;
+import net.runelite.client.plugins.microbot.util.misc.Rs2Food;
+
+@ConfigGroup("bluedragons")
+public interface BlueDragonsConfig extends Config {
+
+    @ConfigItem(
+            keyName = "How To",
+            name = "How to use",
+            description = "How to use the blue dragon plugin",
+            position = 0
+    )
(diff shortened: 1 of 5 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/microbot/zerozero/bluedragons/BlueDragonsConfig.java
```diff
@@ -0,0 +1,95 @@
...
+
+    @ConfigItem(
+            keyName = "How To",
+            name = "How to use",
+            description = "How to use the blue dragon plugin",
+            position = 0
+    )
+    default String howTo()
+    {
+        return "Start with falador runes/teletab and correct number of food from config";
```
chsami: There is a @ConfigInformation annotation you can use to display the description for your script, this can never be overwritten and will be displayed at the top of your script. The content of the annotation is html, so you can add some styling in it etc..

Example of driftnet plugin:

@ConfigInformation("Start this script at the driftet fishing area. <br /> Make sure to have driftnet in your inventory or driftnet stored with anetta.")

chsami (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/zerozero/bluedragons/BlueDragonsConfig.java
changednamedes: All done

chsami (approved):

PR #459, opened 2024-11-02
karambwans and gold bars

Gabulhas:
This includes

- Karambwan fishing script (useful for ironmen)
    - basically fishes at karambwans spot, uses fairy ring to go to zanaris to bank
    - it uses zanaris instead of other fairy ring (ex: closer to a bank) because it's the most afk (and less detectable way) to do so
    - TODO: automatically get more karambwaji (bait) and use the fish barrel
    
- Gold bars at Blast furnace
    - Simple but works
    - No stamina support


.run/Runelite.run.xml
```diff
@@ -1,6 +1,6 @@
 <component name="ProjectRunConfigurationManager">
```
chsami, line 1: This file should not be in your PR, please change it back to the original

runelite-client/src/main/java/net/runelite/client/plugins/gabplugs/goldrush/GabulhasGoldRushConfig.java
```diff
@@ -0,0 +1,47 @@
+package net.runelite.client.plugins.gabplugs.goldrush;
```
chsami, line 1: Please put all your files under plugins/Microbot

This is easier to manage all the microbot specific plugins

chsami:
You should make your PR to development branch and not main.

main gets updated once development is stable, so you can't make a PR to it directly.

chsami (changes requested):

PR #465, opened 2024-11-03
-Added fossil island shortcut near hardwood trees

Ratty-p00ned:


Ratty-p00ned (commented):
I have no idea why simply just adding this to the config makes it cast bones to peaches, if you could tell me why its working that would be good lmao. 

tested on accounts with btp unlocked and locked- it doesnt cast unless it can which is great, but more likely has to do with rs2magic i am guessing?

runelite-client/src/main/java/net/runelite/client/plugins/microbot/magetrainingarena/MageTrainingArenaConfig.java
```diff
@@ -43,6 +43,18 @@ public interface MageTrainingArenaConfig extends Config {
     )
     String graveyardSection = "graveyard";
 
+    @ConfigItem(
```
chsami, line 46: There's no need to add a config. The mage training arena checks if a player can cast bones to peaches.

        if (btp == null)
            btp = Rs2Magic.canCast(MagicAction.BONES_TO_PEACHES);

runelite-client/src/main/resources/net/runelite/client/plugins/microbot/shortestpath/agility_shortcuts.tsv
chsami, line 1: Please remove all the unncessary tabs.

Try to keep your changes as concise as possible

chsami (changes requested):

PR #470, opened 2024-11-03
Development

Ratty-p00ned:
Fossil island agility shortcut
Add moonlight potions to fighter


Ratty-p00ned:
sorry new to all this, is it normal for there to be so many commits even if they were things i reverted? is there any way to clean that up or is it normal/required?

chsami (approved):

PR #484, opened 2024-11-08
Thieving plugin improvements

Cicire:
- Added a toggle in the configuration to enable or disable automatically eating food
- Check if still pickpocketing by checking the animation of the player instead of setting the boolean.
    - This will make sure that when interrupted it will automatically start thieving again instead of waiting for the next round of distraction.
- When thieving wealthy citizen always wait with opening coin pouches until it's done thieving


runelite-client/src/main/java/net/runelite/client/plugins/microbot/thieving/ThievingScript.java
```diff
@@ -73,6 +62,23 @@ public boolean run(ThievingConfig config) {
...
 
+    private void handeFood() {
+        List<Rs2Item> foods = Rs2Inventory.getInventoryFood();
+
+        if (foods.isEmpty()) {
+            openCoinPouches(1);
+            bank();
+            return;
+        }
+        if (Rs2Inventory.isFull()) {
```
Cicire: I've seen that this should not be part of this method I'll get it out otherwise it will cause a bug not to drop items. Looking into it right now.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/thieving/ThievingConfig.java
```diff
@@ -76,11 +76,20 @@ default boolean equipBook() {
     )
     String food = "Food";
 
+    @ConfigItem(
```
chsami, line 79: No need to add an extra config if you do not want to use food. Just change the hitpoints to 0. This will make the script never use food

chsami (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/thieving/ThievingConfig.java
Cicire: Done this because when there is no food in the inventory it will automatically bank. I can also change it to when food is set to 0 it will skip the banking for food. let me know what you think.

chsami: > Done this because when there is no food in the inventory it will automatically bank. I can also change it to when food is set to 0 it will skip the banking for food. let me know what you think.

Ok fair point

chsami (approved):

PR #486, opened 2024-11-08
NMZ plugin improvements (rapid heal flick)

Cicire:
- Fix issue where random was not random because it reset every loop
- Fix issue where it ignored using the rock cake

_I'm aware that the walk to center is still broken most likely caused by webwalker_

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/nmz/NmzScript.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/nmz/NmzScript.java
@@ -180,7 +180,9 @@ public class NmzScript extends Script {
             }
         }
 
-        randomlyToggleRapidHeal();
+        if (config.randomlyTriggerRapidHeal()) {
+            randomlyToggleRapidHeal();
+        }
     }
 
     public void randomlyToggleRapidHeal() {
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/microbot/nmz/NmzScript.java
```diff
@@ -150,27 +159,30 @@ public boolean interactWithObject(int objectId) {
...
+                maxHealth = Random.random(2, 8);
+            }
         }
-    }
 
-    public void handleLocatorOrbUsage() {
-        if (Microbot.getClient().getBoostedSkillLevel(Skill.HITPOINTS) > Random.random(1, 5) && !useOverload
-                && Microbot.getClient().getBoostedSkillLevel(Skill.RANGED) != Microbot.getClient().getRealSkillLevel(Skill.RANGED)) {
-            Rs2Inventory.interact(ItemID.LOCATOR_ORB, "feel");
+        if (config.randomlyTriggerRapidHeal()) {
```
chsami, line 184: why is rapid heal applied here but not inside handleDwarvenRockCake?

It also seems that dwarenRockCake will set maxHealth

Cicire: I'm not that sure where rapid heal is normally used for within NMZ. So this part is a bit confusing to me.

Cicire: I've combined the two methods into this one because they were mostly doing the same thing. Only difference was the interaction with the item.

chsami (approved):

PR #490, opened 2024-11-12
Development

gmason0:
I have created a list of commits to be reviewed!

## Rs2Dialogue 
- Methods to handle combination dialogue, "how many would you like to make"
- Method to fetch the question inside of the dialog

## Rs2Widget
- Refactored implementations of findWidget to rely on new helper methods, in-order to make things more clear.

## Rs2Bank
- Methods to withdrawAllButOne
- Methods to withdraw & deposit runepouch

## Rs2Inventory
- Adjusted getInventory() method in-order add support for natural mouse when shop is open, when GE is open OR when deposit box is open.

## WebWalker
- Added Port Phasmatys Bank Location
- Removed duplicate transport to Zanaris, this will just use bks now.
- Removed enableAutoRun inside of WalkerScript, I feel this should be enabled by script-plugins, not system default.

## BanksShopperPlugin
- Refactored member variables to be inside of the plugin, this allows these values to be updated by the onConfigChanged method
- Forced Banking to only happen when the action is set to buy

## RougesDenPlugin
- Set enabledbydefault to false & added plugin descriptor in-order to find the plugin easier.


runelite-client/src/main/java/net/runelite/client/plugins/microbot/bankjs/BanksShopper/BanksShopperPlugin.java
```diff
@@ -63,15 +90,53 @@ public void onChatMessage(ChatMessage event) {
...
             }
         }
     }
+    
+    @Subscribe
+    public void onConfigChanged(ConfigChanged event) {
+        if (!event.getGroup().equals(BanksShopperConfig.configGroup)) return;
+        
+        if (event.getKey().equals(BanksShopperConfig.npcName)) {
+            npcName = config.npcName();
```
chsami, line 102: I think it would make more sense to call BankShopperScript.npcName instead of defining the npcName in the plugin and injecting the plugin inside the script which is an anti pattern because plugin contains an instance of the BankShopperScript, which creates a infinite loop of references.

chsami (changes requested):
I would try to avoid injecting plugin inside script because a script instance is created inside a plugin. This will create a infinite reference of plugin.script.plugin.script.plugin.script etc...

runelite-client/src/main/java/net/runelite/client/plugins/microbot/bankjs/BanksShopper/BanksShopperPlugin.java
gmason0: I dont like using static fields for member variables if it can be otherwise avoided. I have fixed the aforementioned issue by making the script private inside of the plugin and removing the @Inject annotation from the constructor of the Script class.

You can no longer call plugin.script.plugin.script. etc etc.

gmason0: I have corrected this once more, you still need to have inject annotation on the constructor if you want to pass arguments inside of the constructor. 

The ultimate fix was making the script private inside of the plugin class.

chsami (approved):

PR #496, opened 2024-11-13
ClueSolverV2

Budbomber:
Introduced a new plugin Clue Solver V2 aimed at automating clue-solving tasks.

Configuration: Provides settings for enabling the plugin, task intervals, and cooldowns.

Overlay: Displays the plugin status and ensures necessary plugins are active.

Plugin Setup: Defines the behavior of the plugin during startup and shutdown.

Script: Implements various tasks required to solve different types of clues, ensuring seamless automation.

(somehow managed to include my run config no idea how to remove)


.run/Runelite.run.xml
chsami, line 1: Leave this file out of the commit

chsami:
I saw that you used reflection for getting the values from the clue plugin. Is there a reason why you didn't make variables public and get the values that way?

chsami (commented):
I saw that you used reflection for getting the values from the clue plugin. Is there a reason why you didn't make variables public and get the values that way?

Budbomber:
> I saw that you used reflection for getting the values from the clue plugin. Is there a reason why you didn't make variables public and get the values that way?

I guess it was for reusability, rather than editing the actual cluescrollsplugin, it performs well and has not seemed to cause any issues, meaning only key parts of this code should need updating. At least that's how I saw it when I started building

.run/Runelite.run.xml
Budbomber: Done and removed

chsami:
> > I saw that you used reflection for getting the values from the clue plugin. Is there a reason why you didn't make variables public and get the values that way?
> 
> I guess it was for reusability, rather than editing the actual cluescrollsplugin, it performs well and has not seemed to cause any issues, meaning only key parts of this code should need updating. At least that's how I saw it when I started building

I mean I'm okay with it, I'm just not fond of contributing to reflection stuff, it's complex and can break on runtime. I like simple and compile time, thats why i would rather expose the variables you need for your own plugin or just copy the whole clue plugin in the microbot folder and use it there

.run/Runelite.run.xml
chsami: It's still visible in the PR

chsami (changes requested):

Budbomber (commented):
Ok, I (think) the run is removed, regarding reflection etc, thats fair enough - What would be best bet, just copy a version of cluescrolls or move the cluescrollsplugin and basically set everything I need to public.

PR #497, opened 2024-11-13
AutoSmelting Minor Bug Fixes

Omeesian:
Removed line of sight requirement when looking for a furnace to prevent player for getting stuck indefinitely when out of line of sight but close to furnace.

Added support for Blurite to fix more advanced bars from trying to smelt the incorrect options. Bars after Bluerite were 1 off when selecting which bar to smelt. Eg Trying to smelt steel bars would try and smelt silver bars.


chsami:
The development branch already has these fixes.

PR #501, opened 2024-11-17
ClueSolver V2 - Massively refactored to not use reflection

Budbomber:
No longer using reflection, solely relies on the cluescroll plugin - Have moved it and changed everything that we will need to public, base is pretty much sorted, couple of key things left to do which include

- Firming up walking logic, walker to get there, upon success fastwalk canvas to precise loc, some clues will need defined areas etc as to make walker not fire again until needed.

- Overal logic of the clues needs improving (Currently only been testing on mediums) But the general flow of clues can differ from (Easy, med, hard, elite, master) Ideally each of our tasks "Should" be able to cover this for whatever type of clue.

- Implement custom overlay, (task state, runtime, value gained, collections logged etc)

- Add further clue tasks, MapClues,  Cipher clues, ThreeStep, Skill challenge (Charlie or falo etc) Implement the logic for fairy ring clues and music clues

-  Find a way of getting feedback whether from discord or create a simple logging system that will save debug logs for users of the jar etc, so they can send logs if needed
 

- LOTS MORE


chsami (changes requested):
it seems like you deleted the runelite.run.xml, please undo this

Budbomber (commented):
Sorry, appears I forgot to add both of the run xml's both re-added

.run/Runelite with proxy.run.xml
```diff
@@ -2,8 +2,9 @@
   <configuration default="false" name="Runelite with proxy" type="Application" factoryName="Application">
     <option name="MAIN_CLASS_NAME" value="net.runelite.client.RuneLite" />
     <module name="client" />
-    <option name="PROGRAM_PARAMETERS" value="-developer-mode -proxy=144.49.99.215:8080 -proxy-type=http" />
```
chsami, line 5: please  put back the proxy settings as an example for the proxy.run.xml

chsami (changes requested):
please  put back the proxy settings as an example for the proxy.run.xml

.run/Runelite.run.xml
```diff
@@ -1,13 +1,13 @@
...
-    <module name="client" />
-    <option name="PROGRAM_PARAMETERS" value="-developer-mode -clean-jagex-launcher" />
-    <option name="VM_PARAMETERS" value="-ea" />
-    <method v="2">
-      <option name="Make" enabled="true" />
-    </method>
-  </configuration>
-</component>
+    <configuration default="false" name="Runelite" type="Application" factoryName="Application" singleton="false">
+        <option name="ALTERNATIVE_JRE_PATH" value="temurin-23"/>
```
chsami: temurin-23 should be set back to "11"

chsami (changes requested):

chsami (approved):

chsami:
hmm your build is failing, please take a look

chsami (changes requested):
build failed

Budbomber (commented):
As far as i'm aware, all is good build is fine correct runs etc.

chsami (approved):

PR #506, opened 2024-11-18
Transport Change + BanksShopperScript Fix

gmason0:
## ShortestPathPlugin
- Went through all transports & converted them from 'menuOption menuTarget objectID' to 'menuOption;menuTarget;objectID', this is to fix the issue where menuOption or menuTarget that contain multiple words not being processed correctly

## BanksShopperScript
- Forgot a break statement at the end of the shopping case, causing the plugin to fall through to the banking state, regardless if the state has changed


gmason0:
I also removed levers.tsv & portals.tsv as it seems to contain the same values as teleportation_levers.tsv & teleportation_portals.tsv. These files were not even being added inside of the loadAllResources method.

chsami (approved):

PR #510, opened 2024-11-19
Web Walker Fixes + Giants Foundry Request

gmason0:
## GiantsFoundry
- Added a config option to define the amount of bars used

## ShortestPathPlugin
- Changed how varbits are evaluted by adding an operator field into the TransportVarbit class. This will allow us to specify if a varbit value is greater than, less than or equal to. 
- Added logic inside of the useTransport method to check the charges on the Chronicle & to only use it if there are charges
- Broke out the Rada's blessing inside of the teleportoration_item.tsv to check the usage varbit based on how many daily uses are left

## ShootingStarPlugin
- Added a hashCode method to remove the warning messages when overriding the equals method


gmason0:
I've added a few more varbit checks for charged/daily teleports

1. Kharedst's memoirs
2. Explorer's ring (2)
3. Rada's blessing (Mount Karuulm) (3)

chsami (approved):

PR #518, opened 2024-11-22
Antiban Crash Fix + Webwalker

gmason0:
## Rs2Antiban
- Created a method called validateAndSetBreakDurations to ensure that the microBreakDurationLow & microBreakDurationHigh are always within the constraints of the panel

## ShortestPathPlugin
- Updated DisplayInfo on Farm Teleport for Ardougne Cloak to fix teleport when equipped (still works in inventory)


gmason0:
If you can think of a better way to achieve ensuring that the microBreak values are set properly, let me know!

chsami (approved):

PR #523, opened 2024-11-26
Fixed BirdHunter

changednamedes:
Switched the IDS again, they like changing them back and forth


chsami:
Use ObjectID instead of a magic number

chsami (changes requested):
Use ObjectID of itemid instead of magic numbers

changednamedes:
> Use ObjectID of itemid instead of magic numbers

Done

chsami (approved):

PR #526, opened 2024-11-27
[Agility] Remove delay before starting new course lap

YifanLi5:
This PR reverts changes to RS2Walker that were present in a previous PR, so it only makes changes to the agility plugin's code.

Is moving lags behind when the player actually stops. (ex Player stops first, Rs2Player.isMoving() will return true for another tick or so until it returns false)

Also RS2Walker always ends with a canvas walk (yellow click) which I don't like but will not attempt to fix in this PR. However it was always a part of the agility script (and everything else)


telnorp (approved):
looks good and simple 👍

chsami (approved):

PR #533, opened 2024-11-29
Webwalker Fixes + Plugin Bug Fixes

gmason0:
## ShortestPathPlugin
- Adjusted useTransport by moving isTeleportationItemUsable as the lowest priority to ensure all other fields are completed before checking if the teleport is usable
- Add Draynor Manor Tree to restricted tiles
- Add ladder for Corsair Cove dungeon
- Add Varbits for Locked Quetzals

## Login
- Adjusted No-Args Constructor in Login.java to fetch current world if its a valid world, if not then use getRandomWorld

## AutoWoodcutting
- Adjust logic for 'You can't light a fire here'

## ShootingStarPlugin
- Filter based on if the player is or is not in a Seasonal World

## BreakHandler
- Added Config Option to use RandomWorld with the additions of Login class, by default it will send you to the same world

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/breakhandler/BreakHandlerConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/breakhandler/BreakHandlerConfig.java
@@ -68,11 +68,21 @@ public interface BreakHandlerConfig extends Config {
     }
 
     @ConfigItem(
+            keyName = "useRandomWorld",
+            name = "Use RandomWorld",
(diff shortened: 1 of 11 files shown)


gmason0:
~I also just made an adjustment to recalculate the path if you are no longer near the path. This seems to work fine, it seems to re-start the pathfinding if you do a long agility shortcut (for example, barbarian village to Gnome Stronghold. Still works fine though.~

I've removed this until it can be further debugged

chsami (approved):

PR #540, opened 2024-12-03
Simple walker

YifanLi5:
Adds a new walking function for handling shorter walks.
Its purpose is to get in viewing range of an object to be interacted but is more limited in scope (no teleports/shortcuts).
While the normal walker does work, it will always insist on a yellow click move to finish off before returning. Which this alternate walk loop does not.

Also modifies the agility plugin to utilize this short walk function. 
~~This PR is a draft for now I will modify the agility plugin with a config to use this new walker or not.~~ 
I have added the config option that determines whether or not to use the new walker.

See this line:[ Rs2Walker.java#L313](https://github.com/chsami/Microbot/blob/0defeb33336cbc61df7616ca3d91cee24b81c86a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/walker/Rs2Walker.java#L313)


ghost:
Doesn't Rs2Walker.walkFastCanvas(World point) or Rs2Walker.walkFastLocal (Local point) achieve this? 
There is also Rs2Walker.walkMiniMap if you don't want to yellow x the canvas

chsami:
seems to be no need to have a short version of the walker

PR #543, opened 2024-12-03
Initial commit of Agility pyramid plugin

Blomgreen:
A plugin for automatic completion of the agility pyramid.
Core functionality of this script entails completing the agility pyramid,
exchanging the golden pyramids for GP with Simon the NPC,
and banking to restock waterskins and any given food item.

--
There is a "dead" method within the script, which is when I tried my hand with coming up with logic to intentionally fail and navigate down the pyramid, when out of resources. This function is now implemented in such a way, that it should never get called. It's only semi-functional, so I will keep it this way for now.
Instead I opted for letting the user set a minimum amount of food items and waterskins to have in the inventory, before starting another run. This way it's the users responsibility to not die, instead of mine.

of my current 170+ lap count, the script hasn't gotten stuck in the last 100 of them, which is why I'm confident that I've ironed out all major bugs, and it's now ready for a release.

Screenshot 2024-12-03 184220

In the future I plan to add handling for the large moving bricks that push you off, if you time it wrong. This is the only "skillcheck" on the course, and it's not often a problem, hence why I'm putting it off.

Video of full run / claiming rewards / banking

https://github.com/user-attachments/assets/336dbed2-9262-40ba-862b-d239be4157f7

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/AgilityPyramid/AgilitypyramidScript.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/AgilityPyramid/AgilitypyramidScript.java
@@ -1,0 +1,57 @@
+package net.runelite.client.plugins.microbot.AgilityPyramid;
+
+import net.runelite.client.plugins.microbot.AgilityPyramid.tasks.BankHandler;
+import net.runelite.client.plugins.microbot.AgilityPyramid.tasks.ClaimReward;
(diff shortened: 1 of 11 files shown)


chsami (changes requested):
This script should be put in the agilityscript that we already have and configured as a course. This is to avoid having to many plugins

chsami:
@Blomgreen Are you going to make the changes in the PR or can I close this?

Blomgreen:
At least not as of yet. Not sure if I will

PR #558, opened 2024-12-08
Added proximity alerts

ghost:


runelite-client/src/main/java/net/runelite/client/plugins/microbot/playerassist/loot/LootScript.java
```diff
@@ -38,12 +36,53 @@ public boolean run(PlayerAssistConfig config) {
...
+    }
+
+    private void lootCustomItems(PlayerAssistConfig config) {
+        if (config.customItemsToLoot() == null || config.customItemsToLoot().trim().isEmpty()) {
+            return;
+        }
+
+        List<String> itemsToLoot = parseCSVString(config.customItemsToLoot());
+        for (String itemName : itemsToLoot) {
+            if (Rs2GroundItem.loot(itemName, 1, config.attackRadius())) {
```
chsami, line 80: You should use the `lootItemsBasedOnNames` and pass the config params from the script to the LootingParameters object, see "lootArrows" for example

chsami (changes requested):

PR #565, opened 2024-12-11
Auto hop on player detection

changednamedes:
Will now hop worlds if there is someone else using the safes pot


runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/npc/Rs2Npc.java
```diff
@@ -419,6 +420,27 @@ public static List<NPC> getNpcsInLineOfSight(String name) {
...
+     *
+     * @param condition a {@link Predicate} specifying the condition the NPC must meet
+     * @return the closest NPC that satisfies the condition, or {@code null} if none match
+     * <p> Example Usage:
+     * NPC closestBanker = getClosestNpc(npc -> Arrays.asList(npc.getComposition().getActions()).contains("Bank"));
+     * NPC closestGoblin = getClosestNpc(npc -> "Goblin".equalsIgnoreCase(npc.getName()));
+     * NPC closestAttackable = getClosestNpc(npc -> npc.getCombatLevel() > 0 && !npc.isDead());
+     */
+
+    public static NPC getClosestNpc(Predicate<NPC> condition) {
```
chsami: I don't think we need this function as

```
 public static NPC getNpc(int id) {
        return getNpcs()
                .filter(x -> x.getId() == id)
                .findFirst()
                .orElse(null);
    }
```

getNpcs already returns the closest to the player

chsami (changes requested):

PR #576, opened 2024-12-15
Various plugin updates & bug fixes

See1Duck:
# Changelog

## Break Handler
- **Added**: Option to only allow micro breaks via Antiban this will prevent the plugin from taking longer breaks and logging out.

## Mahogany Homes Plugin
- **Added**: Support for NPC Contact spell to get new contracts.
- **Fixed**: Contract dialogues now correctly handle the "Continue" option before selecting a contract.
- **Changed**: Banking will now be done at the closest bank to the contract location instead of the closest bank to the player.

## QoL Plugin
- **Fixed**: NeverLogOut now correctly handles the logout timer.
- **Changed**: Extracted and refactored some of the code to make it more modular.
- **Added**: Auto-Drop option in "Inventory/Equipment" tab to drop items when inventory is full.

## Motherlode Mine Plugin
- **Fixed**: Nolonger gets stuck in idle state if user interrupts the plugin.
- **Added**: Check in settings if upstairs hopper is unlocked. If not, it will use the downstairs hopper while still mining upstairs.
- - **Updated**: Replaced a deprecated method in the API.

## Amethyst Mine Plugin
- **Fixed**: Nolonger gets stuck in idle state if user interrupts the plugin.
- **Added**: support for gem bag.

## Tempoross Plugin
- **Added**: Antiban support.

## AIO Fighter Plugin
- **Added**: Loot List option to select what items to loot.

## Varrock Anvil Plugin
- **Adjusted** Increased the timeout timer from 5 seconds to 10 seconds when waiting for the anvil interface to open to prevent the plugin from getting stuck if player doesn't have run energy.

## Rs2Player Class
- **Fixed**: ``checkIdleLogout`` now does correct calculations.

## Natural Mouse
- **Fixed** Scroll Up now works properly.

## Rs2Antiban Class
- **Updated**: Replaced a deprecated method in the API.

## Rs2Bank Class
- **Added**: Method to empty gem bag.
- **Added**: Method ``getNearestBank(WorldPoint worldPoint)`` to get the nearest bank to a given world point.

## Global Class
- **Added**: Method ``sleepUntil`` with a runnable action while waiting.

## Rs2UiHelper Class
- **Added**: Helper method ``isRectangleWithinRectangle`` to check if a rectangle is within another rectangle.

## Rs2Magic Class
- **Fixed**: Method ``npcContact`` will now check if npc widget is visible before interacting with it if not visible it will scroll to it.
- **Updated**: Replaced a deprecated method in the API.


runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/bank/Rs2Bank.java
```diff
@@ -1296,6 +1296,38 @@ public static BankLocation getNearestBank() {
         }
         return nearest;
     }
+    /**
+     * Get the nearest bank to world point
+     *
+     * @return BankLocation
+     */
+
+    public static BankLocation getNearestBank(WorldPoint worldPoint) {
```
chsami, line 1305: having the same method with a different signature and different implementation is confusing. I would add the code from getNearestBank without param to yours. and call getNearestBank(WorldPoint worldpoint) inside getNearestBank() 

The param you pass to getNearestBank is Rs2Player.getWorldLocation

chsami (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/bank/Rs2Bank.java
See1Duck: Done refactoring the method

chsami (approved):

PR #578, opened 2024-12-16
Takes snow using NeverLogOut logic - Christmas event 2024

ghost:
- When Snow Taking config is selected and 1 tile from the snow, takes snow.
- When about to logout, takes snow to stop you from logging out.
- To be used at one of the 3 Snow piles at Falador Park.
- Inspired by NeverLogOut in QoL.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/holidayevent/CollectSnow.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/holidayevent/CollectSnow.java
@@ -6,13 +6,11 @@ import net.runelite.client.plugins.microbot.util.gameobject.Rs2GameObject;
 import net.runelite.client.plugins.microbot.util.math.Rs2Random;
 import net.runelite.client.plugins.microbot.util.player.Rs2Player;
 
-import java.awt.event.KeyEvent;
-
 public class CollectSnow {
     private static long randomDelay;
 
     public static boolean nearTheSnow() {
-        return Rs2GameObject.findObjectByIdAndDistance(5067, 1) != null;
+        return Rs2GameObject.findObjectByIdAndDistance(19035, 1) != null;
     }
 
     public static void onGameTick(GameTick event) {
@@ -21,6 +19,7 @@ public class CollectSnow {
             if (Rs2GameObject.getGameObjects(19035) != null
                     && nearTheSnow()) {
                 Rs2GameObject.interact(19035, "Take");
+                Microbot.log("Done tooken snow cus about to log");
             } else {
                 Microbot.log("Restart plugin 1 tile to snow");}
             }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/holidayevent/CollectSnow.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/holidayevent/CollectSnow.java
@@ -1,0 +1,28 @@
+package net.runelite.client.plugins.microbot.holidayevent;
+
+import net.runelite.api.events.GameTick;
+import net.runelite.client.plugins.microbot.Microbot;
+import net.runelite.client.plugins.microbot.util.gameobject.Rs2GameObject;
+import net.runelite.client.plugins.microbot.util.math.Rs2Random;
+import net.runelite.client.plugins.microbot.util.player.Rs2Player;
+
+import java.awt.event.KeyEvent;
+
+public class CollectSnow {
+    private static long randomDelay;
+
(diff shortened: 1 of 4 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/microbot/holidayevent/HolidayPlugin.java
```diff
@@ -67,9 +68,11 @@ protected void shutDown() {
...
     @Subscribe
-    public void onGameTick(GameTick tick)
+    public void onGameTick(GameTick event)
     {
-        //System.out.println(getName().chars().mapToObj(i -> (char)(i + 3)).map(String::valueOf).collect(Collectors.joining()));
+        if (config.collectSnow()) {
+            CollectSnow.onGameTick(event);
+        }
 
         if (ticks > 0) {
```
gmasonPCS, line 77: this logic could be removed if you aren't using it

gmason0, line 75: this logic could be removed if you are not using it

runelite-client/src/main/java/net/runelite/client/plugins/microbot/holidayevent/CollectSnow.java
```diff
@@ -0,0 +1,28 @@
...
+import java.awt.event.KeyEvent;
+
+public class CollectSnow {
+    private static long randomDelay;
+
+    public static boolean nearTheSnow() {
+        return Rs2GameObject.findObjectByIdAndDistance(5067, 1) != null;
+    }
+
+    public static void onGameTick(GameTick event) {
```
gmason0, line 16: any reason to leverage onGameTick here? I think it would do justice scanning GameObjects & checking player location by making this class extend Script

gmason0: Just saying that these actions can take a long period of time & may not be best to do on client thread

ghost: I just copied QoL NeverLogOut logic and class implementation and replaced KeyPress SPACE_BAR with clicking on the snow pile. I figured if webWalker was once on client thread and pressing spacebar is fine on client thread that clicking a game object that you must be 1 tile from for it to trigger would be OK. 

Why are you suggesting I extend script in order for the Rs2Object method to work? 

I was thinking to add a flag that the script class can turn false after doing the action but then I would of put this code in the plugin class instead of it's own class like Duck did and it just seemed more complicated than it needed be. I will be testing it TMR to see.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/holidayevent/HolidayPlugin.java
ghost: Yeah what's that doing there

ghost: yeah turns out needs to be there for the event to fire on CollectSnow.Java

runelite-client/src/main/java/net/runelite/client/plugins/microbot/holidayevent/CollectSnow.java
ghost: Just tested everything and works fine on client thread doesn't even stutter or nuttin :D

runelite-client/src/main/java/net/runelite/client/plugins/microbot/holidayevent/CollectSnow.java
ghost, line 1: I had put the animation ID for taking snow instead of the Object ID of the snow

runelite-client/src/main/java/net/runelite/client/plugins/microbot/holidayevent/HolidayScript.java
ghost, line 1: I deleted most of this and replaced it in the isIdle() commit; I was trying to fit too much functionality into 1 boolean

runelite-client/src/main/java/net/runelite/client/plugins/microbot/holidayevent/HolidayPlugin.java
ghost, line 1: this will never have multiple holiday events at one time so I thought it looked better to take the plurality out of the plugin name

PR #610, opened 2025-01-08
Bury Bones and Scatter Ashes Script

AdamRollinson:
First time developing in Java in a long long time, I had a need for a bone bury script to bury some bones I had in the bank. Thought I would extend this to also scatter ashes in case anyone else ever had a use case.

If you need me to amend anything before this goes into development let me know :)

Love the bot and love that this is open source, great work.


chsami:
We have a bankstanding plugin for this

image

PR #627, opened 2025-01-15
Web Walker (shortest path) plugin now checks if you want to run to bank

ghost:
- Normally when walkToBank and walkToBankAndUseBank is activated if you're not running, then running is automatically turned on. For those that may want to conserve run energy when banking a check box is added to Web Walker configuration (default is set to TRUE) that will skip the Rs2Player.toggleRunEnergy(true); command in Rs2Bank class to the aim of conserving run energy when banking. 

- Previous method for casting spells on other players was not working, This has been fixed (and tested thoroughly). 

- walkFastCanvas when arriving at a bank would start to run run as walkFastCanvas was activated. Now with the changes if runToBanks is de-selected and you are 10 tiles from a bank, walkFastCanvas will not toggle run on.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/bank/Rs2Bank.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/bank/Rs2Bank.java
@@ -10,7 +10,6 @@ import net.runelite.client.plugins.bank.BankPlugin;
 import net.runelite.client.plugins.loottracker.LootTrackerItem;
 import net.runelite.client.plugins.loottracker.LootTrackerRecord;
 import net.runelite.client.plugins.microbot.Microbot;
-import net.runelite.client.plugins.microbot.shortestpath.ShortestPathConfig;
 import net.runelite.client.plugins.microbot.util.antiban.Rs2AntibanSettings;
 import net.runelite.client.plugins.microbot.util.bank.enums.BankLocation;
 import net.runelite.client.plugins.microbot.util.equipment.Rs2Equipment;
@@ -48,13 +47,6 @@ import static net.runelite.client.plugins.microbot.util.npc.Rs2Npc.hoverOverActo
 @SuppressWarnings("unused")
 @Slf4j
 public class Rs2Bank {
-
-    private ShortestPathConfig config; // Declare config without @Inject
-
-    // Assuming you can initialize the config directly, e.g., in the constructor
-    public Rs2Bank(ShortestPathConfig config) {
-        this.config = config;
-    }
 
     public static final int BANK_ITEM_WIDTH = 36;
     public static final int BANK_ITEM_HEIGHT = 32;
@@ -1367,7 +1359,7 @@ public class Rs2Bank {
      */
     public static boolean walkToBank(BankLocation bankLocation) {
         if (Rs2Bank.isOpen()) return true;
-        if(ShortestPathConfig.runToBanks()){Rs2Player.toggleRunEnergy(true);}
+        if(Rs2Walker.config.runToBanks()){Rs2Player.toggleRunEnergy(true);}
         Microbot.status = "Walking to nearest bank " + bankLocation.toString();
         Rs2Walker.walkTo(bankLocation.getWorldPoint(), 4);
         return bankLocation.getWorldPoint().distanceTo2D(Microbot.getClient().getLocalPlayer().getWorldLocation()) <= 4;
@@ -1412,7 +1404,7 @@ public class Rs2Bank {
      */
     public static boolean walkToBankAndUseBank(BankLocation bankLocation) {
         if (Rs2Bank.isOpen()) return true;
-        if(ShortestPathConfig.runToBanks()){Rs2Player.toggleRunEnergy(true);}
+        if(Rs2Walker.config.runToBanks()){Rs2Player.toggleRunEnergy(true);}
         Microbot.status = "Walking to nearest bank " + bankLocation.toString();
         boolean result = bankLocation.getWorldPoint().distanceTo(Microbot.getClient().getLocalPlayer().getWorldLocation()) <= 8;
         if (result) {
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/bank/Rs2Bank.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/bank/Rs2Bank.java
@@ -10,6 +10,7 @@ import net.runelite.client.plugins.bank.BankPlugin;
 import net.runelite.client.plugins.loottracker.LootTrackerItem;
 import net.runelite.client.plugins.loottracker.LootTrackerRecord;
 import net.runelite.client.plugins.microbot.Microbot;
+import net.runelite.client.plugins.microbot.shortestpath.ShortestPathConfig;
 import net.runelite.client.plugins.microbot.util.antiban.Rs2AntibanSettings;
 import net.runelite.client.plugins.microbot.util.bank.enums.BankLocation;
 import net.runelite.client.plugins.microbot.util.equipment.Rs2Equipment;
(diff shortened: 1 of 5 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/walker/Rs2Walker.java
```diff
@@ -526,6 +531,15 @@ public static WorldPoint walkCanvas(WorldPoint worldPoint) {
         return worldPoint;
     }
 
+    private static boolean isNearBank(WorldPoint playerLocation, int radius) {
```
gmason0: Why not use the implementation already inside of Rs2Bank?

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/walker/Rs2Walker.java
ghost, line 1: can someone explain to me why these 29, 32 and 45 imports were removed? lmk if I should add them back.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/walker/Rs2Walker.java
```diff
@@ -484,7 +482,14 @@ public static boolean walkFastCanvas(WorldPoint worldPoint) {
 
     public static boolean walkFastCanvas(WorldPoint worldPoint, boolean toggleRun) {
 
-        Rs2Player.toggleRunEnergy(toggleRun);
+        if (Rs2Walker.config.runToBanks()) {
```
gmason0, line 485: This will affect all walkFastCanvas calls, which is used more than just the final click of the Rs2Walker. Alot of people use this in instances/fighting bosses, dont think this is a good addition here

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/magic/Rs2Magic.java
```diff
@@ -160,8 +160,9 @@ public static void castOn(MagicAction magicSpell, Actor actor) {
         if (actor instanceof NPC) {
             Rs2Npc.interact((NPC) actor);
         } else {
-            Point point = Perspective.localToCanvas(Microbot.getClient(), actor.getLocalLocation(), Microbot.getClient().getPlane());
-            Microbot.getMouse().click(point);
+            //Point point = Perspective.localToCanvas(Microbot.getClient(), actor.getLocalLocation(), Microbot.getClient().getPlane());
+            //Microbot.getMouse().click(point);
+            Rs2Player.cast((Player) actor);
```
gmason0, line 165: Direct Casting of player onto actor without checking the instance of the object

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/walker/Rs2Walker.java
ghost: wym

> Why not use the implementation already inside of Rs2Bank?

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/bank/Rs2Bank.java
```diff
@@ -1358,7 +1359,7 @@ public static boolean walkToBank() {
      */
     public static boolean walkToBank(BankLocation bankLocation) {
         if (Rs2Bank.isOpen()) return true;
-        Rs2Player.toggleRunEnergy(true);
+        if(Rs2Walker.config.runToBanks()){Rs2Player.toggleRunEnergy(true);}
```
gmason0, line 1362: Rs2Walker.config is statically set & never updated after the shortestpath plugin is initialized 

could leverage, Microbot.getConfigManager().getConfiguration("shortestpath","runToBanks");

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/magic/Rs2Magic.java
ghost: it works flawlessly vs what was before (broken)

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/bank/Rs2Bank.java
ghost: is that why turning web walker on and off again fixes some issues that come up? I just tested this and works fine

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/walker/Rs2Walker.java
ghost: yes, changing walkFastCanvas will affect all walkFastCanvas calls, that's what changes to walkFastCanvas does

gmason0: Correct, but im saying you're not taking into consideration where this is being used.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/magic/Rs2Magic.java
ghost: you've already seen multiple videos of it working that i've sent you privately over the last weeks

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/walker/Rs2Walker.java
gmason0: There is already a method inside of Rs2Bank, Rs2Bank.isNearBank()

ghost: oh snap wtf

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/magic/Rs2Magic.java
gmason0: Yes, it may work, but you're not seeing the bigger picture of why checking the instance of an object is important. The instanceof operator is used to test whether an object is an instance of a specific class or subclass, or whether it implements a particular interface.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/bank/Rs2Bank.java
gmason0: Yes it may work fine right now, but say if you disable the configuration setting Rs2Walker.config is never updated past when the shortestpathplugin is started.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/walker/Rs2Walker.java
ghost: neither was the person that put it in the final click of the Rs2Walker

gmason0: The logic that you're implementing just does not fit the signature of this method. This is meant to simply click on the canvas to walk.

ghost: and yet it also checks run and toggles it given certain conditions
I added that it checks run and toggles it given certain conditions

i've essentially changed nothing, maybe we wanna consider removing Rs2Player.toggleRunEnergy(toggleRun); entirely however that's the "Fast" element in walkFastCanvas, even though it's setup to have a toggle and not (true)

perhaps there should be a walkCanvas with a toggle or walkSlowCanvas with (false) and either of those should be what is implemented for the final click of the Rs2Walker instead of walkFastCanvas (which should always be (true))

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/walker/Rs2Walker.java
ghost: > Contributor
the rs2bank method breaks it because it keeps calculating the nearest bank by entering getNearestBank

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/magic/Rs2Magic.java
FireStarter-Instigator: yeah just had to put instance of is not null player, just like it say instance of is not null npc right under it, you coulda just told me, didn't know what you meant at the time, null checks are great to do

PR #632, opened 2025-01-16
AIO Magic Fixes

gmason0:
## AIO Magic
- All Scripts: Simply State switching by checking if we have enough for 1 cast of the spell instead of defined number of casts/total amount of alch items
- SuperHeat Script: Fix hasRequiredItems to only check for Coal in Inventory if SuperHeatItem requires Coal
- Alch Script & TeleAlchScript: Fix stuck in CASTING state if there is no item in alch slot by interacting with the inventory once 
- Alch Script & TeleAlchScript: Switch between setWithdrawAsNote & setWithdrawAsItem based on if the item is stackable or not

## Rs2Magic
- Add hasRequiredRunes method for Rs2Spells, similar to the method that was created for Rs2CombatSpells

## Rs2Bank
- Add getBankItem to retrieve the Rs2Item Object from a bank item

## ShortestPathPlugin
- Increase Duration of staircase near gnome stronghold bank


chsami:
I added the changes outside of the AIO magic changes as I reworked some of the stuff.

PR #635, opened 2025-01-17
various fixes to bankstander plugin

lVidStorm:


chsami:
The code of this plugin is getting out of hand, we should seek to simplify this

FireStarter-Instigator:
the plugin also needs more functionality, so I think simplified overhaul would be the next step with this creating new package of aioBankstander

lVidStorm:
alright I'll work on simplifying the code.

PR #636, opened 2025-01-18
Uses config option in Web Walker to walk instead of run when banking.

FireStarter-Instigator:
1. One might wish to walk when banking to preserve energy. As it stands now it's automatic to run, however this PR will introduce a config option in the Web Walker that when de-selected allows the user to walk when banking.
 
2. This change does not intefere directly with the walkFastCanvas method but instead utilises it's (toggleRun) condition, setting it to false when the method is called during the final tile processing of the web walker.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/shortestpath/ShortestPathConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/shortestpath/ShortestPathConfig.java
@@ -277,6 +277,18 @@ public interface ShortestPathConfig extends Config {
         return 1;
     }
 
+    @ConfigItem(
+            keyName = "runToBanks",
+            name = "Activate Run When Banking",
+            description = "Toggle run to on if it's off when banking",
+            position = 22,
+            section = sectionSettings
+    )
+    default boolean runToBanks() {
+        return true;
+    }
+
+
     @Units(
             value = Units.TICKS
     )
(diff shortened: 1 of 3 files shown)


chsami:
When do you want to preserve energy? Does it really matter that much? This seems to add unnecessary complexity.

FireStarter-Instigator:
> When do you want to preserve energy? Does it really matter that much? This seems to add unnecessary complexity.

In f2p and pvp environments you wanna perserve your run energy for the action. To have it wasted when banking is a pain.

FireStarter-Instigator:
Having toggleRun forced upon the user is what complicates things, that's what the PR is trying to fix. Checking a Web Walker config while on the final tile of web walking in order to stop a forced toggleRun in walkFastCanvas is the proposed solution. 

When the forced toggleRun is activated (in the normal functioning of Rs2Bank class) this proposed change adds a check of the Web Walker config that, if it's desired, selects false instead of true, to make it walk instead. 

It's a desperately simple and unintrusive solution and gives freedom to the user to decide if they want to waste their energy or not while banking. If you can explain more about the unacceptable complexity that you see then allow me to close the request and open another with a simpler solution for you.

chsami:
> Having toggleRun forced upon the user is what complicates things, that's what the PR is trying to fix. Checking a Web Walker config while on the final tile of web walking in order to stop a forced toggleRun in walkFastCanvas is the proposed solution.
> 
> When the forced toggleRun is activated (in the normal functioning of Rs2Bank class) this proposed change adds a check of the Web Walker config that, if it's desired, selects false instead of true, to make it walk instead.
> 
> It's a desperately simple and unintrusive solution and gives freedom to the user to decide if they want to waste their energy or not while banking. If you can explain more about the unacceptable complexity that you see then allow me to close the request and open another with a simpler solution for you.

The change is to niche to justify a new config and the extra code that's all.

FireStarter-Instigator:
ok np

PR #646, opened 2025-01-20
Added Lunar Tabelts script & Fixed a bug in 00's bird hunter

Gage307:
Renamed the example script.

Added correct trap ID's so 00's bird hunter can catch Crimson Swifts.


runelite-client/src/main/java/net/runelite/client/plugins/microbot/zerozero/birdhunter/BirdHunterScript.java
```diff
@@ -121,6 +122,8 @@ private void handleTraps(BirdHunterConfig config) {
         catchingTraps.addAll(Rs2GameObject.getGameObjects(ObjectID.BIRD_SNARE_9373));
 
         List<GameObject> failedTraps = Rs2GameObject.getGameObjects(ObjectID.BIRD_SNARE);
+        failedTraps.addAll(Rs2GameObject.getGameObjects(9344));
```
gmason0: This is already here? ObjectID.BIRD_SNARE = 9344

runelite-client/src/main/java/net/runelite/client/plugins/microbot/zerozero/birdhunter/BirdHunterScript.java
```diff
@@ -121,6 +122,8 @@ private void handleTraps(BirdHunterConfig config) {
...
+        failedTraps.addAll(Rs2GameObject.getGameObjects(9346));
```
gmason0: We should be using the ObjectID class, so for example this should be ObjectID.BIRD_SNARE_9376

PR #660, opened 2025-01-24
Blocking Event Manager

gmason0:
I have created this tool that can be used to simply for the logic that is abstracted inside of the run() method of the Script.java class.

Inside of the Microbot class we initialize the BlockingEventManager, which inside of the constructor of the BlockingEventManager will add the events into the list.

From there, I have removed the majority of the logic inside of the run() method to be contained inside of these events that will be looped over.

It is required to synchronize the BlockingEventManager.class in-order to ensure that multiple threads to do not attempt to execute these events at the same time.

I figured it would be easier to just show you all the code instead of just sending the snippets you've already seen via Discord

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/Script.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/Script.java
@@ -175,7 +175,7 @@ public abstract class Script implements IScript {
                 if (!Microbot.getBlockingEventManager().getBlockingEvents().isEmpty()) {
                     for (BlockingEvent blockingEvent : Microbot.getBlockingEventManager().getBlockingEvents()) {
                         if (blockingEvent.validate()) {
-                            return blockingEvent.execute();
+                            return !blockingEvent.execute();
                         }
                     }
                 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/Script.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/Script.java
@@ -12,12 +12,10 @@ import net.runelite.client.plugins.microbot.util.Global;
 import net.runelite.client.plugins.microbot.util.inventory.Rs2Inventory;
 import net.runelite.client.plugins.microbot.util.keyboard.Rs2Keyboard;
 import net.runelite.client.plugins.microbot.util.player.Rs2Player;
-import net.runelite.client.plugins.microbot.util.settings.Rs2Settings;
 import net.runelite.client.plugins.microbot.util.tabs.Rs2Tab;
 import net.runelite.client.plugins.microbot.util.walker.Rs2Walker;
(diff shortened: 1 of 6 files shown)


FireStarter-Instigator:
"It is required to synchronize the BlockingEventManager.class in-order to ensure that multiple threads to do attempt to execute these events at the same time."

Typo? What mean

runelite-client/src/main/java/net/runelite/client/plugins/microbot/Script.java
FireStarter-Instigator, line 1: Cool and interesting idea makes a lot of sense

runelite-client/src/main/java/net/runelite/client/plugins/microbot/Script.java
```diff
@@ -169,32 +162,20 @@ public boolean run() {
...
-                synchronized (Rs2Widget.class) {
-                    if (!Microbot.getClientThread().runOnClientThread(clickHereToPlayButton::isHidden)) {
-                        Rs2Widget.clickWidget(clickHereToPlayButton.getId());
-
-                        sleepUntil(() -> Microbot.getClientThread().runOnClientThread(clickHereToPlayButton::isHidden), 10000);
+            synchronized (BlockingEventManager.class) {
+                if (!Microbot.getBlockingEventManager().getBlockingEvents().isEmpty()) {
+                    for (BlockingEvent blockingEvent : Microbot.getBlockingEventManager().getBlockingEvents()) {
+                        if (blockingEvent.validate()) {
+                            blockingEvent.execute();
```
chsami, line 178: They are not really blocking because the flow will continue to the bottom and return true in this method. IS this correct?

gmason0: Correct they dont block the logic, but they do process all the events before continuing with any of the logic below in the run method. I guess we could change this to return blockingEvent.execute(); ?

gmason0: actually maybe we make it return !blockingEvent.execute(); ? so that it returns false when scripts are calling

if(!super.run()) return;

chsami: > actually maybe we make it return !blockingEvent.execute(); ? so that it returns false when scripts are calling
> 
> if(!super.run()) return;

Yeah sounds good, once we have a blockingEvent, we should return false to avoid any script code running and messing with the blockingEvent code.

gmason0: Its been changed, I changed it to just explicit return false if an event gets executed

```
if (blockingEvent.validate()) {
    blockingEvent.execute();
    return false;
}
```

chsami (approved):

PR #663, opened 2025-01-24
Added break handling support for AutoHunter

Gage307:
The script will now take the last minute before break to:

Save current trap locations.
Dismantle the traps so they're not lost when we break.
When we're back from the break set the traps back down. Making sure nothing else is there first. 

Then it'll continue like normal.

Tested with 8 breaks in a row no issue.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/hunter/scripts/AutoChinScript.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/hunter/scripts/AutoChinScript.java
@@ -153,6 +153,14 @@ public class AutoChinScript extends Script {
 
     public void handleBreaks() {
          int secondsUntilBreak = BreakHandlerScript.breakIn; // Time until the break
+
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/microbot/hunter/scripts/AutoChinScript.java
```diff
@@ -127,4 +148,79 @@ private void handleLayingState(AutoHunterConfig config) {
         sleep(config.minSleepAfterLay(), config.maxSleepAfterLay());
         currentState = State.IDLE;
     }
+
+    public void handleBreaks() {
+        int secondsUntilBreak = BreakHandlerScript.breakIn; // Time until the break
```
gmason0: What if the break handler isn't turned on?

Gage307: It'll operate normally

chsami (approved):

PR #675, opened 2025-01-29
Fix DisableLevelUpInterfaceEvent + Add Piscatoris Fishing Colony transports & BankLocation

gmason0:
## DisableLevelUpInterfaceEvent
- Added a check for the report button, this will ensure that we are fully logged in as it seems that there is a small moment of time when checking `Microbot.getVarbitValue(Varbits.DISABLE_LEVEL_UP_INTERFACE)` that it will turn to 0 then back to 1 when at the play button.

## BankLocation
- Add Piscatoris Fishing Colony Bank

## ShortestPathPlugin
- Add Transports for Piscatoris Fishing Colony

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/bank/enums/BankLocation.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/bank/enums/BankLocation.java
@@ -72,6 +72,7 @@ public enum BankLocation {
     NEITIZNOT(new WorldPoint(2337, 3807, 0)),
     PEST_CONTROL(new WorldPoint(2667, 2653, 0)),
     PISCARILIUS(new WorldPoint(1803, 3790, 0)),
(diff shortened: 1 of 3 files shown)


gmason0:
Is Levelup Notifications Enabled: true
Disable LevelUp Notifications Value: 0
Is Report button Visible: false
Is Levelup Notifications Enabled: false
Disable LevelUp Notifications Value: 1
Is Report button Visible: true

```java
System.out.println("Is Levelup Notifications Enabled: " + Rs2Settings.isLevelUpNotificationsEnabled());
System.out.println("Disable LevelUp Notifications Value: " + Microbot.getVarbitValue(Varbits.DISABLE_LEVEL_UP_INTERFACE));
System.out.println("Is Report button Visible: " + Rs2Widget.isWidgetVisible(10616833));
```

chsami (approved):

PR #677, opened 2025-01-30
feat: prayer regeneration potion

chsami:


runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/player/Rs2Player.java
```diff
@@ -991,9 +995,10 @@ private static boolean usePotion(Integer ...itemIds) {
     private static boolean usePotion(String... itemNames) {
         if (Rs2Inventory.contains(x -> Arrays.stream(itemNames).anyMatch(name -> x.name.contains(name) && !x.isNoted()))) {
             Rs2Item potion = Rs2Inventory.get(Arrays.stream(itemNames).collect(Collectors.toList()),false);
-            if (potion != null) {
-                return Rs2Inventory.interact(potion, "drink");
-            }
+            if (potion == null) return false;
+            if (potion.getName().toLowerCase().contains("prayer regeneration potion") && hasPrayerRegenerationActive()) return false;
```
gmason0: I am not sure if we would want to implement this here, even thought this adds another method, I think it should just be implemented like so,

```
/**
 * Drink a Prayer Regeneration Potion.
 *
 * @return true if the prayer regeneration effect is active or the prayer regeneration potion was interacted with; 
 *         false otherwise.
 */
public static boolean drinkPrayerRegenerationPotion() {
    if (hasPrayerRegenerationActive()) {
        return true;
    }
    return usePotion(Rs2Potion.getPrayerRegenerationPotion());
}
```

Then we would just need to add this into AIO fighter's PotionManagerScript

```
if (config.usePrayerRegenerationPotion()) {
	if (Rs2Player.drinkPrayerRegenerationPotion())
		Rs2Player.waitForAnimation();
}
```

gmason0 (commented):
I think we should separate the prayer regeneration potion out into its own separate method

PR #685, opened 2025-02-06
alch spell location has changed

yairtu:
alch spell location has changed in the spell book, so I updated inventory slot location for alching

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/magic/aiomagic/scripts/AlchScript.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/magic/aiomagic/scripts/AlchScript.java
@@ -78,7 +78,7 @@ public class AlchScript extends Script {
                             Microbot.log("Missing alch items...");
                             return;
                         }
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/microbot/magic/aiomagic/scripts/AlchScript.java
```diff
@@ -78,7 +78,7 @@ public boolean run() {
                             Microbot.log("Missing alch items...");
                             return;
                         }
-                        int inventorySlot = Rs2Player.getRealSkillLevel(Skill.MAGIC) >= 55 ? 11 : 4;
+                        int inventorySlot = Rs2Player.getRealSkillLevel(Skill.MAGIC) >= 55 ? 4 : 4;
```
gmason0, line 81: Are you sure this isn't just how you have your spellbook re-ordered? For me on default settings, the previous 11 : 4 seems correct to me.

image

yairtu: After the last update my spells got reordered, thought it was the update but it was the RL plugin, reset it and its back to normal. good catch

PR #689, opened 2025-02-09
Option to drop ores when using autominer

Aholicknight:
This pull request includes a new feature to the `AutoMining` plugin. The new change is auto drop all ores once the inventory is full. By default, it would drop the entire inventory except the pickaxe.

### New Feature Addition:
* Added a new configuration option `dropOres` to the `AutoMiningConfig` interface, which allows users to enable or disable the automatic dropping of ores when the inventory is full. The default setting for this option is disabled.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/mining/AutoMiningScript.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/mining/AutoMiningScript.java
@@ -30,7 +30,7 @@ enum State {
 
 public class AutoMiningScript extends Script {
 
-    public static final String version = "1.4.3";
+    public static final String version = "1.4.4";
     private static final int GEM_MINE_UNDERGROUND = 11410;
     private static final int BASALT_MINE = 11425;
     State state = State.MINING;
@@ -96,6 +96,8 @@ public class AutoMiningScript extends Script {
                                     return;
                             }
 
+                        } else if (config.dropOres()) {
+                            Rs2Inventory.dropAll(config.ORE().getName());
                         } else {
                             Rs2Inventory.dropAllExcept("pickaxe");
                         }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/mining/AutoMiningScript.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/mining/AutoMiningScript.java
@@ -98,6 +98,10 @@ public class AutoMiningScript extends Script {
 
                         } else if (config.dropOres()) {
                             Rs2Inventory.dropAll(config.ORE().getName());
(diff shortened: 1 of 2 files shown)


gmason0:
@Aholicknight - We do not merge directly into main, please change this PR to merge into the development branch

Aholicknight:
> @Aholicknight - We do not merge directly into main, please change this PR to merge into the development branch

@g-mason0 changed

runelite-client/src/main/java/net/runelite/client/plugins/microbot/mining/AutoMiningScript.java
```diff
@@ -96,6 +96,8 @@ public boolean run(AutoMiningConfig config) {
                                     return;
                             }
 
+                        } else if (config.dropOres()) {
+                            Rs2Inventory.dropAll(config.ORE().getName());
```
gmason0, line 100: Wont the script eventually get stuck from receiving gems & clues in the inventory?

gmason0 (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/mining/AutoMiningScript.java
Aholicknight: What do you mean? I tried it on my account, and it didn't. There is also another plugin you can use to dismiss random events.

gmason0: While mining ore, there is a 1/256 chance to receive a gem when mining ore, if this is left for long periods of time, wouldn't the inventory eventually fill up with gems? Shouldn't we do something about that?

Aholicknight: I'll update the code to fix this.

Aholicknight: I don't think so. If I added a check for this, what if the inventory gets filled with gems? Then what?

gmason0: I would assume we would bank them? I'm not sure what else we would do with them considering if we disable the useBank feature, it will drop all items in the inventory except the pickaxe

gmason0: Considering we can't align on the design idea, I have implemented the feature here: https://github.com/chsami/Microbot/commit/0fc483e602f7656a3235b87ee4bc91bf4eac7918

Have a great evening!

PR #697, opened 2025-02-13
 Feature - Added Herb and Gem Bag to AIO Fighter and Rs2Banking

xdorito:
Added Herb Bag and Gem Bag Support to AIO Fighter

Added emptyHerbBag to Rs2Banker


xdorito:
Closing pending inventory setups rework.

PR #699, opened 2025-02-14
AIO Fighter fixes & New QoL functionality 

See1Duck:
# Changelog

---

## AIOFighterPlugin
**Improve combat, potion, and loot handling.**  
- Refactored the plugin to streamline potion consumption into a unified system, removing manual toggles for individual potions.  
- Updated loot logic to dynamically determine minimum free inventory slots, enhancing flexibility while looting.  
- Simplified state checks for broader compatibility across banking, walking, and combat scenarios.

**Improve banking logic and walking behavior.**  
- Enhanced the banking process by ensuring the bank interface is closed after transactions for better flow.  
- Updated walking logic to handle edge cases where center location has invalid coordinates (e.g., `[0, 0, 0]`).  
- Improved state transitions by switching to `IDLE` upon successful navigation.  

These changes result in more reliable bot behavior and fewer interruptions during automated activities.

### Highlights
- **Consolidated potion management**: Ensures consistent usage of antidotes, combat potions, and prayer potions without separate config toggles.  
- **Dynamic inventory calculations**: Replaced hardcoded slot checks with a system tied to banking preferences.  
- **New slayer mode configurations**: Added hidden features to prepare for future Slayer-specific functionality.

---

## CraftingManager
**Add quick crafting for dragonhide.**  
- Introduced a "Quick Craft" menu entry when using a needle on leather, automating interactions and reducing manual input.  

## GemCuttingManager
**Add quick gem cutting functionality.**  
- Enabled automated gem cutting through a "Cut Gem" menu entry when using a chisel.  
- Supports level-based gem selection, speeding up crafting operations while ensuring only valid gems are processed.

## UncutGems & DragonhideCrafting
**Add crafting and cutting enum support.**  
- Created enums for uncut gems and dragonhide items, streamlining retrieval of craftable/cutable items based on player skill levels.  
- Improves inventory management logic and overall code organization.

## QoLPlugin
**Extend support for crafting and alchemy features.**  
- Enhanced "Quick High Alch" functionality by adding a menu entry for high-alching profitable items directly from the inventory.  
- Registered both `CraftingManager` and `GemCuttingManager` for a more seamless experience with crafting tasks.

## WintertodtScript
**Improve warmth management and food usage.**  
- Added a function to monitor the player's warmth during Wintertodt.  
- Automatically consumes rejuvenation potions when below a certain threshold to improve survival and efficiency.

---

## AIOFighterPlugin
**Remove unused potion-handling scripts.**  
- Deleted `AntiPoisonScript` and `CombatPotionScript` as they were outdated and no longer in use.  
- Reduced code complexity and redundant logic for better maintainability.

---

## PrayerPotionScript
**Fix unnecessary prayer potion check removal.**  
- Removed the `togglePrayerPotions` configuration check, which caused failures in potion usage logic.  
- Ensures correct handling of prayer potion consumption based solely on prayer level thresholds.


runelite-client/src/main/java/net/runelite/client/plugins/microbot/aiofighter/combat/PotionManagerScript.java
```diff
@@ -16,44 +16,54 @@ public boolean run(AIOFighterConfig config) {
...
-            } catch(Exception ex) {
+            } catch (Exception ex) {
                 System.out.println(ex.getMessage());
             }
         }, 0, 600, TimeUnit.MILLISECONDS);
         return true;
     }
+
+    public int getPrayerPotionThreshold() {
+        int maxRestore = 7 + (Rs2Player.getRealSkillLevel(Skill.PRAYER) / 4);
```
gmason0: Prayer potions follow this math for the threshold, but super restores use the following

image

Then Moonlight Potions are the following:

image

Possibly we manage these sort of checks inside of a Rs2Player.drinkPrayerPotion to determine the threshold based on what is in the inventory & then drink the potion using their respective threshold

gmason0 (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/aiofighter/combat/PotionManagerScript.java
gmason0: I did make this as apart of the suggestion that was created yesterday, I am not sure if we want to leverage this or something else you had in mind?

```java
public static boolean drinkPrayerPotion() {
    int maxPrayer = getRealSkillLevel(Skill.PRAYER);
    int maxHerblore = getRealSkillLevel(Skill.HERBLORE);
    int restoreAmount;

    if (hasPotion("moonlight moth mix")) {
        restoreAmount = 22;
    } else if (hasPotion("moonlight potion")) {
        int prayerRestore = (maxPrayer / 4) + 7;
        int herbloreRestore = (int) Math.floor((maxHerblore * 3.0 / 10.0)) + 7;
        restoreAmount = Math.max(prayerRestore, herbloreRestore);
    } else if (hasPotion("super restore") || hasPotion("blighted super restore")) {
        restoreAmount = (maxPrayer / 4) + 8;
    } else {
        restoreAmount = (maxPrayer / 4) + 7;
    }

    int threshold = maxPrayer - restoreAmount;
    int randomizedThreshold = Rs2Random.randomGaussian(threshold - 5, 3);
    randomizedThreshold = Math.min(randomizedThreshold, threshold);
    System.out.println("Threshold: " + randomizedThreshold);

    return drinkPrayerPotionAt(randomizedThreshold);
}
```

See1Duck: Yeah i get the point and it would be better to move it to the main method instead. The only issue i see it that if you have multiple different potions in the inv or if some are noted as a drop resulting in the wrong calc. not sure how we can get around that

See1Duck: Lets go with your approach, ill implement that instead in my pr instead

gmason0: Sounds good, I also will probably refactor this to be based on item order in the inventory but i think now having priorities is alright as that is how the current  drinkPrayerPotionAt functions

PR #703, opened 2025-02-15
Roguesden add lower level

acun9323:
### Added support for lower level Rogues Den by extending the original plugin, which had only support for +80 thieving.
- Based on player's thieving level it decide which maze path to take. 
- Add overlay for "Maze runs" completed counter
- Changed skill requirement to 50 thieving & agility

### Note
The original code was not easily open for extension, so I have added  `originalObstacles` and some filtering to make it work. I am not happy about it but it does the job.

### Highlight
Unfortunately, I have no account to test if level 80 thieving path still works


acun9323:
image
It got tested now, seems working well

chsami (approved):

PR #710, opened 2025-02-20
Updating thieving script to work with Rs2Model

yairtu:
Updating thieving script to work with Rs2Model


gmason0:
Fixed in 6b5eeed0

PR #712, opened 2025-02-21
Update Rs2Reflection.java

MortezaPourmoghaddam:
update reflection values


gmason0:
1. Can you please keep the comments of where these values can be found? Think that it could be useful when they do eventually need to be updated with each rev.
2. Change the base branch of this PR to development, we do not merge directly from main.

PR #726, opened 2025-03-01
Make targetmenu accessible

Gaisberg:


runelite-client/src/main/java/net/runelite/client/plugins/microbot/Microbot.java
```diff
@@ -77,6 +78,16 @@ public class Microbot {
     public static String status = "IDLE";
     public static boolean enableAutoRunOn = true;
     public static boolean useStaminaPotsIfNeeded = true;
+    public interface MenuOptionClickedHandler {
+        void onMenuOptionClicked(MenuOptionClicked event);
+    }
+    public static final List<MenuOptionClickedHandler> menuOptionClickedHandlers = new ArrayList<>();
+    public static void registerMenuOptionClickedHandler(MenuOptionClickedHandler callback) {
```
gmason0: I think these methods deserve their own class, look at how the blocking event manager is implemented.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/bank/Rs2Bank.java
```diff
@@ -146,7 +147,7 @@ public static List<Rs2ItemModel> bankItems() {
      */
     public static boolean closeBank() {
         if (!isOpen()) return false;
-        Rs2Widget.clickChildWidget(786434, 11);
+        Microbot.doInvoke(new NewMenuEntry("Close", "", 1, MenuAction.CC_OP, 11, 786434, false), new Rectangle(1, 1));
```
gmason0: Update this from Rectangle 1,1,  we should be using the widget bounds.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/bank/Rs2Bank.java
```diff
@@ -703,6 +717,8 @@ private static void withdrawOne(Rs2ItemModel rs2Item) {
         } else {
             invokeMenu(2, rs2Item);
         }
+
+        sleepUntil(() -> Rs2Inventory.hasItem(rs2Item.id), 2500);
```
gmason0: Couldn't we just leverage Rs2Inventory.waitForInventoryChanges?

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/bank/Rs2Bank.java
```diff
@@ -992,6 +1008,7 @@ private static void wearItem(Rs2ItemModel rs2Item) {
         container = BANK_INVENTORY_ITEM_CONTAINER;
 
         invokeMenu(8, rs2Item);
+        sleepUntil(() -> !Rs2Inventory.hasItem(rs2Item.id), 1800);
```
gmason0: Couldn't we just leverage Rs2Inventory.waitForInventoryChanges?

gmason0 (changes requested):
I think we need to implement the MenuOptionClickedHandler by creating an instance of it inside of Microbot.java instead of defining methods & interfaces directly in the Microbot class.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/bank/Rs2Bank.java
```diff
@@ -146,7 +147,12 @@ public static List<Rs2ItemModel> bankItems() {
      */
     public static boolean closeBank() {
         if (!isOpen()) return false;
-        Rs2Widget.clickChildWidget(786434, 11);
+        Widget closeWidget = Rs2Widget.getWidget(786434, 11);
```
gmason0, line 150: Using this still returns null for closeWidget as this method is used for returning widgets based on their parent & child id for example, 12,1. clickChildWidget was using widget.getChild() which is different than the Parent/Child ID system

image

image

Using Rs2Widget findWidget would allow you to search from all child types

`Widget closeWidget = Rs2Widget.findWidget("Close", List.of(Rs2Widget.getWidget(786434)));`

image

gmason0 (changes requested):
Lets fix this last method, but everything else looks good

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/bank/Rs2Bank.java
Gaisberg: Im actually not sure if this is even needed anymore. I decided to go another route:

https://github.com/chsami/Microbot/pull/732

PR #732, opened 2025-03-04
GB OneClicker

Gaisberg:
- NMZ heavily inspired by existing NMZ script.
- Features manual clicking, auto-clicker and a basic NMZ script.


runelite-client/src/main/java/net/runelite/client/plugins/microbot/gboc/scripts/HerbRunScript.java
```diff
@@ -0,0 +1,350 @@
+//package net.runelite.client.plugins.microbot.gboc.scripts;
```
gmason0, line 1: Can we just not include this if you're going to comment out the entire class?

gmason0 (changes requested):
Lets just remove HerbRunScript until its completed

gmason0:
Speaking with some of the other developers, we aren't going to accept this PR. The current script ( NMZ ) is already implemented in our client, feel free to make adjustments to the existing plugin!

PR #738, opened 2025-03-06
Added Buy Quest Items Requirements to GE Buyer

subinznz:
Auto buys quest items that are shows in requirements


acun9323:
borat

acun9323:
I wonder what happens when "item requirements" section of Quest Helper contains untradeable items. I've never tested the script with untradeables, will it get stuck and repeatedly try to buy or just skips it any idea?  

For example The slug menance quest requires 1x Commorb

subinznz:
I only tested a few quests , I added a filter to ignore coins since you can't buy it in GE. when I saw this problem it just hanged there when it couldn't find. So yeah it's just in beta. I couldn't figure out how to get the untradeable info out. Not sure how to make it skip , suggestions are welcome.

acun9323:
We can add something like "when item was not found, skip to next" think this will keep it simple and make it work

Edit: #740

jst94 (approved):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/autoBuyer/AutoBuyerScript.java
```diff
@@ -24,8 +26,20 @@ public class AutoBuyerScript extends Script {
     public boolean run(AutoBuyerConfig config) {
 
         Microbot.enableAutoRunOn = false;
+        String listOfItemsToBuy;
+        if (getQuestHelperPlugin().getSelectedQuest() != null && config.buyQuest()) {
+            listOfItemsToBuy = getQuestHelperPlugin().getSelectedQuest().getItemRequirements()
+                    .stream()
+                    .filter(item -> !item.getName().equalsIgnoreCase("coins")) // Ignore items with name "coins"
+                    .map(item -> item.getName() + "[" + item.getQuantity() + "]") // Format: "Name (Quantity)"
+                    .collect(Collectors.joining(",")); // Joins names with ", "
```
gmason0: I think we should add a filter for items that aren't in the bank.

gmason0 (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/autoBuyer/AutoBuyerScript.java
subinznz: something like this ?

findBankItem(String name, boolean exact)

gmason0: .filter(item -> !Rs2Bank.hasBankItem(item.getName(), item.getQuantity(), true))

subinznz: added checks

subinznz (commented):
added the bank item changes and also added handling of unnnoted

acun9323:
1. I've added item skipping logic, but I cannot commit it to this PR so will create new one (#740)
2. ❌ Player must have opened his bank manually before running the script, otherwise item's in bank are not recognized. IDK if there is a Microbot API to check if bank was opened before, otherwise we can make GE buyer open and close.
3. image

Script can't handle items above:
- ❌ 1x Feathers ✅ 1x Feather
- ❌ 1x Attack potion (2) ✅ 1x Attack potion(2) <- No space allowed here
- Any axe & Any bow are not indicated red (they exist in bank) but GE buyer tries to buy these from GE.

I believe this PR is acceptable as 'Beta' and we can improve it to handle edge cases in the future

See1Duck:
# Example

Here is an example that demonstrates how to retrieve only the missing items that are tradeable and get the correct item names.  It checks if a quest is active, retrieves a list of item IDs to process, filters them for tradeable items, and then formats the item names with their required quantities.

## Key Components

1. **Filtering Tradeable Items**:  
   If a quest is active, it retrieves a list of item IDs via the quest helper. The list is processed to remove duplicates, filter only the tradeable items (using the `isTradeable` method), and limit the list to 250 entries.

2. **Formatting the Output**:  
   The filtered list is used to retrieve the item requirements. For each requirement, the item name (obtained using `getItemName`) and quantity are concatenated into a string with the format `"ItemName[Quantity]"`, and then the list is joined with commas.

3. **Helper Methods**:  
   - **`getItemName(int itemId)`**: Retrieves the name of an item.
   - **`getItemRequirements()`**: Gets the list of item requirements from the selected quest.
   - **`isTradeable(int itemId)`**: Checks if an item is tradeable.
   - **`getQuestHelperPlugin()`**: Finds the quest helper plugin instance.


## Code Listing

```java
public class ExampleScript extends Script {

    public boolean run(ExampleConfig config) {
        Microbot.enableAutoRunOn = false;

            try {
                if (!super.run()) return;
                long startTime = System.currentTimeMillis();

                String listOfItemsToBuy;
                if (getQuestHelperPlugin().getSelectedQuest() != null) {
                    final List<Integer> idsList = Microbot.getClientThread().runOnClientThread(getQuestHelperPlugin()::itemsToTag);

                    final Set<Integer> ids = idsList.stream()
                            .distinct()
                            .filter(this::isTradeable)
                            .limit(250)
                            .collect(Collectors.toCollection(TreeSet::new));

                    listOfItemsToBuy = getItemRequirements()
                            .stream()
                            .filter(item -> ids.contains(item.getId())) // Filter items based on the ids list
                            .map(item -> getItemName(item.getId()) + "[" + item.getQuantity() + "]") // Format: "Name[Quantity]"
                            .collect(Collectors.joining(",")); // Joins names with ", "

                    Microbot.log("Using quest helper item list: " + listOfItemsToBuy);

                } else {
                    Microbot.log("Using manual item list");
                }

                long endTime = System.currentTimeMillis();
                long totalTime = endTime - startTime;
                System.out.println("Total time for loop " + totalTime);
            } catch (Exception ex) {
                System.out.println(ex.getMessage());
            }
        }, 0, 1000, TimeUnit.MILLISECONDS);
        return true;
    }

    public String getItemName(int itemId) {
        ItemComposition item = Microbot.getClientThread().runOnClientThread(() -> Microbot.getItemManager().getItemComposition(itemId));
        return item.getName();
    }

    public List<ItemRequirement> getItemRequirements() {
        return Microbot.getClientThread().runOnClientThread(() -> getQuestHelperPlugin().getSelectedQuest().getItemRequirements());
    }

    public boolean isTradeable(int itemId) {
        ItemComposition item = Microbot.getClientThread().runOnClientThread(() -> Microbot.getItemManager().getItemComposition(itemId));
        return item.isTradeable();
    }

    protected QuestHelperPlugin getQuestHelperPlugin() {
        return (QuestHelperPlugin) Microbot.getPluginManager().getPlugins().stream().filter(x -> x instanceof QuestHelperPlugin).findFirst().orElse(null);
    }

    @Override
    public void shutdown() {
        super.shutdown();
    }
}
```

runelite-client/src/main/java/net/runelite/client/plugins/microbot/autoBuyer/AutoBuyerConfig.java
```diff
@@ -9,25 +9,44 @@
 @ConfigInformation("Acun's GE buyer <br><br> Start near GE. <br> Make sure to spell item names correct. " +
         "Collects items to bank. " +
         "<br> Be cautious and monitor when using, because there are no failsafes added yet. " +
-        "<br><br> Item name[quantity] for example: rune arrow[50],amulet of glory(6)[1]"
+        "<br><br> Item name[quantity] for example: rune arrow[50],amulet of glory(6)[1]"+
+        "<br><br>If Buy Quest Items selected then buy List will be ignored. Ensure Quest Helper has quest selected before starting."
 )
 public interface AutoBuyerConfig extends Config {
+
+    @ConfigItem(
```
chsami, line 17: I'm not sure if this should be added here. It seems more logical to add this feature to the mquester

subinznz: MQuester code was a bit complex for me wasn’t really sure where to put this. I also didn’t want to rebuild the GE Logic , i do agree ideally if there is missing items it would be good to walk to ge and buy , maybe activate the GE buyer once it reaches there.

subinznz (commented):
fixed item handling it is more clean now

acun9323:
In case of `Any axe` how does it handle that one? Does it buy the cheapest axe if it's missing?

subinznz:
> In case of `Any axe` how does it handle that one? Does it buy the cheapest axe if it's missing?

Seems to work for Lost City it buys bronze axe, but on other quests it doesn't because it refers to an ItemCollection which points to the first item on the list which is crystal axe which is non-tradeable .  

Seems to pick the first item on the ItemCollections e.g.


```
public enum ItemCollections {
    // Tools
    AXES("Axe", ImmutableList.of(
            ItemID.CRYSTAL_AXE,
            ItemID.CRYSTAL_FELLING_AXE,
            ItemID._3RD_AGE_AXE,
            ItemID.ECHO_AXE,
            ItemID.INFERNAL_AXE_OR,
            ItemID.INFERNAL_AXE,
            ItemID.INFERNAL_AXE_UNCHARGED_25371,
            ItemID.INFERNAL_AXE_UNCHARGED,
            ItemID.DRAGON_AXE_OR,
            ItemID.DRAGON_AXE,
            ItemID.DRAGON_FELLING_AXE,
            ItemID.RUNE_AXE,
            ItemID.RUNE_FELLING_AXE,
            ItemID.GILDED_AXE,
            ItemID.ADAMANT_AXE,
            ItemID.ADAMANT_FELLING_AXE,
            ItemID.MITHRIL_AXE,
            ItemID.MITHRIL_FELLING_AXE,
            ItemID.BLACK_AXE,
            ItemID.BLACK_FELLING_AXE,
            ItemID.STEEL_AXE,
            ItemID.STEEL_FELLING_AXE,
            ItemID.IRON_AXE,
            ItemID.IRON_FELLING_AXE,
            ItemID.BRONZE_AXE,
            ItemID.BRONZE_FELLING_AXE
    )),
    


```

 If I change the list so that BRONZE_AXE is first it works perfect, just not sure which road to take here because it tags the first item in the collection .

chsami (approved):

PR #745, opened 2025-03-09
Lunar Altar RC script

KevinFC93:
Change ZMI script for Lunar altar


gmason0:
You should create your own plugin instead of demolishing the existing zmi plugin..

KevinFC93:
Thanks G - first time programming not sure how git hub and pulling works. Thought this would just be another plug in instead of replacing zmi plug in.

PR #751, opened 2025-03-13
shuts down WT script if player dies

crispycritter99:
If you die to the cold in WT, the script should now shut off instead of continuing to try to play.


gmason0 (changes requested):
Fix your building issue then good to merge!

crispycritter99:
build issue fixed

PR #755, opened 2025-03-14
Feature/hunter pyrefox plugin

BarryPooter:
Simple plugin which catches pyrefox for its meat (and maybe low level hunter xp).

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/kaas/pyrefox/managers/PyreFoxScript.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/kaas/pyrefox/managers/PyreFoxScript.java
@@ -1,0 +1,349 @@
+package net.runelite.client.plugins.microbot.kaas.pyrefox.managers;
+
+import net.runelite.api.GameObject;
+import net.runelite.api.ItemID;
+import net.runelite.api.Skill;
+import net.runelite.api.coords.WorldPoint;
+import net.runelite.client.plugins.microbot.Microbot;
+import net.runelite.client.plugins.microbot.Script;
+import net.runelite.client.plugins.microbot.kaas.pyrefox.PyreFoxConfig;
+import net.runelite.client.plugins.microbot.kaas.pyrefox.PyreFoxConstants;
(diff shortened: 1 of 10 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/microbot/kaas/pyrefox/managers/PyreFoxScript.java
```diff
@@ -0,0 +1,351 @@
...
+
+    /**
+     * What should happen when we reach 25 hitpoints?
+     */
+    private void _handleGatheringLogs()
+    {
+        // Find nearby tree.
+        // Deliberate while loop, I want the gamestate to not influence the chopping process.
+        while (Rs2Inventory.count("Logs") < PyreFoxConstants.GATHER_LOGS_AMOUNT)
+        {
```
gmason0, line 110: Inside of these while loops, its advised to add if (!isRunning()) break;

Without this, if your script is shutdown by the user via the plugin panel it will continue this while loop until the condition is met.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/kaas/pyrefox/managers/PyreFoxScript.java
```diff
@@ -0,0 +1,351 @@
...
+
+        // Deposit all non-locked inventory items.
+        Rs2Bank.depositAll();
+        sleep(50, 1200);
+
+        // Auto-eat until full.
+        if (_config.AutoEat())
+        {
+            while (!Rs2Player.isFullHealth())
+            {
```
gmason0, line 311: Same thing here as well, we should be adding if (!isRunning()) break;

gmason0 (changes requested):
Lets make these small adjustments to the while loop, otherwise good to merge

runelite-client/src/main/java/net/runelite/client/plugins/microbot/kaas/pyrefox/managers/PyreFoxScript.java
BarryPooter: Good catch! Pushed the fix

gmason0 (approved):

PR #759, opened 2025-03-17
Refactor Script.java

Gaisberg:
- Make item prerequisite in base script
  - Move sleeps to global.java and make them interruptable. Rename redundand sleepUntilTrue to sleepUntil
  - Make birdhouse script less "botty" and use the new prerequisite logic
  - Fix wrong default worldview in NewMenuEntry


Gaisberg:
I noticed I removed the sleepUntil with reset, other than that its OK. Will fix later...

Gaisberg:
Will be splitting this into smaller PRs...

PR #765, opened 2025-03-19
Add prifddinas east bank

crispycritter99:
rename the previous prifdinnas bank to prifdinnas west (there are two banks in the city)


crispycritter99:
gonna make the PRs nicer before i reopen

PR #766, opened 2025-03-19
Add Prifddinas furnace location to jewelry crafting script

crispycritter99:
pretty close to a bank


crispycritter99:
its not passing the check because it references BankLocation.Prifddinas_East which is new from my other PR. I can combine the PRs

PR #769, opened 2025-03-19
Refactor the sleeping methods to use a common builder

Gaisberg:


runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/Global.java
```diff
@@ -22,167 +23,399 @@ public static ScheduledFuture<?> awaitExecutionUntil(Runnable callback, BooleanS
         return scheduledFuture;
     }
 
-    public static void sleep(int start) {
-        if (Microbot.getClient().isClientThread()) return;
-        try {
-            Thread.sleep(start);
-        } catch (InterruptedException e) {
-            System.out.println(e.getMessage());
+    public static class SleepBuilder {
```
chsami, line 26: This class should be seperated from the global.java

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/Global.java
```diff
@@ -22,167 +23,399 @@ public static ScheduledFuture<?> awaitExecutionUntil(Runnable callback, BooleanS
...
+        /**
+         * Executes the sleep operation with all configured parameters.
+         *
+         * @param <T> The return type, either Boolean for condition-based waits or the result type for untilNotNull
+         * @return For condition-based waits: Boolean.TRUE if the condition was met, Boolean.FALSE if timed out
+         * For untilNotNull: The non-null result if found, null if timed out
+         */
+        @SuppressWarnings("unchecked")
+        public <T> T execute() {
+            if (isSimpleSleep) {
```
chsami, line 200: To avoid using multiple boolean flags for each builder step, you can use a list of `BuilderStep` objects, each representing a step in the builder process. This approach will help you encapsulate the logic of each step within its own class, making the `execute` method cleaner and more maintainable.

Here's an example of the concept generated by github copilot

1. Define an interface or abstract class for the `BuilderStep`:

```java
public interface BuilderStep {
    void execute();
}
```

2. Implement different steps as classes that implement the `BuilderStep` interface:

```java
public class UntilNotNull implements BuilderStep {
    @Override
    public void execute() {
        // Logic for step one
        System.out.println("Executing step one");
    }
}

public class StepTwo implements BuilderStep {
    @Override
    public void execute() {
        // Logic for step two
        System.out.println("Executing step two");
    }
}

// Add more steps as needed
```

3. Use a list to manage and execute the steps:

```java
import java.util.ArrayList;
import java.util.List;

public class Builder {
    private List<BuilderStep> steps = new ArrayList<>();

    public Builder addStep(BuilderStep step) {
        steps.add(step);
        return this;
    }

    public void execute() {
        for (BuilderStep step : steps) {
            step.execute();
        }
    }

    public static void main(String[] args) {
        Builder builder = new Builder();
        builder.addStep(new StepOne())
               .addStep(new StepTwo())
               // Add more steps as needed
               .execute();
    }
}
```

With this approach, you can easily add, remove, or modify steps without changing the logic in the `execute` method. Each step's logic is encapsulated within its own class, promoting better separation of concerns and easier maintenance.

chsami: I appreciate the work you've put into this. However, I’m concerned that the changes to our existing code might introduce unexpected issues. Can we revert those changes for now and take a more gradual, step-by-step approach when integrating the new sleeper? This way, we can catch any potential bugs early..

chsami (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/Global.java
Gaisberg: Sure, lets look at this another time...

PR #771, opened 2025-03-20
feat: enhance script framework and add new plugins

Gaisberg:
This commit includes several significant improvements:
- Refactor Script class to support prerequisites handling and better script lifecycle management
- Add new plugins: DailyTasks and ScriptScheduler for automated task management
- Improve HerbRun plugin with better patch detection and inventory setup integration
- Fix plugin start/stop mechanism to use separate thread instead of SwingUtilities
- Add support for inventory setup configuration across multiple plugins
- Implement better error handling and script shutdown procedures


runelite-client/src/main/java/net/runelite/client/plugins/microbot/Script.java
FireStarter-Instigator, line 1: why deleted all the sleepuntil's?

Gaisberg: I moved them to global and Script extends that now

runelite-client/src/main/java/net/runelite/client/plugins/microbot/Script.java
```diff
@@ -57,109 +66,84 @@ public Duration getRunTime() {
...
         try {
-            while (System.nanoTime() - startTime < timeoutNanos) {
-                action.run();
-                if (awaitedCondition.getAsBoolean()) {
-                    return true;
+            List<Plugin> plugins = new ArrayList<>(Microbot.getPluginManager().getPlugins());
+            for (Plugin plugin : plugins) {
+                Field[] fields = plugin.getClass().getDeclaredFields();
+                for (Field field : fields) {
+                    field.setAccessible(true);
```
gmason0, line 93: I am not sure how I feel about using reflection here.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/Script.java
```diff
@@ -57,109 +66,84 @@ public Duration getRunTime() {
...
+            if (Rs2Bank.isOpen()) Rs2Bank.closeBank();
         }
-        return awaitedCondition.getAsBoolean();
-    }
 
-    public void sleepUntilOnClientThread(BooleanSupplier awaitedCondition) {
-        sleepUntilOnClientThread(awaitedCondition, 5000);
-    }
+        if (startingLocation != null) {
+            boolean arrived = Rs2Walker.walkTo(startingLocation);
```
gmason0, line 134: How would something like this work for AIO Fighter being scheduled, where you can start the plugin where ever?

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/Global.java
```diff
@@ -67,28 +55,25 @@ public static <T> T sleepUntilNotNull(Callable<T> method, int time) {
...
     }
 
+    public static boolean sleepUntil(BooleanSupplier awaitedCondition) {
+        return sleepUntil(awaitedCondition, 5000);
+    }
 
     public static boolean sleepUntil(BooleanSupplier awaitedCondition, int time) {
         if (Microbot.getClient().isClientThread()) return false;
-        boolean done;
+        boolean done = false;
```
gmason0, line 64: booleans are initialized as false

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/Rs2InventorySetup.java
```diff
@@ -25,20 +27,20 @@
  * the correct items are equipped and in the inventory.
  */
 public class Rs2InventorySetup {
-
+    @Getter
     InventorySetup inventorySetup;
 
-    ScheduledFuture<?> _mainScheduler;
+    Script _runningScript;
```
gmason0, line 33: Like I had mentioned to you, I dont think we should be passing the entire script object here. Considering that scripts can contain multiple threads, or scheduled futures. I think that it would still be most ideal that we keep this to only passing the mainScheduledFuture which typically contains the root logic for the scripts.

gmason0 (changes requested):
I am liking what I see thus far but there is still a few concerns I have this system. 

1. How does the Scheduler understand Script State? Considering our scripts are considered 'state machines', how does the scheduler know when its ideal to switch tasks when we aren't in the middle of something?

2. How can we handle scripts that don't typically have a 'start location' & that they can just be started any where, AIO Fighter & Bank Stander come to mind.

3. The Inventory Setups changes, I do think that we should only be passing mainScheduledFuture instead of passing the entire script object.

4. Lets remove the other new plugins that you have within this PR & try to keep it solely about the scheduler considering this in itself already a large feature.

Can you also include some snippets of what it would take to make a script able to be scheduled?

@chsami - anything you think I am not thinking about here?

runelite-client/src/main/java/net/runelite/client/plugins/microbot/scriptscheduler/ScriptSchedulerPlugin.java
```diff
@@ -0,0 +1,286 @@
...
+    @Subscribe
+    public void onGameStateChanged(GameStateChanged event) {
+        if (event.getGameState() == GameState.LOGIN_SCREEN) {
+            stopCurrentScript();
+        }
+    }
+
+    @Subscribe
+    public void onPluginChanged(PluginChanged event) {
+        if (currentScript != null && event.getPlugin() == currentScript.getPlugin() && !currentScript.isRunning()) {
```
chsami, line 196: This code will be used to detect if a plugin has been shutdown, we could move on to the next plugin if this happened

Gaisberg:
As per comments and discussion in discord the scheduler will only start and stop plugins and will not care about script states. The plugins should handle pre and post logic. Lets continue discussion about the scheduler here: https://github.com/chsami/Microbot/pull/775

runelite-client/src/main/java/net/runelite/client/plugins/microbot/Script.java
Gaisberg: This will be removed in the upcoming implementation.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/Script.java
Gaisberg: That is up to the plugin the scheduler will not take any part in handling pre / post logic as per discussion.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/Global.java
Gaisberg: Oh thats neat, intellij was whining about this.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/Rs2InventorySetup.java
Gaisberg: This will be removed in the upcoming implementation.

PR #774, opened 2025-03-22
fix: add thread interception to sleep methods

Gaisberg:
- fixes issues where sleep is blocking thread from shutting down


FireStarter-Instigator:
wow interesting why is this neccesary?

Gaisberg:
> wow interesting why is this neccesary?

DRY and ensure sleeps aren't locking threads.

chsami (approved):

PR #775, opened 2025-03-22
Plugin scheduler

Gaisberg:
Plugin Scheduler


runelite-client/src/main/java/net/runelite/client/plugins/microbot/pluginscheduler/SchedulerPlugin.java
```diff
@@ -0,0 +1,285 @@
...
+                log.error("Failed to stop plugin: " + currentPlugin.getCleanName());
+            }
+        }
+        updatePanels();
+    }
+
+    @Subscribe
+    public void onGameStateChanged(GameStateChanged event) {
+        if (event.getGameState() == GameState.LOGIN_SCREEN) {
+            stopCurrentPlugin();
```
chsami: This might mess with breakhandler? If a script takes a break we will stop the plugin.

.run/Runelite with proxy.run.xml
chsami, line 1: can you put these files back to their original? You're not supposed to make any changes to these files

.run/Runelite.run.xml
chsami, line 1: can you put these files back to their original? You're not supposed to make any changes to these files

chsami (changes requested):

chsami (approved):

PR #776, opened 2025-03-22
refactor: herb runner

Gaisberg:
now takes into consideration available herb patches and uses inventorysetups


runelite-client/src/main/java/net/runelite/client/plugins/microbot/herbrun/HerbrunConfig.java
```diff
@@ -0,0 +1,163 @@
...
+    String settingsSection = "Settings";
+
+    @ConfigItem(
+            keyName = "inventorySetup",
+            name = "InventorySetup Name",
+            description = "Name of inventory setup to use",
+            position = 1,
+            section = settingsSection
+    )
+    default String inventorySetup() {
```
chsami: Type String can be of InventorySetup with default return value of null

This will show a dropdown with all available inventory setups

runelite-client/src/main/java/net/runelite/client/plugins/microbot/herbrun/HerbrunScript.java
```diff
@@ -0,0 +1,262 @@
...
+                populateHerbPatches();
+                if (herbPatches.isEmpty()) {
+                    log("No herb patches ready to farm");
+                    Microbot.stopPlugin(plugin);
+                    return;
+                }
+                var inventorySetup = new Rs2InventorySetup(config.inventorySetup(), mainScheduledFuture);
+                if (!inventorySetup.doesInventoryMatch() || !inventorySetup.doesEquipmentMatch()) {
+                    boolean arrived = Rs2Walker.walkTo(Rs2Bank.getNearestBank().getWorldPoint(), 20);
+                    sleepUntil(() -> arrived, 100000);
```
chsami: not sure if we need to sleepUntil arrived, just let the script  loop continue if it hasn't arrived imo.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/herbrun/HerbrunScript.java
```diff
@@ -0,0 +1,262 @@
...
+            if (Rs2Inventory.hasItem("Weeds")) {
+                Rs2Inventory.drop("Weeds");
+            }
+            if (currentPatch == null) getNextPatch();
+            if (currentPatch == null) {
+                HerbrunPlugin.status = "Finishing up";
+                if (config.goToBank()) {
+                    BankLocation bankLocation = Rs2Bank.getNearestBank();
+                    boolean arrived = Rs2Walker.walkTo(bankLocation.getWorldPoint());
+                    sleepUntil(() -> arrived, 20000);
```
chsami: same here, no need to sleepUntil arrived, you  could be waiting 20 secs if arrived has return false.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/herbrun/HerbrunScript.java
```diff
@@ -0,0 +1,262 @@
...
+                }
+                HerbrunPlugin.status = "Finished";
+                shutdown();
+            }
+
+            if (!currentPatch.isInRange(10)) {
+                HerbrunPlugin.status = "Walking to " + currentPatch.getRegionName();
+                boolean arrived = Rs2Walker.walkTo(currentPatch.getLocation(), 20);
+                sleepUntil(() -> arrived);
```
chsami: same here, no need to sleepUntil arrived

chsami (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/herbrun/HerbrunScript.java
Gaisberg: I honestly thought walkTo generated some kind of executor that assigned the value to the boolean once completed and script would continue looping, no idea where I got that notion from...

chsami (approved):

PR #779, opened 2025-03-22
Player Monitor Config for Litemode defaulted to false/off - Added specific overload potion handling in NMZ

FireStarter-Instigator:
I kept having my nmz bots getting in a loop at the potions leading to a ban on one of them and since adding this little fix I have seen some improvements so thought I would add to the project.

I mistakenly as well put litemode in playermonitor to be automatically on so changed that to off.  

Also changed all instances of Random random to Rs2Random.between()

fixed indentation issue in the run of player monitor and also removed super.return in order for action cooldowns to not stall the plugin.


runelite-client/src/main/java/net/runelite/client/plugins/microbot/nmz/NmzScript.java
```diff
@@ -161,6 +164,26 @@ public boolean interactWithObject(int objectId) {
         return false;
     }
 
+    private void fetchOverloadPotions(int objectId, String itemName, int requiredAmount) {
```
slest1, line 167:
```Java
private void fetchPotions(int objectId, String itemName, int requiredAmount) {
    int currentAmount = Rs2Inventory.count(itemName);

    if (currentAmount >= requiredAmount) return;

    int neededAmount = requiredAmount - currentAmount;

    Rs2GameObject.interact(objectId, "Take");
    String widgetText = "How many doses of ";
    sleepUntil(() -> Rs2Widget.hasWidget(widgetText));

    if (Rs2Widget.hasWidget(widgetText)) {
        Rs2Keyboard.typeString(Integer.toString(neededAmount * 4));
        Rs2Keyboard.enter();
        sleepUntil(() -> Rs2Inventory.count(itemName) == requiredAmount);
    }
}
```

can we use the already generic function that can be applied to both, instead of creating separate function, unless its needed.

FireStarter-Instigator: Once overloads are sorted it doesn't really matter how many absorptions you withdraw, trying to be exact in that regard actually adds more confusion, that's why I also took out having the desired amount of absorptions as prereq for going into the dream as all that's neccesary is correct amount of (4) overloads. Absorptions are always withdrawn without fail. Code is rock solid on that. Shouldn't be giving any more opportunity for potential cycling.

FireStarter-Instigator: I kinda see what you mean though they're basically the same

FireStarter-Instigator: I think it's one of those situations where you take the same potato and put it on a different plate and it somehow tastes better

FireStarter-Instigator: I just deleted a couple of my comments as I was confused - I thought you were sending me the current `fetchPotions` method, which is:

```java
private void fetchPotions(int objectId, String itemName, int requiredAmount) {
    if (Rs2Inventory.count(itemName) == requiredAmount) return;
    Rs2GameObject.interact(objectId, "Take");
    String widgetText = "How many doses of ";
    sleepUntil(() -> Rs2Widget.hasWidget(widgetText));
    if (Rs2Widget.hasWidget(widgetText)) {
        Rs2Keyboard.typeString(Integer.toString(requiredAmount * 4));
        Rs2Keyboard.enter();
        sleepUntil(() -> Rs2Inventory.count(itemName) == requiredAmount);
    }
}

FireStarter-Instigator: the one above works perfect to just withdraw a bunch of absorptions and doesn't give any trouble as I stated before, we don't need to be exact on absorptions only overloads, and giving the same code for precision is redundant

slest1: > Once overloads are sorted it doesn't really matter how many absorptions you withdraw, trying to be exact in that regard actually adds more confusion, that's why I also took out having the desired amount of absorptions as prereq for going into the dream as all that's neccesary is correct amount of (4) overloads. Absorptions are always withdrawn without fail. Code is rock solid on that. Shouldn't be giving any more opportunity for potential cycling.

then just add a boolean as a param? i still dont need a reason for a seperate function? when it can be reused

FireStarter-Instigator: sounds like a good idea, I don't know how I would go about adding boolean as a param, can you show it here or add it and test when you push your rapid heal changes?

chsami (approved):

chsami (approved):

PR #786, opened 2025-03-24
feat: logout toggle, support longer shutdown logic

Gaisberg:


runelite-client/src/main/java/net/runelite/client/plugins/microbot/Microbot.java
```diff
@@ -261,13 +262,26 @@ public static boolean hopToWorld(int worldNumber) {
...
+                final JOptionPane optionPane = new JOptionPane(
+                        message,
+                        JOptionPane.INFORMATION_MESSAGE,
+                        JOptionPane.DEFAULT_OPTION
+                );
+
+                final JDialog dialog = optionPane.createDialog("Message");
+
+                // Set up timer to close the dialog after 10 seconds
+                Timer timer = new Timer(10000, e -> {
```
chsami: why are we closing this after 10 seconds? 

The idea of the message is that it is discarded by the user

Gaisberg: Rs2InventorySetup.load* uses this message, I can modify that also to not use this blocking message if need be. I just noticed this PR still has issues so please dont merge it just yet.

Gaisberg: In the latest commit I added a new method with given timeout and made Rs2InventorySetups use that.

chsami: Okay feel free to cleanup the code in comment and i'll take a look at it again. thanks!

chsami (approved):

PR #788, opened 2025-03-24
Some fixes to Vorkath

iMoisty:
fix: adjust Vorkath interaction timing and not dodging fireball after Zombie phase.
fix: reduce wait time for Vorkath animation to avoid melee attack at the start.
fix: issue with  timing and eating that cause death during wooxwalk.


gmason0 (changes requested):
Please change the upstream repo to development to fit within our development cycle. We don't merge changes directly into main

PR #794, opened 2025-03-25
Update PluginDescriptor

Misty-osrs:
Adding name to PluginDescriptor for future releases.


FireStarter-Instigator:
why do people do this why not when upload a release also update the plugin descriptor

gmason0:
I appreciate you taking the time to create a pull request, but you can simply add your own plugin descriptor when you release your plugins to microbot.

PR #795, opened 2025-03-26
fix: Rs2Walker close dialogues before using minigame teleport

Gaisberg:


runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/dialogues/Rs2Dialogue.java
```diff
@@ -121,6 +121,8 @@ private static boolean hasSpellFilterContinue() {
         return Rs2Widget.isWidgetVisible(162, 43);
     }
 
+    private static boolean hasGenericContinue() { return Rs2Widget.findWidget("Click here to continue") != null; }
```
chsami: bad actors will write "click here to continue" in chat to mess with scripts, we should try to avoid generic findWidget for actions that will result in input from the bot.

chsami (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/walker/Rs2Walker.java
```diff
@@ -1580,6 +1580,12 @@ private static boolean handleMinigameTeleport(Transport transport) {
         @Component final int SELECTED_MINIGAME = 4980747; // 76.11
         @Component final int TELEPORT_BUTTON = 4980768; // 76.32
 
+        // Minigame teleports cant be used if a dialogue is open.
+        if (Rs2Dialogue.isInDialogue()) {
```
chsami, line 1584: @g-mason0 Is there any specific reason you didn't want this code to live here? I don't see any issue with it, but I might be missing context.

gmason0: In my opinion, this just seems to be a workaround to the real issue which a plugin that was made isn't completing its dialogue.  

If a bot/plugin is interacting with an object or npc that consists of dialogue the plugin should be handling that dialogue as well instead of adding extra measures inside of the walker to deal with it. If this is the route you would like to take @chsami, by all means feel free but the real issue at hand is that whatever plugin is causing this problem should simply handle its dialogue so these other checks aren't required.

Gaisberg: If I may chime in this blocks use of minigame teleports in general. Instead of having this code in multiple plugins/bots why not have it in once place?

gmason0: Ideally, they would be selecting whatever options they need to finish their own dialogue or pressing esc to close the dialogue, instead of implementing a failsafe, it should be the plugins duty to handle its own actions before invoking the walker.

chsami: I think it makes sense to include it under the minigame teleport. I get that we shouldn’t just throw everything in there, but to me, this feels like it fits as part of the minigame teleport.

PR #798, opened 2025-03-26
Fix Rs2Npc.interact() to work with using item on npc and for no input .interact()

crispycritter99:
Previous refactor had oversight for using items on npcs (quest, phials unnoting, etc.) where "use" was not in the list of actions, and for blank .interact(), which in both cases the index stayed at -1 and returns false. 
The way the method now works: 
.interact() interacts the first menu action option of the npc
 .interact("[correct action]") interacts with the inputted menu action
 .interact("[wrong action]") returns false if "wrong action" isn't in the list of actions for the npc
 .interact() or interact(npc,"use") will use an item on the npc if the item is already selected. Otherwise, interact(npc,"use") will return false

fix by slest

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/npc/Rs2Npc.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/npc/Rs2Npc.java
@@ -6,6 +6,7 @@ import net.runelite.api.coords.WorldPoint;
 import net.runelite.client.game.npcoverlay.HighlightedNpc;
 import net.runelite.client.plugins.microbot.Microbot;
 import net.runelite.client.plugins.microbot.util.antiban.Rs2AntibanSettings;
+import net.runelite.client.plugins.microbot.util.inventory.Rs2Inventory;
 import net.runelite.client.plugins.microbot.util.camera.Rs2Camera;
 import net.runelite.client.plugins.microbot.util.combat.Rs2Combat;
 import net.runelite.client.plugins.microbot.util.coords.Rs2WorldPoint;
@@ -141,7 +142,7 @@ public class Rs2Npc {
             if (npcName == null || npcName.isEmpty()) return false;
             return (exact ? npcName.equalsIgnoreCase(name) : npcName.toLowerCase().contains(name.toLowerCase())) && x.getInteracting() == Microbot.getClient().getLocalPlayer();
         }).sorted(Comparator.comparingInt(value -> value.getLocalLocation().distanceTo(Microbot.getClient().getLocalPlayer().getLocalLocation())))
-          .collect(Collectors.toList());
+                .collect(Collectors.toList());
     }
 
     /**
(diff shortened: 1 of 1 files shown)


slest1:
Thanks for making the pr 🫶

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/npc/Rs2Npc.java
```diff
@@ -577,28 +578,48 @@ public static boolean interact(Rs2NpcModel npc, String action) {
...
-            int index = -1;
-            String[] actions = npcComposition.getActions();
-            if (actions != null) {
-                for (int i = 0; i < actions.length; i++) {
-                    if (actions[i] != null && actions[i].equalsIgnoreCase(action)) {
-                        index = i;
-                        break;
+            MenuAction menuAction;
+            if (Rs2Inventory.isItemSelected()) {
+                menuAction = MenuAction.WIDGET_TARGET_ON_NPC;
```
gmason0, line 583: This part isn't required, as getMenuAction(int) will already return `MenuAction.WIDGET_TARGET_ON_NPC` if isWidgetSelected()

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/npc/Rs2Npc.java
```diff
@@ -577,28 +578,48 @@ public static boolean interact(Rs2NpcModel npc, String action) {
...
+            } else {
+                int index = -1;
+                String[] actions = npcComposition.getActions();
+
+                if (action == null || action.isEmpty()) {
```
gmason0, line 588: I am not sure if we want to leverage a stream with optionals for more concise logic? 

Example: 

```java
NPCComposition npcComposition = Microbot.getClientThread().runOnClientThread(
        () -> Microbot.getClient().getNpcDefinition(npc.getId()));

if (npcComposition == null || npcComposition.getActions() == null) {
    Microbot.log("Error: Could not get NPC composition or actions for NPC: " + npc.getName());
    return false;
}

int index = -1;
String[] actions = npcComposition.getActions();

if (action == null || action.isEmpty()) {
    OptionalInt optionalIndex = IntStream.range(0, actions.length)
            .filter(i -> actions[i] != null && !actions[i].isEmpty())
            .findFirst();

    if (optionalIndex.isPresent()) {
        index = optionalIndex.getAsInt();
        action = actions[index];
    }
}
else {
    String finalAction = action;
    OptionalInt optionalIndex = IntStream.range(0, actions.length)
            .filter(i -> actions[i] != null && actions[i].equalsIgnoreCase(finalAction))
            .findFirst();

    if (optionalIndex.isPresent()) {
        index = optionalIndex.getAsInt();
    }
}

if (index == -1) {
    Microbot.log("Error: Action '" + action + "' not found for NPC: " + npc.getName());
    return false;
}

MenuAction menuAction = getMenuAction(index);
if (menuAction == null) {
    Microbot.log("Error: Could not get menu action for action '" + action + "' on NPC: " + npc.getName());
    return false;
}
```

gmason0 (changes requested):
I think this is a good idea to prevent action from being null but I feel this could also have unexpected outcomes. 

@chsami - What are you thoughts on this change?

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/npc/Rs2Npc.java
slest1: stream-based approach for finding the action index. I agree is more concise, and it leverages `Optional`

gmason0:
I have committed the commented changes to remedy the current issues with this method. Thanks for taking a look @ it.

PR #799, opened 2025-03-26
feat: make nmz scheduable, handle case where mobs arent attacking

Gaisberg:


runelite-client/src/main/java/net/runelite/client/plugins/microbot/nmz/NmzConfig.java
```diff
@@ -1,9 +1,11 @@
 package net.runelite.client.plugins.microbot.nmz;
 
 import net.runelite.client.config.*;
+import net.runelite.client.plugins.microbot.inventorysetups.InventorySetup;
+import net.runelite.client.plugins.microbot.util.Rs2InventorySetup;
 
 @ConfigGroup("nmz")
-@ConfigInformation("Start outside NMZ with gp in Dominic's coffer and make sure bankpin has been unlocked <br />" +
```
gmason0, line 6: We probably should still include this information for the user to know that they need to fill their coffer before using the plugin.

gmason0 (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/nmz/NmzConfig.java
slest1: Maybe also include, that they need to have setup NMZ at-least once before. Don't know if it was there before

runelite-client/src/main/java/net/runelite/client/plugins/microbot/nmz/NmzPlugin.java
FireStarter-Instigator, line 1: i like that canBeScheduled = true

runelite-client/src/main/java/net/runelite/client/plugins/microbot/nmz/NmzConfig.java
Gaisberg: The part about dream being setup atleast once is left unchanged and on the next line. Ah yes I'll add the GP part back.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/nmz/NmzScript.java
FireStarter-Instigator, line 1: interesting to add to fight back like that you would get more xp per hour that way but might increase actions per minute  - i would have it as a config of "optimise" to get that xp boost to attack the npc when not in combat

runelite-client/src/main/java/net/runelite/client/plugins/microbot/nmz/NmzConfig.java
FireStarter-Instigator, line 1: also though as someone that never uses inventory setups would this affect me to use nmz?

runelite-client/src/main/java/net/runelite/client/plugins/microbot/nmz/NmzScript.java
Gaisberg: The original issue was that sometimes the NPCs blocked each other and I had the bot stand for hours.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/nmz/NmzConfig.java
Gaisberg: Yes you would have to setup an inventory setup to use this after these changes.

FireStarter-Instigator: pls no

runelite-client/src/main/java/net/runelite/client/plugins/microbot/nmz/NmzScript.java
FireStarter-Instigator: interesting, then you could maybe have it that if not attacking npc for 30 seconds that it would attack

Gaisberg: So you think the current 20 seconds isnt enough?

FireStarter-Instigator: oh from when i read it over it looked like it was immediate or almost immediate to me didn't realise you had a 20 second trigger

runelite-client/src/main/java/net/runelite/client/plugins/microbot/nmz/NmzScript.java
```diff
@@ -46,26 +53,47 @@ public class NmzScript extends Script {
...
+    public boolean run() {
         prayerPotionScript = new PrayerPotionScript();
         Microbot.getSpecialAttackConfigs().setSpecialAttack(true);
         mainScheduledFuture = scheduledExecutorService.scheduleWithFixedDelay(() -> {
             try {
                 if (!Microbot.isLoggedIn()) return;
+                if (!initialized) {
+                    initialized = true;
+                    var inventorySetup = new Rs2InventorySetup(config.inventorySetup(), mainScheduledFuture);
+                    if (!inventorySetup.doesInventoryMatch() || !inventorySetup.doesEquipmentMatch()) {
```
chsami, line 80: ignore the inventorysetup code if config.inventorySetup() is null

chsami (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/nmz/NmzConfig.java
FireStarter-Instigator: nooooooooooooooooooooooooooooooooooooo

PR #802, opened 2025-03-26
add Rs2Player.hasThrallActive to check if thrall is active

crispycritter99:
Checks varbits for thrall cooldown and thrall active. if either one is true, method returns true

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/player/Rs2Player.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/player/Rs2Player.java
@@ -56,6 +56,8 @@ public class Rs2Player {
     public static int antiPoisonTime = -1;
     public static int teleBlockTime = -1;
     public static int goadingTime = -1;
+    public static boolean thrallActive = false;
+    public static boolean thrallCooldown = false;
     public static Instant lastAnimationTime = null;
     private static final long COMBAT_TIMEOUT_MS = 10000;
     private static long lastCombatTime = 0;
@@ -113,6 +115,9 @@ public class Rs2Player {
         } else return antiVenomTime < VENOM_VALUE_CUTOFF;
     }
 
+    public static boolean hasThrallActive() {
+        return thrallActive||thrallCooldown;
+    }
     public static boolean hasAntiPoisonActive() {
         return antiPoisonTime > 0;
     }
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/player/Rs2Player.java
```diff
@@ -181,6 +186,38 @@ public static void handleTeleblockTimer(VarbitChanged event){
...
+    /**
+     * Handles updates to the teleblock timer based on changes to the {@link Varbits#TELEBLOCK} varbit.
+     *
+     * @see Varbits#TELEBLOCK
+     */
+    public static void handleThrallCooldown(VarbitChanged event){
+        if (event.getVarbitId() == Varbits.RESURRECT_THRALL_COOLDOWN) {
+//            int time = event.getValue();
+
+            if (event.getValue() == 1) {
```
gmason0, line 214: I am not sure if we need to monitor both varbits here as they seem to pretty much do the same thing.

Why don't we just change this to:

```java
    public static boolean canCastThrall() {
        return Microbot.getVarbitValue(Varbits.RESURRECT_THRALL_COOLDOWN) != 1;
    }
```
These other methods that leverage the onVarbitChanged event are due to them providing a system time to count down the time remaining, as opposed to this being simply either 0 or 1 to determine whether you can cast the thrall.

I also feel that this would be better placed into Rs2Magic instead of Rs2Player,

gmason0 (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/player/Rs2Player.java
crispycritter99: Good idea to put it into Rs2Magic, i had it in Rs2Player because I was using the antifire/potion timers as a reference on how to check the varbits

I check both varbits because RESURRECT_THRALL switches on 5 ticks after the spell is casted, while RESSURECT_THRALL_COOLDOWN switches on immediately upon spell cast, and then switches off 9 ticks after RESURRECT_THRALL switches on. Checking both should ensure that hasThrallActive returns true for the time between the spell being casted and the thrall despawning. However I do think that .getVarbitValue is a better method to use, I had no idea that was a thing
image

crispycritter99: I'm gonna close it and then reopen with the method in Rs2Magic

PR #821, opened 2025-04-02
fix: strings not working in bankstander config

FireStarter-Instigator:


chsami:
Have you Found out why they don't work?

FireStarter-Instigator:
> Have you Found out why they don't work?

nope, but I had similar issue with wildy moss killer after some updates and fix was just to make it as an ID

PR #829, opened 2025-04-06
Side load plugins from github

chsami:
image

image
image
image
image
image
image
image

You can package jars to side load through maven inside pom.xml of the runelite client project.

image

image

You can then choose to host the jar somewhere on github and side loade it.


gmason0 (approved):
Great work on this! I'll let you merge it but I like what i see!

PR #844, opened 2025-04-10
Update handleAmount for use with quantities 1,5,10

crispycritter99:
I noticed that if you input a quantity of 1, 5, or 10 into the methods, the method will still choose the x option and type in the value. With these changes, it should instead choose the menu action for 1, 5, and 10, skipping the setting of X and being much faster overall. 

Also returns false if the "Enter amount:" widget never appears, so that the plugin doesn't type a number into the public chat.


gmason0:
I understand the thought process behind this but maybe we shouldn't over complicate the method by leveraging withdraw 5 & withdraw 10 & keep this method focused on just withdrawx option

PR #848, opened 2025-04-10
Chaos altar plugin up to date with 1.8.4

FireStarter-Instigator:
- rs2grounditem fix

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/bee/chaosaltar/ChaosAltarPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/bee/chaosaltar/ChaosAltarPlugin.java
@@ -5,14 +5,11 @@ import lombok.extern.slf4j.Slf4j;
 import net.runelite.client.config.ConfigManager;
 import net.runelite.client.plugins.Plugin;
 import net.runelite.client.plugins.PluginDescriptor;
-import net.runelite.client.plugins.PluginInstantiationException;
-import net.runelite.client.plugins.microbot.Microbot;
 import net.runelite.client.plugins.microbot.accountselector.AutoLoginPlugin;
 import net.runelite.client.plugins.microbot.storm.plugins.PlayerMonitor.PlayerMonitorPlugin;
 import net.runelite.client.ui.overlay.OverlayManager;
 
 import javax.inject.Inject;
-import java.awt.*;
 
 @PluginDescriptor(
         name = PluginDescriptor.Bee + "Chaos Altar",
@@ -42,12 +39,10 @@ public class ChaosAltarPlugin extends Plugin {
 
 
     @Override
-    protected void startUp() throws AWTException, PluginInstantiationException {
+    protected void startUp() {
         if (overlayManager != null) {
             overlayManager.add(chaosAltarOverlay);
         }
-            if (!Microbot.getPluginManager().isPluginEnabled(playerMonitorPlugin)) {Microbot.getPluginManager().startPlugin(playerMonitorPlugin);}
-            if (!Microbot.getPluginManager().isPluginEnabled(autoLoginPlugin)) {Microbot.getPluginManager().startPlugin(autoLoginPlugin);}
         chaosAltarScript.run(config);
     }
 
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/bee/chaosaltar/ChaosAltarPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/bee/chaosaltar/ChaosAltarPlugin.java
@@ -1,0 +1,64 @@
+package net.runelite.client.plugins.microbot.bee.chaosaltar;
+
+import com.google.inject.Provides;
+import lombok.extern.slf4j.Slf4j;
+import net.runelite.api.events.GameTick;
+import net.runelite.client.config.ConfigManager;
+import net.runelite.client.eventbus.Subscribe;
+import net.runelite.client.plugins.Plugin;
+import net.runelite.client.plugins.PluginDescriptor;
+import net.runelite.client.ui.overlay.OverlayManager;
+
+import javax.inject.Inject;
+import java.awt.*;
+
+@PluginDescriptor(
+        name = PluginDescriptor.Bee + "Chaos Altar",
+        description = "Automates bone offering at the Chaos Altar",
+        tags = {"prayer", "bones", "altar"},
+        enabledByDefault = false
+)
+@Slf4j
+public class ChaosAltarPlugin extends Plugin {
+    @Inject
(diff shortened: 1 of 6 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/microbot/bee/chaosaltar/ChaosAltarPlugin.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    @Inject
+    private OverlayManager overlayManager;
+    @Inject
+    ChaosAltarOverlay chaosAltarOverlay;
+
+
+    @Override
+    protected void startUp() {
```
gmason0, line 36: I notice you removed starting the auto login plugin & player monitor with the script. 

Any reason why something like this couldn't work for enabling the plugin? This should atleast point you in the direction of creating a method that can turn off or on the plugin by passing in a flag 

```java

private void enableBreakHandler() {
        Plugin breakHandlerPlugin = Microbot.getPluginManager().getPlugins().stream()
                .filter(x -> x.getClass().getName().equals(BreakHandlerPlugin.class.getName()))
                .findFirst()
                .orElse(null);

        Microbot.startPlugin(breakHandlerPlugin);
    }
```

gmason0 (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/bee/chaosaltar/ChaosAltarPlugin.java
FireStarter-Instigator: if you'd like to put in auto enable feel free after release I just don't think it's a good idea to force them on like that.

gmason0: I dont think the plugin functions correctly without them enabled? If you're not going to implement your own logic for player detection as well as log in & out, we should be enabling / disabling them to reduce the amount of human error.

FireStarter-Instigator: if a user would like to use this plugin without autologin and without player monitor it will work perfectly fine with no errors, they'll just die a lot more often, it's up to them what their goals are when using it. currently the meta on microbot is to manually do chaos altar with player monitor and autologin, now there is a plugin that means you don't have to manually do it.

gmason0: Not to disregard what you said, but we dont see other plugins that require the break handler plugin to function correct say "make sure you enable break handler". Considering this plugin is a wilderness activity for users have to perform this additional configuration is unnecessary

FireStarter-Instigator: if you'd like to talk about previous merges that are unrelated to this plugin which you are quoting from and editing them to enhance them then let's do that over the relevant PR.

For Chaos Altar Plugin autologin is NOT necessary to this plugin and player monitor is an OPTIONAL enhacement DEPENDING how you are using it. The user may be with a team and requires no player monitor and no autologin. They would then have to do additionally unnecessary configuration to get it to work how it should = offers bones and rebanks

gmason0: I am not referring to another pull request. I am referring to other plugins in the client to do not tell you to enable other plugins for them to function correctly.

gmason0: >  The user may be with a team and requires no player monitor and no autologin.

The player monitor plugin can be configured in a manner to ignore players on your friends list or in your clan. So I'd assume that would be the team of people that you want to ignore.

FireStarter-Instigator: > I am not referring to another pull request. I am referring to other plugins in the client to do not tell you to enable other plugins for them to function correctly.

Moss Killer Plugin will not work unless breakhandler is on. So why do you say we DONT see this?

FireStarter-Instigator: > > The user may be with a team and requires no player monitor and no autologin.
> 
> The player monitor plugin can be configured in a manner to ignore players on your friends list or in your clan. So I'd assume that would be the team of people that you want to ignore.

then the user can configure the player monitor how they like, if they want it on or not, chaos altar plugin should not be interacting with player monitor directly turning it on or off.

gmason0: shocking, another one of your GPT plugins. 

Take a look at the plugins created by any of the official developers. For example, the antiban turns on the breakhandler for micro-breaks, the shooting star plugin turns on the break handler for it to break at a bank, the giant mole plugin turns on the bossing timers plugin. If your plugin DEPENDS on another plugin to get expected functionality, then you should be enabling it.

FireStarter-Instigator: > shocking, another one of your GPT plugins.
> 
> Take a look at the plugins created by any of the official developers. For example, the antiban turns on the breakhandler for micro-breaks, the shooting star plugin turns on the break handler for it to break at a bank, the giant mole plugin turns on the bossing timers plugin. If your plugin DEPENDS on another plugin to get expected functionality, then you should be enabling it.

irrelevant comment, chaos altar plugin does not rely on any plugin apart from webwalker to function

gmason0: You're just avoiding the point here.

Training Prayer at the Chaos altar would put the bot in a situation where you would want to avoid losing bones to a PK'er. Even when there is a team protecting you there, you would still want to logout & switch worlds when there is a potentially dangerous player around you that could kill you & take your bones. 

So if you want to be lazy & not add that logic into this plugin, I think that we should be turning on these plugins to have that functionality.

FireStarter-Instigator: As an open source project anyone is free to add or take away what they like, the chaos altar plugin takes bones from the bank, uses them on altar, and rebanks. Nothing more nothing less. If people wanna add to it, feel free, if people wanna use other plugins alongside it, also be free.

PR #852, opened 2025-04-11
feat: add setup and other additions to MLM script

h5law:
This PR does the following:
1. Gear up and navigate to MLM if not done
2. Monitor sack size changes (deposits and upgrades)
3. Mine exact sack size then stop and deposit 
4. Always use best available pickaxe and upgrade if possible
5. Overlay render tweaks


gmason0:
I am going to close this pull request, considering it looks as you're re-naming the microbot project to nanobot..

PR #854, opened 2025-04-12
Chaos Altar Plugin + Moss Killer Plugin fixes + API additions

FireStarter-Instigator:
1. New Chaos Altar Plugin which runs bones to chaos altar, dies when bones are all used up, gets out more bones from the bank and repeats. included are two placeholder alternate methods in config suggested by the community to be added, graveyard looter and f2p walk. 
2. Wildy Safer mode is now fixed for f2p rangers, there was a bug that made it so you had to restart the plugin when banking and the normal moss killer script has had an update related to shutdown case not shutting the script down and introduced some hardcoded failsafe's for lower level users. 
3. Changes made to rs2ground item to have loot from world point to work. Changes made to rs2player and rs2food to accommodate an optional 1-tick food prioritisation overload to eatAt (for maximum healing) and changes to rs2npc to stop rs2npc.attack from interacting with npcs with 0 hp.

Chaos Altar Plugin demo: https://youtu.be/1Vqz4162qz4


acun9323:
1. It doesn't equip the burning amulet so spare one extra inventory space.
2. On the way to the altar it takes a lot of damage from the Chaos Fanatic. I think if someone using this script at level 3 he will die immediately

---
3 runs later: 
It got stuck at the altar, did not 'use' dragon bones on the altar I had to manually take action

7 runs later:
It did not logout when pker logged in???

Edit: I didn't know we had to activate Player Monitoring manually when I was testing this

acun9323:
image
Third time I died, I don't think the script is ready yet

FireStarter-Instigator:
> 1. It doesn't equip the burning amulet so spare one extra inventory space.
> 2. On the way to the altar it takes a lot of damage from the Chaos Fanatic. I think if someone using this script at level 3 he will die immediately
> 
> ---
> 3 runs later: 
> It got stuck at the altar, did not 'use' dragon bones on the altar I had to manually take action
> 
> 7 runs later:
> It did not logout when pker logged in???

Hey Acun thank you so much for leaving a response.

1. Yes it doesn't wear the burning amulet, just complicates equipment checks. To do that.

2. Yeah webwalker needs to path to the south instead of the north but that's the predicted path. I could add that it goes south of the chaos altar and then inside, would fix the pathing, thanks for pointing it out.

3. Considering the script is a couple hundred lines long at most, could you take a quick peek to see why it wouldn't use the bones on the altar on your third run?

4. Player monitor can be like that, nothing to do with chaos altar plugin 

5. Death is inevitable this plugin makes no promises otherwise, it simply automatises the chaos altar process. For improvements with dying to pkers I would suggest having a guard alt or finding ways to enhance player monitor.

FireStarter-Instigator:
Death is inevitable this plugin makes no promises otherwise, it simply automatises the chaos altar process. For improvements with dying to pkers I would suggest having a guard alt or finding ways to enhance player monitor. 

> image
> Third time I died, I don't think the script is ready yet

FireStarter-Instigator:
> 3 runs later: It got stuck at the altar, did not 'use' dragon bones on the altar I had to manually take action

Hi Acun, again thank you for testing the plugin. Here is the offer bones condition, I think the problem might be that the webwalker pathed the player to outside the bounds of "isInAltar". I've been meaning to make it bigger but you risk the plugin thinking you're at the altar when you're really outside it and there's a closed door and you will spam click the altar. 

Maybe "atAltar" could be true if the altar object is within line of site and within a larger altar area, I think that would fix that issue. 

if (inWilderness && hasAnyBones && atAltar) {
            return State.OFFER_BONES;
        }

FireStarter-Instigator:
to add:
1. improve accuracy of atAltar boolean
2. path to the south of rock pile at entrance to chaos altar and then to inside the chaos altar to avoid attacks from chaos fanatic
3. explain in the config how best to use chaos altar plugin for first time users and testers
4. (from the discord) add a potential f2p option, which eliminates some risk of pkers as you run around in f2p worlds to the altar instead of traditional burning amulet usage.  

Edit: all added

chsami (approved):

PR #860, opened 2025-04-16
BloodsRc

Zeus-Xora:
Fixes for Ring of duelling issue if using POH and fixes for unknownpouch


runelite-client/src/main/java/net/runelite/client/plugins/microbot/frosty/bloods/BloodsScript.java
```diff
@@ -93,6 +95,10 @@ public boolean run() {
                 }
 
                 if (Rs2Inventory.anyPouchUnknown()) {
+                    if (Rs2Bank.isOpen()){
+                        Rs2Keyboard.keyPress(KeyEvent.VK_ESCAPE);
```
gmason0, line 99: use closeBank, this will use the esc key to close the bank if "Esc to close interface" Setting is configured

runelite-client/src/main/java/net/runelite/client/plugins/microbot/frosty/bloods/BloodsScript.java
```diff
@@ -455,17 +453,33 @@ private void handleWalking() {
...
-        if (plugin.getMyWorldPoint().equals(firstCaveExit)) {
-            Microbot.log("Walking to ruins: " + outsideBloodRuins);
-            Rs2Walker.walkTo(outsideBloodRuins);
-            sleepUntil(() -> plugin.getMyWorldPoint().equals(outsideBloodRuins), 1200);
+        if (plugin.getMyWorldPoint().equals(firstCaveExit) &&
+                Rs2Player.getRealSkillLevel(Skill.AGILITY) > 93) {
+            Microbot.log("Walking to blood ruins " +
+                     outsideBloodRuins93);
+            Rs2Walker.walkTo(outsideBloodRuins93);
+            sleepUntil(() -> Rs2Player.getWorldLocation().equals(outsideBloodRuins93), 1200);
```
gmason0, line 461: Something like this can cause unexpected delays as the final tile in walkTo is randomized, so you may not always land on the expected world point. would be better to see if you are within 5-10 tiles of this world point

runelite-client/src/main/java/net/runelite/client/plugins/microbot/frosty/bloods/BloodsScript.java
```diff
@@ -455,17 +453,33 @@ private void handleWalking() {
...
-        if (plugin.getMyWorldPoint().equals(outsideBloodRuins)) {
-            Microbot.log("Current location after walking: " + plugin.getMyWorldPoint());
-            if (plugin.getMyWorldPoint().equals(outsideBloodRuins)) {
-                state = State.CRAFTING;
-            }
+        if (plugin.getMyWorldPoint().equals(firstCaveExit) &&
+        Rs2Player.getRealSkillLevel(Skill.AGILITY) < 93 && Rs2Player.getRealSkillLevel(Skill.AGILITY) >74) {
+            Microbot.log("Walking to ruins: " + outsideBloodRuins74);
+            Rs2Walker.walkTo(outsideBloodRuins74);
+            sleepUntil(() -> plugin.getMyWorldPoint().equals(outsideBloodRuins74), 1200);
```
gmason0, line 468: Something like this can cause unexpected delays as the final tile in walkTo is randomized, so you may not always land on the expected world point. would be better to see if you are within 5-10 tiles of this world point

runelite-client/src/main/java/net/runelite/client/plugins/microbot/frosty/bloods/BloodsScript.java
```diff
@@ -455,17 +453,33 @@ private void handleWalking() {
...
+        }
+
+        if (plugin.getMyWorldPoint().equals(firstCaveExit) && Rs2Player.getRealSkillLevel(Skill.AGILITY) <74) {
+            Microbot.log("Walking to ruins: " + outsideBloodRuins73);
+            Rs2Walker.walkTo(outsideBloodRuins73);
+            sleepUntil(() -> plugin.getMyWorldPoint().equals(outsideBloodRuins73), 1200);
```
gmason0: Something like this can cause unexpected delays as the final tile in walkTo is randomized, so you may not always land on the expected world point. would be better to see if you are within 5-10 tiles of this world point

gmason0 (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/frosty/bloods/BloodsScript.java
Zeus-Xora: These are the world points after the agility shortcuts. you can only land on one

I then have a region check to transition into crafting now

Maybe I should change the non shortcut method to not equals

PR #874, opened 2025-04-22
changed torag's tile, added canReach to the brother check, made entering mounds more consistent

Gage307:
changed torag's tile in the enum
added canReach to the brother check
Made entering mounds more consistent

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/barrows/BarrowsScript.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/barrows/BarrowsScript.java
@@ -173,6 +173,7 @@ public class BarrowsScript extends Script {
 
                                 // We're not in the mound yet.
                                 Rs2Walker.walkTo(mound);
+                                sleep(300, 600);
                                 if (mound.distanceTo(Rs2Player.getWorldLocation()) <= 1) {
                                     if(!Rs2Player.isMoving()) {
                                         sleepUntil(()-> Rs2Player.isMoving(), Rs2Random.between(1000,2000));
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/microbot/barrows/BarrowsScript.java
```diff
@@ -908,7 +911,7 @@ public void checkForBrother(){
             shouldWalk = false;
             Rs2Walker.setTarget(null);
             Rs2PrayerEnum neededprayer = Rs2PrayerEnum.PROTECT_MELEE;
-            if (currentBrother != null) {
+            if (currentBrother != null && Rs2Walker.canReach(currentBrother.getWorldLocation())) {
```
gmason0: In this case, where you want to check if you can reach the brother/not stuck behind a door, it would probably be better to check line of sight instead of canReach. canReach will generate a path between the player & the worldpoint that is passed through. In the case of these doors, they are ignore as collision unless the door is locked/unopenable

gmason0 (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/barrows/BarrowsScript.java
Gage307: Damn I was hoping that wasn't the case. How can I check line of sight?

gmason0: `Rs2Npc.hasLineOfSight(rs2npcmodel)`

Gage307: :O It's that easy. Thank you george!

gmason0 (approved):

PR #876, opened 2025-04-22
Logiko's Antiban Preloader

logikodied:
**Forces Rs2AntibanSettings from config.**
Overrides all antiban plugin values with a pre-fixed preset.
image


chsami:
We'll add a checkbox to always use antiban settings instead of the script settings

PR #884, opened 2025-04-24
Added a Revenant killing script

Gage307:
Added a Revenant killing script

Supports the breakhandler.
Very good anti PK, and anti patterning.
Supports worldhopping if the script killed the revs quickly
Revs supported Imps, Goblins, Pyrefiends, Cyclops, Demons, Darkbeasts, and Orks.


Gaisberg:
https://github.com/chsami/Microbot/blob/main/docs%2Fscheduler%2FREADME.md

😉

Gage307:
> https://github.com/chsami/Microbot/blob/main/docs%2Fscheduler%2FREADME.md
> 
> 😉

Ohh nice! That'll come in very handy

FireStarter-Instigator:
I don't see any anti-pk in your script. Maybe you mean it's good at hopping or running when a pker is near?

runelite-client/src/main/java/net/runelite/client/plugins/microbot/revKiller/revKillerScript.java
```diff
@@ -0,0 +1,847 @@
...
+            Microbot.log("We have enough loot");
+            return true;
+        }
+        Microbot.log("We have "+value+" worth of loot");
+        return false;
+    }
+
+    public boolean areWeEquipped(){
+
+        if(Rs2Equipment.get(EquipmentInventorySlot.AMMO)!=null) {
```
chsami: Early returns could help for clarity

Gage307:
Yes I mean it's excellent at getting away from PKers before teleblock is casted.

chsami (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/revKiller/revKillerScript.java
Gage307: Could I have an example please?

slest1:
```Java
public boolean areWeEquipped() {
    return Rs2Equipment.get(EquipmentInventorySlot.AMMO) != null;
}
```

```Java
if (isPkerAround()) {
    getAwayFromPker();
    return;
}
```
so you can early exit when PKer is detected.

Gage307: Oh I see, thank you! I'll do it now

chsami: Suggestion:
Rewrite the areWeEquipped() method to use early returns instead of deep nested if blocks.
The current structure is hard to read and maintain. Using early returns makes the intent clear and avoids a pyramid of doom.

```suggestion
public boolean areWeEquipped() {
    if (Rs2Equipment.get(EquipmentInventorySlot.AMMO) == null) {
        Microbot.log("We have no ammo!");
        return false;
    }

    if (Rs2Equipment.get(EquipmentInventorySlot.AMMO).getQuantity() < LowOnArrowsCount) {
        Microbot.log("We don't have enough ammo!");
        return false;
    }

    if (!Rs2Inventory.contains(it -> it != null && it.getName().contains("Stamina"))) {
        Microbot.log("We have no stam!");
        return false;
    }

    if (!Rs2Inventory.contains(it -> it != null && it.getName().contains("Ranging"))) {
        Microbot.log("We have no Ranging potion!");
        return false;
    }

    if (!Rs2Inventory.contains("Shark")) {
        Microbot.log("We're out of sharks!");
        return false;
    }

    if (Rs2Inventory.count("Shark") < 2) {
        Microbot.log("We have less than 2 sharks!");
        return false;
    }

    if (Rs2Equipment.get(EquipmentInventorySlot.RING) == null) {
        Microbot.log("ring is null");
        return false;
    }

    if (!Rs2Equipment.get(EquipmentInventorySlot.RING).getName().contains("dueling")) {
        Microbot.log("We don't have our ring of dueling");
        return false;
    }

    if (isItTimeToGo()) {
        Microbot.log("We have too much loot! Banking");
        return false;
    }

    if (Rs2Equipment.get(EquipmentInventorySlot.AMULET) == null) {
        Microbot.log("amulet is null");
        return false;
    }

    if (!Rs2Equipment.get(EquipmentInventorySlot.AMULET).getName().contains("Amulet of glory(")) {
        Microbot.log("amulet is not charged");
        return false;
    }

    Microbot.log("We're fully equipped and ready to go.");
    return true;
}

```

runelite-client/src/main/java/net/runelite/client/plugins/microbot/revKiller/revKillerScript.java
```diff
@@ -0,0 +1,906 @@
...
+
+                    if(Rs2Player.getWorldLocation().distanceTo(revimp)>10){
+
+                        WalkToRevs();
+
+                    } else {
+
+                        equipArrows();
+
+                        if(isPkerAround()){
```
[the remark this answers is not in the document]
FireStarter-Instigator: yup or have your plugin class checking if isPkerAround() and your script class only reads that boolean rather than working on the detection itself, that way you don't have to bother with asynch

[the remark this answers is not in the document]
FireStarter-Instigator: yup or have your plugin class checking if isPkerAround() and your script class only reads that boolean rather than working on the detection itself, that way you don't have to bother with asynch

runelite-client/src/main/java/net/runelite/client/plugins/microbot/revKiller/revKillerScript.java
```diff
@@ -0,0 +1,918 @@
...
+            }
+            if(!Rs2Bank.isOpen()) {
+                sleepUntil(() -> BreakHandlerScript.isBreakActive(), 2000000);
+            }
+        }
+    }
+
+    public void DidWeDie(){
+        if(Rs2Player.getWorldLocation().distanceTo(BankLocation.LUMBRIDGE_FRONT.getWorldPoint()) < 50){
+            Microbot.log("We died");
```
gmason0, line 182: This isn't a good way to handle if we have died, considering players can change their respawn location from lumbridge to somewhere else. Here is a better way to handle if we had died.

```java
@Subscribe
public void onActorDeath(ActorDeath event) {
    if (event.getActor() == Microbot.getClient().getLocalPlayer()) {
        // set a var like a boolean like "hasDied"
    }
}
```
OR you could listen for the chat message
```java
@Subscribe
public void onChatMessage(ChatMessage event) {
    if (event.getType() != ChatMessageType.GAMEMESSAGE) return;
    
    if (event.getMessage().equalsIgnoreCase("oh dear, you are dead!")) {
       // do something
    }
}
```

These are things that would land in the plugin class, or you could register a different class to the event bus, if you wanted to.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/revKiller/revKillerScript.java
```diff
@@ -0,0 +1,918 @@
...
+        if(Rs2Player.getHealthPercentage() <= generateRandomNumber(70,80)){
+            if(Rs2Inventory.contains("Shark")){
+                if(Rs2Inventory.interact("Shark", "Eat")){
+                    sleepUntil(()-> isPkerAround(),generateRandomNumber(2500,3500));
+                }
+            }
+        }
+    }
+
+    public boolean WeAreInTheCaves(){
```
gmason0, line 155: Instead of checking the Y Level, you can ensure that you are in the Rev Caves by checking for RegionID

```java
private boolean isInsideCave() {
    List<Integer> regions = List.of(12957, 12701, 12702, 12703, 12959, 12958);
    return regions.stream().anyMatch(r -> r == Rs2Player.getWorldLocation().getRegionID());
}
```

runelite-client/src/main/java/net/runelite/client/plugins/microbot/revKiller/revKillerScript.java
```diff
@@ -0,0 +1,918 @@
...
+    @Override
+    public void shutdown() {
+        super.shutdown();
+    }
+
+    public void EatFood(){
+        if(Rs2Player.getHealthPercentage() <= generateRandomNumber(70,80)){
+            if(Rs2Inventory.contains("Shark")){
+                if(Rs2Inventory.interact("Shark", "Eat")){
+                    sleepUntil(()-> isPkerAround(),generateRandomNumber(2500,3500));
```
gmason0, line 149: Would we want to be sleeping Until a PKer is near when eating? What if we are taking damage from a rev monster?

```java
private void eatFood() {
    while (Rs2Player.getHealthPercentage() < Rs2Random.between(70,80)) {
        if (!isRunning()) break;

        Rs2Inventory.interact(Rs2Inventory.getInventoryFood().stream().map(Rs2ItemModel::getName).collect(Collectors.toList()), "eat");
        Rs2Inventory.waitForInventoryChanges(1800);
    }
}
```

runelite-client/src/main/java/net/runelite/client/plugins/microbot/revKiller/revKillerScript.java
```diff
@@ -0,0 +1,918 @@
...
+        );
+    }
+
+    public void futurePKCheck(){
+        if(isPkerAround()){
+            getAwayFromPker();
+        }
+    }
+
+    public void hopToNewWorld(){
```
gmason0, line 419: If you want to only use worlds from a specific region, why not leverage the following? This would allow anyone to stay within their specific world region

```java
World currentWorld = Microbot.getWorldService().getWorlds().findWorld(Microbot.getClient().getWorld());
int newRandomWorld = Login.getRandomWorld(Login.activeProfile.isMember(), currentWorld.getRegion());
```

runelite-client/src/main/java/net/runelite/client/plugins/microbot/revKiller/revKillerScript.java
```diff
@@ -0,0 +1,918 @@
...
+        return false;
+    }
+
+    Microbot.log("We're fully equipped and ready to go.");
+    return true;
+
+    }
+
+
+    public boolean isPkerAround(){
```
gmason0, line 871: This logic could be trimmed up a bit by leveraging Java Streams, this is very helpful in a sense as streams do not block the logic flow compared to using a traditional for loop.

```java
public boolean isPkerNear() {
    List<String> weapons = Arrays.asList(
            "staff", "shadow", "wand", "sceptre", "ballista",
            "crossbow", "dragon dagger", "dragon claws", "dark bow"
    );

    return Rs2Player.getPlayersInCombatLevelRange().stream()
            // This log message may get spammy, we could add a DEBUG log level
            .peek(player -> Microbot.log("There may be a pker around us " + player.getName(), Level.DEBUG))
            .anyMatch(player -> {
                String weaponName = Rs2Player.getPlayerEquipmentNames(player).get(KitType.WEAPON);
                if (weaponName != null) {
                    // This may also be spammy if alot of players are nearby, adding DEBUG log level
                    Microbot.log("They have a " + weaponName, Level.DEBUG);
                    String lower = weaponName.toLowerCase();
                    if (weapons.stream().anyMatch(weap -> lower.contains(weap.toLowerCase()))) {
                        Microbot.log("This player is wielding a " + weaponName + " which is used to pk so we're outy.");
                        return true;
                    }
                }

                Actor target = player.getInteracting();
                if (target instanceof Player) {
                    Rs2PlayerModel targetPlayer = new Rs2PlayerModel((Player) target);
                    // Check if the player found nearby is currently interacting w/ us, by leveraging player ID
                    boolean attackingUs = targetPlayer.getId() == Rs2Player.getLocalPlayer().getId();

                    if (attackingUs) {
                        Microbot.log(targetPlayer.getName() + " is attacking us");
                        return true;
                    }
                }


                return false;
            });
}
```

runelite-client/src/main/java/net/runelite/client/plugins/microbot/revKiller/revKillerScript.java
```diff
@@ -0,0 +1,918 @@
...
+
+    public boolean run(revKillerConfig config) {
+        this.config = config;
+        Microbot.enableAutoRunOn = false;
+        mainScheduledFuture = scheduledExecutorService.scheduleWithFixedDelay(() -> {
+            try {
+                if (!Microbot.isLoggedIn()) return;
+                if (!super.run()) return;
+                long startTime = System.currentTimeMillis();
+
```
gmason0, line 85: If you decide to take the suggestion of leveraging a boolean like `shouldFlee` we would place this at the top of this thread to prevent further execution

```java
if (shouldFlee) return;

File: runelite-client/src/main/java/net/runelite/client/plugins/microbot/revKiller/revKillerScript.java [hunk 9]
```diff
@@ -0,0 +1,918 @@
...
+            super.shutdown();
+        }
+    }
+
+    public void WalkToRevs(){
+        drinkStamPotion();
+        if(!WeAreInTheCaves()){
+            //we must walk to the cave entrence
+            if(Rs2Player.getWorldLocation().distanceTo(cave) > 6){
+                if(isPkerAround()){
```
gmason0: We wouldn't want this here, considering we are leveraging the other future to handle PK Fleeing & Detection

runelite-client/src/main/java/net/runelite/client/plugins/microbot/revKiller/revKillerScript.java
```diff
@@ -0,0 +1,918 @@
...
+                    if(Rs2GameObject.interact(31555, "Enter")){
+                        sleepUntil(()-> Rs2Dialogue.isInDialogue(), generateRandomNumber(1000,3000));
+                    }
+                }
+                if(Rs2Dialogue.isInDialogue()){
+                    if(Rs2Dialogue.hasContinue()){
+                        Rs2Dialogue.clickContinue();
+                        sleep(500,1000);
+                    }
+                    if(Rs2Dialogue.getDialogueOption("Yes", false)!=null){
```
gmason0, line 208: Why are we duplicating the options here?

runelite-client/src/main/java/net/runelite/client/plugins/microbot/revKiller/revKillerScript.java
```diff
@@ -0,0 +1,918 @@
...
+                        Rs2Dialogue.clickOption("accept", false);
+                        sleep(500,1000);
+                    }
+                }
+            }
+        } else {
+            if(WeAreInTheCaves()){
+                if(isPkerAround()){
+                    getAwayFromPker();
+                    return;
```
gmason0: Same here, we would want to keep leveraging the secondary future to check & flee away from PKers

runelite-client/src/main/java/net/runelite/client/plugins/microbot/revKiller/revKillerScript.java
```diff
@@ -0,0 +1,918 @@
...
+
+    public boolean TeleTimerIsThere(){
+
+        if(!Rs2Widget.isWidgetVisible(162, 558)){
+            return true;
+        }
+
+        return false;
+    }
+    public void getAwayFromPker(){
```
gmason0, line 235: This logic is overly complex & most of it is un-needed. The web walker knows wilderness levels when using teleports & will only use teleport items from the inventory. this could be simplified to something like 

```java
private void getAwayFromPker() {
    if (Rs2Player.isTeleBlocked()) {
        Rs2Walker.disableTeleports = true;
        Rs2Walker.walkTo(BankLocation.FEROX_ENCLAVE.getWorldPoint());
    } else {
        Rs2Walker.walkTo(BankLocation.EDGEVILLE.getWorldPoint());
    }


    if (Rs2Walker.disableTeleports) {
        Rs2Walker.disableTeleports = false;
    }
    
    shouldFlee = false;
}
```

runelite-client/src/main/java/net/runelite/client/plugins/microbot/revKiller/revKillerScript.java
```diff
@@ -0,0 +1,918 @@
...
+        }
+    }
+
+    public void Bankfortrip(){
+        if(!Rs2Bank.isOpen()){
+            if(WeAreInTheCaves()){
+                Microbot.log("Teleing to bank");
+                WorldPoint safe1 = (new WorldPoint(3199, 10071, 0));
+                WorldPoint safe2 = (new WorldPoint(3226, 10067, 0));
+                if(Rs2Pvp.getWildernessLevelFrom(Rs2Player.getWorldLocation()) > 20){
```
gmason0, line 576: Similar to what I had mentioned above, there is no need to constantly check the wilderness level. The Webwalker knows how to leverage teleports based on wilderness level & will walk to the wilderness line in-order to use the proper teleport

runelite-client/src/main/java/net/runelite/client/plugins/microbot/revKiller/revKillerScript.java
```diff
@@ -0,0 +1,918 @@
...
+        } else {
+            if(!Rs2Player.isInCombat()) {
+                Microbot.log("No revs found, hopping");
+                sleepUntil(()-> isPkerAround(), generateRandomNumber(0,1200));
+                hopToNewWorld();
+            }
+        }
+    }
+
+    private void startPkerDetection() {
```
gmason0: This is a good start, but there is already a secondary future inside of the script super class, by leveraging one of these, you can ensure proper shutdown/canceling logic.

From Script.java
```java
    protected ScheduledFuture<?> scheduledFuture;
    protected ScheduledFuture<?> mainScheduledFuture;
```

So for example, what I would do here is 

```java

private boolean shouldFlee; // initialized as false if no value is defined

public void startPkerDetection() {
    scheduledFuture = scheduledExecutorService.scheduleWithFixedDelay(() -> {
        try {

            if (!shouldFlee) {
                // Prevent checking if we should flee if we are not in the cave
                if (!isInsideCave()) return;
                shouldFlee = isPkerNear();
                return;
            }

            getAwayFromPker();
        } catch (Exception e) {
            Microbot.logStackTrace(getClass().getSimpleName(), e);
        }
    }, 0, 100, TimeUnit.MILLISECONDS);
}
```

By using scheduledFuture instead of creating your own, it will be properly canceled/shutdown when calling the script.shutdown() method in the plugin class.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/revKiller/revKillerScript.java
```diff
@@ -0,0 +1,918 @@
...
+                selectedArrow = config.selectedArrow().getArrowID();
+                // set it to our script
+                revimp = selectedWP;
+                randomdelay = generateRandomNumber(400,600);
+
+                DidWeDie();
+
+                if(areWeEquipped()){
+
+                    if(checkForPKerFuture == null || checkForPKerFuture.isDone() || checkForPKerFuture.isCancelled()){
```
gmason0: Instead of starting the thread here, we can call this method from the startUp method in the plugin class

```java
script.run()
script.startPkerDetection();
```

gmason0 (commented):
These are just some more suggestions / clean up suggestions.

Another general rule of thumb that I am seeing scattered around this code is it does not follow proper casing conventions used in java. When defining variable names or names of methods, they should follow camelCase. Not really an issue, just was glaring to me

runelite-client/src/main/java/net/runelite/client/plugins/microbot/revKiller/revKillerScript.java
Gage307: Good god how could I forget lmao. This is soooo much better. Thank you!

runelite-client/src/main/java/net/runelite/client/plugins/microbot/revKiller/revKillerScript.java
Gage307: Ahhh that's really cool I didn't know this was an option. This is something I can utilize with my barrows script as well. Thank you for letting me know!

runelite-client/src/main/java/net/runelite/client/plugins/microbot/revKiller/revKillerScript.java
Gage307: This is old code back when the PK check was just a simple if statement rather than a future. The idea was if a pker walked in while the script was eating the sleepUntil would stop sleeping and pass over to isPkerAround. Not the best solution but it worked at the time 😂

runelite-client/src/main/java/net/runelite/client/plugins/microbot/revKiller/revKillerScript.java
Gage307: It's that easy? Huh. 🤦

runelite-client/src/main/java/net/runelite/client/plugins/microbot/revKiller/revKillerScript.java
Gage307: I see what a beauty! For me, a lil confusing I'll need to stare at this for a while (no pun intended) { }

runelite-client/src/main/java/net/runelite/client/plugins/microbot/revKiller/revKillerScript.java
Gage307: Haha this was before .tolowercase I'll change it now

runelite-client/src/main/java/net/runelite/client/plugins/microbot/revKiller/revKillerScript.java
Gage307: While I was making the script I did assume that's how the walker would handle it. When it came to practice the walker while in the rev caves says destination unreachable. Regardless of whether it has a teleport or not.

gmason0: From the sounds of it then, the web walker must be missing transports for the staircases/exits that would prevent it from pathfinding using the teleport.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/revKiller/revKillerScript.java
Gage307: Got it. This is all such excellent info I'm going to need to refer back to this PR for some time lmao. Thank you so much george!

runelite-client/src/main/java/net/runelite/client/plugins/microbot/revKiller/revKillerScript.java
Gage307: Ohhh yes that makes a lot more sense!

runelite-client/src/main/java/net/runelite/client/plugins/microbot/revKiller/revKillerScript.java
Gage307: For sure. I didn't want to bother you with it since I could just work around it.

gmason0: I'll get it fixed, no worries

gmason0: As I communicated to you via Discord, the transports for the rev caves have been added

runelite-client/src/main/java/net/runelite/client/plugins/microbot/revKiller/revKillerScript.java
FireStarter-Instigator: > Got it. This is all such excellent info I'm going to need to refer back to this PR for some time lmao. Thank you so much george!

george been meaning to make a rev plugin for a while so it's something of a soft spot for him, great job with it

PR #897, opened 2025-05-01
Add "useExactNaming" to Rs2Shop.java

SteffenCarlsen:
Rs2Shop does not work in areas with multiple characters with similar names. Example "Shantay pass" if you want to buy bronze bars from Shantay. The BankShopper script will constantly try to interact with the Shantay Guard.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/shop/Rs2Shop.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/shop/Rs2Shop.java
@@ -52,11 +52,11 @@ public class Rs2Shop {
      *
      * @return
      */
-    public static boolean openShop(String NPC) {
+    public static boolean openShop(String NPC, boolean useExactNaming) {
         Microbot.status = "Opening Shop";
         try {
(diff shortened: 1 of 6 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/shop/Rs2Shop.java
```diff
@@ -52,11 +52,11 @@ public static boolean isOpen() {
      *
      * @return
      */
-    public static boolean openShop(String NPC) {
+    public static boolean openShop(String NPC, boolean useExactNaming) {
```
gmason0: Instead of having to adjust where the openShop method is used all throughout the project, we can create an overloaded implementation of `openShop(String npc)`

```java
public static boolean openShop(String Npc) {
    return openShop(npc, false);
}

public static boolean openShop(String NPC, boolean exact) {
// implementation
}
```

Another thing, can we change the name of this param to `exact`? just follows the method signatures of other classes like Rs2Inventory or Rs2Bank

gmason0 (changes requested):

PR #914, opened 2025-05-07
Moss killer plugin updates - making it schedulable 

FireStarter-Instigator:
Main update:

- makes moss killer schedulable 
- makes moss killer regear if neccesary when scheduled start 
- scheduled stop from scheduler gives one minute to put bot in safe location
- fixes some chaos altar bugs
- slight improvement to logic with anti-pk in wildy killer script

to be added in later release: ge rebuyer for consumables or absence/change in gear


Voxsylvae:
I roughly looked through your implemented way of making the plugin schedule. I want to say, it's not safe in the way you implemented it. I would not recommend using it in that way with the scheduler.
The first problem I see, you can't ensure it starts in your designed areas. Second, it can be and will be stopped at any time.alao during combat. I will give you an example of how I think you can properly implement the getStopConditions and getStartConditions interfaces if you like. Or I could also implement a "first" version of making it scheduled if you like. I would too it asap, if you wish.

FireStarter-Instigator:
> I roughly looked through your implemented way of making the plugin schedule. I want to say, it's not safe in the way you implemented it. I would not recommend using it in that way with the scheduler. The first problem I see, you can't ensure it starts in your designed areas. Second, it can be and will be stopped at any time.alao during combat. I will give you an example of how I think you can properly implement the getStopConditions and getStartConditions interfaces if you like. Or I could also implement a "first" version of making it scheduled if you like. I would too it asap, if you wish.

not sure what you mean, I just put what worked for the chicken plugin test I did, meaning the scheduler recognises moss killer as schedulable and is able to turn it on and off when stop conditions are met or when it's scheduled. I put in draft now anyway until it's ironed out.

FireStarter-Instigator:
@Voxsylvae i've just tried the scheduler out with your new updates and works amazing. Can you help me to setup now preparation methods for Moss Killer Plugin for it to work with the 3 modes that moss killer plugins has?

for normal/basic/boss mode I imagine inventory setups would get involved (hopefully easy for users to implement and to be able to do an equipment check on) I have never deal with inventory setups so I don't really know. generally the plugin is used that you just turn it on with armour and weapons on and it plays automatically from there so ideally we can sort out all that equipment and regearing check in however scheduler will be organising onstartup methods.  

for wildysafer and wildykiller they both have set outfits and integrated checks if they're not wearing those outfits so I think they will cover themselves but I will review them.

I will also likely add a system for regaining consumables if you run out, making it less likely that the EXIT_SCRIPT softstop will trigger and so that it can stay schedulable for longer times. 

I attached photo of the config just for reference 
Screenshot 2025-05-08 224120

Edit: I know that soon will be added to be able to force configs (I think actually this could be done on startup method, I have been experimenting with it succesfully for my pvp script (turning on and off combat modes for example through an in-game command)) however for now for the sake of having functioning in scheduler today I want to just implement it that the user decides which mode they wanna do for moss killer plugin by pre-setting up the configuration panel of moss killer and the scheduler turning it off and on when scheduled or conditions met etc. As Moss Killer Plugin is right now in this PR it works excellently with scheduler I have been testing.

FireStarter-Instigator:
> The first problem I see, you can't ensure it starts in your designed areas. Second, it can be and will be stopped at any time.alao during combat. I will give you an example of how I think you can properly implement the getStopConditions and getStartConditions interfaces if you like. Or I could also implement a "first" version of making it scheduled if you like. I would too it asap, if you wish.

Edit: posted in the discord my ideas/draft after reviewing #934 - also thanks for your responses there ik what I should be doing now

chsami (approved):

PR #934, opened 2025-05-10
Added schedulable Pest Control

subinznz:
This is opened so that other community members can take a look , 

I have added 
- Inventory setup
- world hop 
- Lock state

Note there are still bugs that needs to ironed out with scheduler , so just submitting for initial review . 

Currently requires a secondary condition to be set in stop condition when running as default plugin. Vox will be fixing this issue in new PR.  So don't merge until this is all finalised.


FireStarter-Instigator:
thanks, I needed this!

chsami (approved):

PR #941, opened 2025-05-13
handle safeStop better while lock mechanism is getting fixed 

subinznz:
this makes sure that other plugins dont immediately exit the pest control during PC game session , improved world hopping as well.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/pestcontrol/PestControlPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/pestcontrol/PestControlPlugin.java
@@ -4,11 +4,14 @@ import com.google.inject.Provides;
 import lombok.extern.slf4j.Slf4j;
 import net.runelite.api.ChatMessageType;
 import net.runelite.api.events.ChatMessage;
+import net.runelite.api.events.GameTick;
+import net.runelite.api.widgets.ComponentID;
 import net.runelite.client.config.ConfigManager;
 import net.runelite.client.eventbus.Subscribe;
 import net.runelite.client.plugins.Plugin;
 import net.runelite.client.plugins.PluginDescriptor;
 import net.runelite.client.plugins.microbot.Microbot;
+import net.runelite.client.plugins.microbot.nmz.NmzScript;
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/microbot/pestcontrol/PestControlPlugin.java
```diff
@@ -74,6 +79,10 @@ public LogicalCondition getStopCondition() {
     public void onPluginScheduleEntrySoftStopEvent(PluginScheduleEntrySoftStopEvent event) {
         if (event.getPlugin() == this) {
             Microbot.log("Scheduler about to turn off Pest Control");
+            if(pestControlScript.isInPestControl()) {
+                Microbot.log("Waiting to finish current game");
+                Global.sleepUntil(pestControlScript::isOutside, 1200000);
```
gmason0, line 84: Sleeping on the client thread can cause freezes to the client.

gmason0 (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/pestcontrol/PestControlPlugin.java
subinznz: I initially tried to do what NMZ does with  using runOnSeperateThread  but it didnt work :'(.

Anyways the lock condition should be fixed so it's best to wait for that so I dont have to handle this in the first place. 

            Microbot.getClientThread().runOnSeperateThread(() -> {
                if(!nmzScript.isOutside()) {
                    Rs2GameObject.interact(26276, "Drink");
                    Global.sleepUntil(nmzScript::isOutside, 10000);
                }
                Microbot.stopPlugin(this);
                return true;
            });

gmason0: If you're wanting to wait for the scheduler to be fixed, maybe we close this until then?

subinznz: yep

PR #943, opened 2025-05-13
add lockstate changes to GOTR script so it breaks between matches

crispycritter99:
implemented in the current onChatMessage. Locks when the script begins to mine 30 seconds before the game start and unlocks when the game is over.


runelite-client/src/main/java/net/runelite/client/plugins/microbot/runecrafting/gotr/GotrPlugin.java
```diff
@@ -104,11 +107,13 @@ public void onChatMessage(ChatMessage chatMessage) {
         }
 
         if (msg.contains("The rift becomes active!")) {
+            BreakHandlerScript.setLockState(true);
```
gmason0: We should be checking if the BreakHandler Plugin is enabled

```
if (Microbot.isPluginEnabled(BreakHandlerPlugin.class)) {
// do xyz
}
```

gmason0 (changes requested):

gmason0 (approved):

PR #944, opened 2025-05-13
Monk Killer Plugin

FireStarter-Instigator:
### FOR DEFENCE PURES

1. Arrives at Monk's in Edgeville from anywhere in the game
2. Attacks Monks when arrives
3. Re-engages with Monk if the monk's health bar disappears
4. Hops world (logs out) if no Monks are available to attack
5. Walks outside of combat zone when breakhandler is about to break (for logout)
6. Script shutdown when defence level is reached and turns off breakhandler and autologin

https://github.com/user-attachments/assets/5e1d8a36-0b73-4ea6-b793-c668ac2eb98f


FireStarter-Instigator:
For example, let's say this is my attackMonk method, how would I get it to stop using ActorModel.java getInteracting and start using Rs2NpcModel.java getInteracting? _Edit: Rs2NpcModel.Java does not have a getInteracting. Therefore I need to use Rs2Npc??_

  private boolean attackMonk(Rs2NpcModel monk) {
        if (!underAttack()
                && monk.getInteracting() != Microbot.getClient().getLocalPlayer()
        && monk.getInteracting() == null) {
            Microbot.log("found a monk, attacking");
            Rs2Npc.attack(monk);
            return true;
        }
        return false;
    }
    
   @chsami @g-mason0

FireStarter-Instigator:
Screenshot 2025-05-16 131145
because even Rs2Npc uses ActorModel.Java getInteracting - so if I was to wrap it in Rs2Npc, Rs2NpcModel, it would still eventually be using ActorModel.Java and create freezing

Screenshot 2025-05-16 131502
I guess wrapping it like that?, turning the ActorModel into an Rs2NpcModel.... so that getinteracting is covered by raw runelite api, essentially, but when any interaction comes a new rs2npcmodel is made from it. Would appreciate feedback/understanding.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/bee/monkscript/MonkKillerScript.java
```diff
@@ -0,0 +1,205 @@
...
+        walkTo(monkPoint);  // Walk to a specific point in the monk area
+    }
+
+    private Rs2NpcModel findAvailableMonk1() {
+        Rs2PlayerModel localPlayer = Rs2Player.getLocalPlayer();
+        List<Rs2NpcModel> monks = Rs2Npc.getNpcs("Monk").collect(Collectors.toList());
+
+        for (Rs2NpcModel monk : monks) {
+            if (monk.getInteracting() != null &&
+                    monk.getInteracting().getName().equals(localPlayer.getName())) {
```
gmason0: This

runelite-client/src/main/java/net/runelite/client/plugins/microbot/bee/monkscript/MonkKillerScript.java
```diff
@@ -0,0 +1,205 @@
...
+                        sleep(5000,15000);
+                        if (!underAttack()) {hopWorld();}
+                    }
+                } else if (!isInMonkArea()){
+                    walkToMonkArea();
+                } else {
+                    Rs2NpcModel monk1 = findAvailableMonk1();
+                    if (monk1 != null) {
+                        boolean isAttackingMe = monk1.getInteracting() != null &&
+                                monk1.getInteracting().getName().equals(Microbot.getClient().getLocalPlayer().getName());
```
gmason0: This

runelite-client/src/main/java/net/runelite/client/plugins/microbot/bee/monkscript/MonkKillerScript.java
FireStarter-Instigator: this what? this is how it's done or how it's not done?
Screenshot 2025-05-17 004558

runelite-client/src/main/java/net/runelite/client/plugins/microbot/bee/monkscript/MonkKillerScript.java
FireStarter-Instigator: > This

please be more specific as to what you're trying to point out

gmason0:
Regarding your comment about getInteracting isn't necessarily true of what is freezing up your script.. It would be the places I just pointed out above. 

Simply checking if the getInteracting() object from ActorModel is equal to another object, will not constitute as a freeze. 

Accessing properties that require the client thread such as getInteracting().getName() or getInteracting().getCombatLevel() WILL freeze the client as your are accessing a property that requires the client thread, while not being on the client thread. 


If want to leverage this getInteracting() method & check if they are interacting with the local player, here is a small example when interacting with a goblin. You can conform this to your script.

```java
private boolean isGoblinInteractingWithPlayer() {
    List<Rs2NpcModel> goblinsForPlayer = Rs2Npc.getNpcsForPlayer("Goblin");

    return !goblinsForPlayer.isEmpty();
}
```

https://github.com/user-attachments/assets/626119ff-1c60-4826-a42f-814d1404b461

FireStarter-Instigator:
ok so just use getnpcforplayer instead thanks, will update reverting actormodel changes and let you know if there are anymore freezes before submitting final version

gmason0:
Correct, either leverage  Rs2Npc.getNpcsForPlayer OR you can compare `getInteracting() == Microbot.getClient().getLocalPlayer()` if you REALLY want to do it yourself, but seems un-necessary as getNpcsForPlayers does the same thing

FireStarter-Instigator:
> ```java
> Rs2Npc.getNpcsForPlayer
> ```

the goblin one works for findavaialblemonk1 but for findavailablemonk I need to find attackable npcs, aka an npc that isn't interacting, or isn't interacting with my player if in multi, so i look at rs2npc attackable npcs and find this.... 
Screenshot 2025-05-17 010448
therefore if I was to use that it would just be passing the buck as it's the same line of code that's freezing client (note please that this is also the method used in AIO fighter's attack npc script) and it explains why aio fighter is reported getting many freezes, as it's the exact same as the "trouble" code you just pointed out in monk killer script

gmason0:
The method you pointed out above does not provided any freezes on my end.

FireStarter-Instigator:
> The method you pointed out above does not provided any freezes on my end.

you've ran the monk killer script?

Why would you point out it's what's causing freezes in my code, but then when you see it in rs2npc, and users complaining about freezes for aio fighter, it's suddenly a non-issue?

gmason0:
No, I was referring to you mentioning that getAttackableNPCs freezing the client. I have experienced no such thing

The current state of the monk killer script would cause freezes as your are comparing getInteracting().getName() to localplayer name. using the getName() method from getInteracting requires the client thread.

Simply comparing if getInteracting() is equal to another object as its done inside of getAttackableNpcs would not constitute a freeze as you're not accessing a property that requires to be ran on the client thread.

Do you have any evidence that it does? The burden of proof would be on your end.

gmason0:
> users complaining about freezes for aio fighter, it's suddenly a non-issue?

Not that its not an issue, there is just no evidence to back up the claims that this freezes the client when i can use AIO fighter for multiple hours without issue.

FireStarter-Instigator:
did not even notice that it was getting .getName involved lol

FireStarter-Instigator:
> > users complaining about freezes for aio fighter, it's suddenly a non-issue?
> 
> Not that its not an issue, there is just no evidence to back up the claims that this freezes the client when i can use AIO fighter for multiple hours without issue.

yeah you just gotta stress the resources a bit, run 4-10 clients using aio fighter at a time while you do your normal work flow and you should see issues. or just run 2-4 clients on a weaker machine.

gmason0:
> Do you have any evidence that it does? The burden of proof would be on your end.

FireStarter-Instigator:
> > Do you have any evidence that it does? The burden of proof would be on your end.

I can send you a timelapse of 10 aio fighter clients if you like and you can see them start to freeze. I also have many photos I uploaded onto discord of the freezes in earlier versions. There is also a bug report thread about it, someone posting just yesterday that is happens still. 
Screenshot 2025-05-14 193600

gmason0:
But according to you, what shows that getAttackableNpcs is the cause of this freeze that would require us to modify the actor model class?

AIO fighter is a very complex plugin with several running threads, so it can be a host of things. What proof shows that it is related to getInteracting comparison, that matches almost exactly to what is being done inside of getNPCsForPlayer, just inverted.

FireStarter-Instigator:
> But according to you, what shows that getAttackableNpcs is the cause of this freeze that would require us to modify the actor model class?

if i'm honest I still see freezes with aio fighter even with the proposed changes to actorModel, maybe a slight improvement but nothing to write home about

gmason0:
Thats my point exactly, if someone is still experiencing freezes with AIO fighter, its not related to either of the methods we are talking about.

FireStarter-Instigator:
> Thats my point exactly, if someone is still experiencing freezes with AIO fighter, its not related to this method we are talking about.

well we'll see, i'm part of that someone, I still experience freezes, it's just part of using aio fighter. i'll see if taking out the .getname stops the freezes with monk killer, again honestly had no idea it was doing that

gmason0:
Correct, either remove getInteracting().getName() & just compare getInteracting() w/ Microbot.getClient().getLocalPlayer() or simply use getAttackableNpcs

FireStarter-Instigator:
Screenshot 2025-05-17 014530
update lol

edit: 3 freezes so far with the new changes not impressed when previous version with actormodel modification was 100% stable

FireStarter-Instigator:
after all accounts froze overnight Kept changes to script class (removing .getname and other improvements to flow/more use of rs2npc methods) but modified again actormodel class, testing now over 12 hours for freezes

edit: no freezes after 3 hours with 3 clients, 100% cpu usage and 90% ram usage (high-stress and resource-heavy environment), continuing testing

gmason0:
I have been debugging with Duck to more things that can crash w/ AIO fighter (I know that AIO fighter is not the main topic here) & made a discovery

At first testing, I was using a level 3 account killing chickens in lumbridge & I was only about to go about 2 hours before I saw the client crash.

We then made changes to the AIO fighter plugin itself, still seeing crashes

After that, we made a slight adjustment to Rs2Combat.inCombat, I was then able to go a full 6hr log with no crashes

```java
    public static boolean inCombat() {
        if (!Microbot.isLoggedIn()) return false;
        if (Microbot.getClientThread().runOnClientThreadOptional(() -> !Microbot.getClient().getLocalPlayer().isInteracting() ||
                Objects.requireNonNull(Microbot.getClient().getLocalPlayer().getInteracting()).getCombatLevel() < 1).orElse(true)) return false;
        return Rs2Player.isInteracting() || Rs2Player.getAnimation() != -1;
    }
```

I am going to be reverting all other changes I made to AIOFighter & trying to leverage just these changes, let me know if this helps your crashing situation as this method is used inside of Rs2NPC.attack() which I had seen you're using here

FireStarter-Instigator:
> I am going to be reverting all other changes I made to AIOFighter & trying to leverage just these changes, let me know if this helps your crashing situation as this method is used inside of Rs2NPC.attack() which I had seen you're using here

interesting! InCombat is a goated method. I will take off the actormodel changes, put the isincombat changes and then will see if any freezes. 

edit: so far so good, usually at least 1 client freezes basically on first login when the code's problematic and they're all ok atm, will post back in few hours update

edit: sorry was a false start, no change, all 3 clients froze within an hour; will let you know if these isincombat changes WITH actormodel changes is ok

edit: not sure if related but bot doesn't activate findavailablemonk1 and to attack it when the healthbar of the monk dissappears and he's still in combat, when it worked before, so I added a failsafe incase re-attacking the monk doesn't work a bit like QoL neverlogout to stimulate auto-retaliate and also reverted incombat changes incase it wasn't registering if in combat as it was in previous versions

gmason0:
Can you push your latest script changes? Lets forget about the ActorModel changes as I am going to deny them out of the gate.

FireStarter-Instigator:
> Can you push your latest script changes? Lets forget about the ActorModel changes as I am going to deny them out of the gate.

yeah sure I can update it gimme a mo, only works with actormodel changes though ye gatekeeper

FireStarter-Instigator:
I've made some more changes, mostly relabelling monk1 as attackingMonk and putting conditions such as no defence xp drop in 30 seconds before going into re-attacking attackingMonk block as it was firing in there too often only realised after I put logs. also was no point checking if they getinteracting with you twice, so just focused on health bar being there or not

FireStarter-Instigator:
> Lets forget about the ActorModel changes as I am going to deny them out of the gate.

@g-mason0 why don't we add a method to actorModel.java called getInteractingModel for these use cases and leave getInteracting alone ?

FireStarter-Instigator:
Screenshot 2025-05-18 055635
with this, I would just use getInteractingModel in monk killer instead of getInteracting and eliminate freezes and it would be safe to merge

gmason0:
As I mentioned let's forget about the actor model changes as it's more of a problem with your plugin, not the actor model. What you've implemented is a workaround instead of troubleshooting the root issue. 

I am going to commit the changes today for AIO fighter today so you can see that'll go for a full 6hr log without crashing.

FireStarter-Instigator:
> As I mentioned let's forget about the actor model changes as it's more of a problem with your plugin, not the actor model. What you've implemented is a workaround instead of troubleshooting the root issue.
> 
> I am going to commit the changes today for AIO fighter today so you can see that'll go for a full 6hr log without crashing.

Don't wanna step on your toes but we've gone over this that my plugin now uses .getInteracting in a normal way unless you can show otherwise real quick and i'll change it, if it's not too bothersome?

Edit: if this in fact not a fix to a root issue but a workaround like you say, then you've acknowledged that there is an issue, and until the issue is resolved then this is the fix. so when the issue is resolved I'll just test it to confirm and change it in a moment's notice at nobody's inconvenience.

FireStarter-Instigator:
Screenshot 2025-05-18 175132
Screenshot 2025-05-18 175148
for your convenience, here are the only usages in order to evaluate the problem lying in my plugin (if you still believe the root issue is in my plugin? (even though you are admitting and have shared attempts publicly to push changes in util classes to fix similar issue that arise in AIO fighter)).

Edit: just as a reminder, if I use normal getInteracting instead of getInteractingModel I get freezes (in the current state of the script)

FireStarter-Instigator:
@g-mason0 from what i'm seeing from your commits (for some reason you blocked commenting on one of them?) the issue may be with monk.getInteracting == || =! localPlayer and to fix to put an object qualifier?

Screenshot 2025-05-19 180039

gmason0:
Correct, using Objects.equals() provides better null safety compared to using direct equals comparison. 

With the changes I've made to Rs2NPC, Rs2Combat & slight modifications to AIO Fighter, I've been able to do several hours without crashing & without modifying the ActorModel class. I have also tested this in a "resource heavy environment" as you alluded to where the crashing seems to happen more often. My testing environment has been me running two runelite clients, the one in the screenshot below is running in a complied jar & I have my main account running through the IDE, then I also am playing counter-strike 2 at the same time. This pushes my computer to about ~40% CPU usage & ~85% Memory usage.

image

Now that we have a similar understanding here & AIO fighter issues have been resolved. What is the point of this plugin? AIO Fighter can do all the things mentioned above except world hop if no NPCs are found. With that being said, I feel this falls under the category of duplicate functionality

FireStarter-Instigator:
> Now that we have a similar understanding here & AIO fighter issues have been resolved. What is the point of this plugin? AIO Fighter can do all the things mentioned above except world hop if no NPCs are found. With that being said, I feel this falls under the category of duplicate functionality

There are 2 key factors about fighting monks as a defence pure
1. To attack the npc when the healthbar dissapears to re-initiate combat (auto-retaliate and "attacking" fails after a set amount of time while runelite still registers as mutually interacting with each other)
2. As you mentioned, to hop when no monks are available (as people usually fight monks indefinitely) as it's an afk activity

In terms of QoL of the plugin that AioFighter does not support: 
1. To walk to a safespot to exit combat when breakhandler logout is about to be called
2. Hardcoded to go to monk area if not at the area (scheduler compliant)

I am seeing to add an element such as an f2p prayer training script, if this mode is selected to fire randomly or to override, it will go to the ge, purchase and scatter fiendish ashes until a prayer level is reached or an amount of ashes scattered and then return to killing monk. This fits with the monk theme and also helps to spread out really repetitive activities such as f2p prayer with an afk activity. Also defence pures with high prayer are really goated builds. 

As mentioned in the config this is not intended as a fighter plugin but an afk sustainer plugin for defence pure training. Therefore an AIO doesn't hit the specific requirements.

gmason0:
I appreciate you taking the time to articulate the features of this plugin. From your reasoning, I really do not see a need for maintaining overlapping features within the client. We could be improving the existing with some of the features you had mentioned

1. World Hopping
2. Leveraging the Health Ratio, not only getInteracting() inside of the AttackNpcScript
3. AIO Fighter already has skilling goals if you want to only focus defense based combat skills 
4. Going somewhere when a break is about to start

Your reasoning about going to the desired area when using the scheduler is already a planned feature: https://github.com/chsami/Microbot/blob/development/docs/scheduler/roadmap.md?plain=1#L30

If you would truly like to keep the plugin, as is, I think it would be a good candidate as a sideloadable plugin.  https://github.com/chsami/Microbot/blob/development/runelite-client/pom.xml#L472, adjust the paths here to where your package is in the microbot repo & define the "finalname".

Again appreciate the time of making this plugin & discussing the related issues with AIO fighter, but I find this plugin too overlapping with existing features

FireStarter-Instigator:
> I appreciate you taking the time to articulate the features of this plugin. From your reasoning, I really do not see a need for maintaining overlapping features within the client. We could be improving the existing with some of the features you had mentioned
> 
> 1. World Hopping
> 2. Leveraging the Health Ratio, not only getInteracting() inside of the AttackNpcScript
> 3. AIO Fighter already has skilling goals if you want to only focus defense based combat skills
> 4. Going somewhere when a break is about to start
> 
> Your reasoning about going to the desired area when using the scheduler is already a planned feature: https://github.com/chsami/Microbot/blob/development/docs/scheduler/roadmap.md?plain=1#L30
> 
> If you would truly like to keep the plugin, as is, I think it would be a good candidate as a sideloadable plugin. https://github.com/chsami/Microbot/blob/development/runelite-client/pom.xml#L472, adjust the paths here to where your package is in the microbot repo & define the "finalname".
> 
> Again appreciate the time of making this plugin & discussing the related issues with AIO fighter, but I find this plugin too overlapping with existing features

ok bro, you can tell you've never made a defence pure lol, aio fighter would not work, unless i guess just copy paste the script and have it run in "defence pure mode" in aio figher

gmason0:
I probably have not made a defense pure, but i can read & understand code. The most primitive aspects to this plugin are 

1. Travel to Monk Location
2. Attack Monk
3. If we aren't interacting OR the health ratio goes to -1 re-attack a monk
4. Run away when a break is about to start
5. World Hopping no monks are found 

All these things can be generically implemented into AIO fighter if effort is put in, not just simply dropping your script  & making a config option

And like I said, if you don't feel like doing that, sideloading is always an option !

PR #947, opened 2025-05-13
Rs2Bank and Rs2Prayer Update

slest1:
#### Bank Collection Box Interaction (Rs2Bank.java)
- **New methods to manage the Bank Collection Box:**
    - `openCollectionBox()`: Opens the collection box by interacting with the nearest bank, GE booth, or banker.
    - `bankAllCollectionBoxItems()`: Opens the collection box and banks all items.
    - `inventoryAllCollectionBoxItems()`: Opens the collection box and moves all items to inventory.
    - `closeCollectionBox()`: Closes the collection box interface.
    - `collectionBoxIsOpen()`: Checks if the collection box is currently open.

#### Prayer Management (Rs2Prayer.java)
- **New `swapOverHeadPrayer(Rs2PrayerEnum prayer)` method:**
    - Efficiently swaps the active overhead protection prayer. If the target prayer is already active, it does nothing. Otherwise, it deactivates any current overhead prayer and activates the new one.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/bank/Rs2Bank.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/bank/Rs2Bank.java
@@ -1281,7 +1281,7 @@ public class Rs2Bank {
     public static boolean bankAllCollectionBoxItems() {
         openCollectionBox();
         sleepUntil(Rs2Bank::collectionBoxIsOpen, 5000);
-        Rs2Widget.clickWidget("Bank");
+        Rs2Widget.clickWidget(26345476);
         return true;
     }
     /**
@@ -1293,7 +1293,6 @@ public class Rs2Bank {
     public static boolean inventoryAllCollectionBoxItems() {
         openCollectionBox();
         sleepUntil(Rs2Bank::collectionBoxIsOpen, 5000);
-        Rs2Widget.clickWidget("Inventory");
         return true;
     }
 
@@ -1301,7 +1300,8 @@ public class Rs2Bank {
      * Closes the collection box interface.
      */
     public static void closeCollectionBox() {
-        Rs2Widget.clickWidget("Close");
+        Widget[] closeWidget = Rs2Widget.getWidget(402,2).getDynamicChildren();
+        Rs2Widget.clickWidget(closeWidget[3]);
         sleepUntil(() -> !collectionBoxIsOpen(), 5000);
     }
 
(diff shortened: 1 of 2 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/bank/Rs2Bank.java
```diff
@@ -1215,6 +1210,108 @@ public static boolean openBank() {
...
+    /**
+     * Collects items from the collection box and deposits them into the bank.
+     *
+     * @return true if the operation is successfully initiated and completes processing.
+     */
+
+    public static boolean bankAllCollectionBoxItems() {
+        openCollectionBox();
+        sleepUntil(Rs2Bank::collectionBoxIsOpen, 5000);
+        Rs2Widget.clickWidget("Bank");
```
gmason0: We should fetch the widget object here & click that or define the widget Ids, using clickWidget(string) like this can lead to clicking messages in the chat if there was someone nefarious trying to type "bank" in the chat

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/prayer/Rs2Prayer.java
```diff
@@ -138,6 +138,17 @@ public static boolean isMeleePrayerActive() {
         ).anyMatch(Rs2Prayer::isPrayerActive);
     }
 
+    public static void swapOverHeadPrayer(Rs2PrayerEnum prayer) {
+        Rs2PrayerEnum activeProtectionPrayer = getActiveProtectionPrayer();
+        if (activeProtectionPrayer == prayer) {
+            return;
+        }
+        if (activeProtectionPrayer != null) {
```
gmason0: If we already have a protection prayer enabled & we attempt to switch to a different protection prayer, we don't need to disable the current one as you can only have 1 protection prayer enabled

gmason0 (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/bank/Rs2Bank.java
slest1: makes sense. Will update this

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/prayer/Rs2Prayer.java
slest1: yea oversight on my part, will push an update soon.

slest1:
Should be good to merge.

gmason0 (approved):

PR #949, opened 2025-05-14
ActorModel.java update to getInteracting() to prevent freezes/crashes

FireStarter-Instigator:


gmason0:
As I mentioned to you before. there are other ways to compare getInteracting() instead of using ActorModel.getInteracting(). Please use those strategies over modifying the actormodel class

FireStarter-Instigator:
> As I mentioned to you before. there are other ways to compare getInteracting() instead of using ActorModel.getInteracting(). Please use those strategies over modifying the actormodel class

there's other ways to skin a cat doesn't mean one is better than the other, until you offer a working solution for not getting freezes using getinteracting in actormodel then this should stay

PR #951, opened 2025-05-14
Updates actorModel.Java getInteracting() to stop freezes/crashes

FireStarter-Instigator:


gmason0:
This is a breaking change for other comparisons that already exist in the client, that compare against a raw actor in an area where the client thread is being used. This change is only required due to the way YOU are comparing getInteracting()

FireStarter-Instigator:
> This is a breaking change for other comparisons that already exist in the client, that compare against a raw actor in an area where the client thread is being used. This change is only required due to the way YOU are comparing getInteracting()

hi please elaborate on the breaking changes, currently you're just making stuff up, this is actually a fixing and not breaking change

PR #952, opened 2025-05-14
Updates actorModel.Java getInteracting() to stop freezes/crashes

FireStarter-Instigator:
ActorModel.java is a class made by @g-mason0 
A lot of npc and player interactions involve using the getInteracting method presented in this class.
Daily people are complaining about freezes and crashes in the client particularly when dealing with npc's.

This update is a tested fix to the issue. And unless it's demonstrated as an actual breaking change should be considered to be merged and not immediately closed as it has been twice now. @chsami


gmason0:
There are places where this is a breaking change. Please stop re-opening the same pull request as I've already stated why this is not an acceptable changes. please use other methods to compare getInteracting with another object. 


https://github.com/chsami/Microbot/blob/main/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/npc/Rs2Npc.java#L126

https://github.com/chsami/Microbot/blob/main/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/npc/Rs2Npc.java#L107

https://github.com/chsami/Microbot/blob/main/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/npc/Rs2Npc.java#L321

https://github.com/chsami/Microbot/blob/main/runelite-client/src/main/java/net/runelite/client/plugins/microbot/jad/JadScript.java

FireStarter-Instigator:
> There are places where this is a breaking change. 

I appreciate you sending me a list of instances where this method of get interacting is used, however this doesn't satisfy a proper to test to see if it actually breaks it or not and you have nothing to lose leaving it up as a draft as the community tries to work on this freezing and crashing issue to do with getInteracting.

gmason0:
I will continue to close these, I have all day

FireStarter-Instigator:
> I will continue to close these, I have all day

not very professional, looks ugly, and an abuse of power but yeah suite yourself, if you can offer 1 test I can do to demonstrate this breaking change i'll video it and prove your own point for you

FireStarter-Instigator:
and the silence was deafening

PR #961, opened 2025-05-14
Crashes from getInteracting resolved, patch to actorModel.java

FireStarter-Instigator:
The getInteracting method continues to be a major source of issues.
This PR includes a small, verified fix that significantly improves stability.
The burden of proof is now on those who want this closed without testing.


gmason0:

FireStarter-Instigator:

am not able to replicate the crashes you talk about, would you mine showing me how? which method do I invoke for instant crash? lmk

PR #962, opened 2025-05-14
Lightweight freeze fix for ActorModel.java getInteracting - tested & reproducible

FireStarter-Instigator:
Tested fix to the freezing issue caused by ActorModel.java.
Particularly affects bots interacting with NPCs.
@g-mason0 Immediately closes without proof or reproduction just delaying needed improvements.


gmason0:

This creates client crashes due to the way you are trying to compare getInteracting(), if you would like to run this in your local build by all means but this will break other mechanisms in our code base. So please stop re-opening the same changes.

FireStarter-Instigator:
> 
> This creates client crashes due to the way you are trying to compare getInteracting(), if you would like to run this in your local build by all means but this will break other mechanisms in our code base. So please stop re-opening the same changes.

sorry george am not able to replicate the crashes you talk about, would you mine showing me how? which method do I invoke for instant crash? lmk

PR #964, opened 2025-05-14
getInteracting crash fix. closed without review, again

FireStarter-Instigator:
Crash fix. Tested. Stable.
George's behaviour is blocking progress. Even just place for discussion. 
No explanation, just instant shutdowns.
That’s not leadership, that’s ego.


gmason0:

This creates client crashes due to the way you are trying to compare getInteracting(), if you would like to run this in your local build by all means but this will break other mechanisms in our code base. So please stop re-opening the same changes.

FireStarter-Instigator:
> 
> This creates client crashes due to the way you are trying to compare getInteracting(), if you would like to run this in your local build by all means but this will break other mechanisms in our code base. So please stop re-opening the same changes.

sorry george am not able to replicate the crashes you talk about, would you mine showing me how? which method do I invoke for instant crash? lmk

PR #972, opened 2025-05-14
Where’s the method, George? Still nothing. Fix remains valid

FireStarter-Instigator:
This is a working fix. If it breaks something, show me how.
All I’ve asked is for a method I can test.
Instead, I get silence and shutdowns. That’s not normal behavior.


gmason0:
I have already provided you with methods that cause this issue. also please stop creating commits of removing & adding a space. just shows you have 0 clue how to use git.

FireStarter-Instigator:
> I have already provided you with methods that cause this issue. also please stop creating commits of removing & adding a space. just shows you have 0 clue how to use git.

please share the method here and I will test it

PR #973, opened 2025-05-14
Asking for 1 test to prove me wrong, still getting ignored

FireStarter-Instigator:
Asking questions isn’t harassment. Submitting solutions isn’t wrong.
George’s pattern of closing PRs without feedback is the real problem.
Tell me what method makes it break. Otherwise, let this be considered.


gmason0:
You clearly aren't reading what I am telling you. I have already provided examples of what this can break. 

For example, the equality check inside of Rs2NPC's method for getAttackableNPCs, getNPCsForPlayer & other methods that compare the NPC's getInteracting with the player. 

https://github.com/chsami/Microbot/blob/development/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/npc/Rs2Npc.java#L107
https://github.com/chsami/Microbot/blob/development/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/npc/Rs2Npc.java#L126
https://github.com/chsami/Microbot/blob/development/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/npc/Rs2Npc.java#L157
https://github.com/chsami/Microbot/blob/development/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/npc/Rs2Npc.java#L321

I have already provided you with these examples but you are too hard headed to even read my response & insist this is the way to fix your problem, instead of realizing this is an issue of how you are utilizing the API.

FireStarter-Instigator:
you've provided examples of what it CAN break, you're not telling me methods to test with the actormodel.java changes that will result in freezes and crashes

gmason0:
Leverage other methods that are in the API to perform this instead of modifying the Actor Model

FireStarter-Instigator:
from what I can gather from what you shared, these are the methods that when run with the changes made in actorModel.java will break/freeze the client

Rs2Npc.getNpcsForPlayer()
Rs2Npc.getNpcsForPlayer()
Rs2Npc.getNpcsForPlayer()
Rs2Npc.getAttackableNpcs()

i'll go test them now

gmason0:
Not that they will freeze the client, that the object comparison will not pass.

gmason0:
Either way, clean up the commit history if you truely want to see this merged, I am not going to accept something that has 25 commits of you adding & removing a space because you're too antiquated to figure out how to just re-open a pull request

FireStarter-Instigator:
> Not that they will freeze the client, that the object comparison will not pass.

the methods will run without bricking the client but they will be inaccurate or return null/error is what you're trying to say?

FireStarter-Instigator:
I currently use these methods in my scripts with the changes to actormodel and I don't get any issues you can see line 125 in rs2npc if you ctrl click there is wildykiller script which I run every night on 4 different accounts for 6 hours, never any errors or problems with that method using actormodel changes

gmason0:
Like I have already mentioned, leverage other methods that are in the API to perform this instead of modifying the Actor Model

FireStarter-Instigator:
> Like I have already mentioned, leverage other methods that are in the API to perform this instead of modifying the Actor Model

to perfom what? i'm talking about use of Rs2Npc.getNpcsForPlayer() - the problematic method with these changes, I use this with actormodel changes and there's no issues

FireStarter-Instigator:
hmm i am noticing it returns an empty list though on example script testing

FireStarter-Instigator:
Screenshot 2025-05-14 190358
Screenshot 2025-05-14 190405

FireStarter-Instigator:
although when I change getinteracting back, it gives same result (I am fighting a bear at the time)

gmason0:
> hmm i am noticing it returns an empty list though on example script testing

This is exactly the point I am trying to make, making this adjustment breaks the predicate inside of the Rs2Npc methods when comparing getInteracting with another object as these are not the same object type anymore.

Therefore, you should change the implementations of how you are comparing getInteracting within your plugin instead of changing ActorModel

PR #974, opened 2025-05-14
Hi, please help me to devise a test for these changes. 

FireStarter-Instigator:
Screenshot 2025-05-14 190358
Screenshot 2025-05-14 190405

1. Rs2Npc.getNpcForPlayer (and it's 3 varieties)
2. Rs2Npc.getAttackableNpcs

these are the two methods pointed out would not work with these changes. when I do the test above, as an example, with changes to actor model or without changes, the result is the same.

please help me to design a test to investigate the effects of these changes to actormodel.java conclusively.


gmason0:
PLEASE for the love everything, I have already provided you a solution & its not to modify the actormodel class

PR #975, opened 2025-05-14
hi, no reason to automatically close this PR, this is an ongoing investigation, open to anybody into helping stop the client from freezing

FireStarter-Instigator:
here is the test i've conducted and it's results, having actormodel the same and having it changed like in the PR. there is no difference. please allow assistance in determining a proper test by keeping this PR open. 

Screenshot 2025-05-14 190358
Screenshot 2025-05-14 190405

Rs2Npc.getNpcForPlayer (and it's 3 varieties)
Rs2Npc.getAttackableNpcs
these are the two methods pointed out would not work with these changes. when I do the test above, as an example, with changes to actor model or without changes, the result is the same.

please help me to design a test to investigate the effects of these changes to actormodel.java conclusively.


gmason0:
closing due per the last 20 PRs messages. its not an API problem, its your plugin that is the issue.

PR #976, opened 2025-05-14
No reason to close. investigating client freezes, help welcome

FireStarter-Instigator:
I ran tests with the patch and without.
Result: No change. Methods like getNpcForPlayer and getAttackableNpcs behave the same.
I’m asking for help designing a better test, not being confrontational.

Screenshot 2025-05-14 190358
Screenshot 2025-05-14 190405


gmason0:
Please fix your plugin, its not a problem with the API

PR #980, opened 2025-05-14
Investigating freezing, proposed Actormodel.java changes

FireStarter-Instigator:
The screenshot shows the result of a "trouble" method that should be affected by the proposed actormodel changes. However when testing like in the screenshots I get same output with modified actormodel.java or unmodified and in practice I get less freezes and crashes when interacting with npc's and players with the modified version in my local repo. 

Help me understand what else is needed to validate or invalidate these proposed changes before it can become acceptable or dismissed. Very interested to improve the user experience of Microbot and to find solution to the freezing we keep seeing, most investigations bring it down to something happening in getInteracting() or getName(). 

Considering only recently we changed over to new model classes maybe these are the places to start looking and to test for. 

Screenshot 2025-05-14 190358
Screenshot 2025-05-14 190405


gmason0:
Pull requests are not used for troubleshooting, create an Issue on the repo if you would like to have an open discussion

FireStarter-Instigator:
> Pull requests are not used for troubleshooting, create an Issue on the repo if you would like to have an open discussion

fair, not troubleshooting, just wishing to have more evidence of your false claims and divorce them from your personal attacks, shame there is no fairness and justice here when you've got the wheel and that the community suffers for it - just keep the PR up in a draft for a week, allow me to do some tests, help me with them, to improve the client, but no, you can't get over yourself

PR #983, opened 2025-05-15
Motherload Miner: The script will now avoid mining too close to another player

Gage307:


gmason0:
Lets clean up the commit history here, 105 commits for only ~80 line changes is kinda overboard

PR #987, opened 2025-05-16
Support for Multiple Item Alching Agility Plugin 

NLG-code:
Added Support for Multiple Item Alching Agility Plugin 
image


runelite-client/src/main/java/net/runelite/client/plugins/microbot/agility/AgilityScript.java
```diff
@@ -75,8 +69,6 @@ private List<AgilityObstacleModel> getCurrentCourse(MicroAgilityConfig config) {
                 return varrockCourse;
             case GNOME_STRONGHOLD_AGILITY_COURSE:
                 return gnomeStrongholdCourse;
-            case CANIFIS_ROOFTOP_COURSE:
```
gmason0: Why are we removing this? We would want to keep the canafis course under getCurrentCourse

gmason0 (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/agility/AgilityScript.java
NLG-code: Added it back sorry abt that

gmason0 (approved):

PR #999, opened 2025-05-18
Add gem bag functionality

FunkyMonkeyCloud:
- Prior to the hopper deposit sequence, fill the gem bag. If inventory space was created by filling the gem bag, then return to mining, or else complete the hopper deposit sequence
- While banking, empty the gem bag directly to bank
- Do not deposit the gem bag while banking


gmason0 (changes requested):
Any reason why we aren't using Rs2Gembag class?

PR #1003, opened 2025-05-19
Fix for dodgy necklace only withdrawing correctly once and added new feature Automatically detection and wearing of rogue set - Thieving 1.6.5

NLG-code:
1. The issue should be because:

If Rs2Inventory.waitForInventoryChanges(3000); is not added before withdrawing dodgy necklace it withdraws the same amount of food with it adding this fixes the issue based on my testing

Also , withdraw deficit was not working properly hence changed to withdrawX to implement the appropiate withdrawal functionality and lowered the inventorychanges wait timings too

2. Added New feature : Automatically detection and wearing of rogue set

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/thieving/ThievingScript.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/thieving/ThievingScript.java
@@ -31,7 +31,7 @@ import static net.runelite.client.plugins.microbot.util.equipment.Rs2Equipment.i
 
 public class ThievingScript extends Script {
 
-    public static String version = "1.6.4";
+    public static String version = "1.6.5";
     ThievingConfig config;
 
     public boolean run(ThievingConfig config) {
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/microbot/thieving/ThievingScript.java
```diff
@@ -256,35 +256,51 @@ private void bank() {
...
+                        Rs2Inventory.waitForInventoryChanges(3000);
                         Rs2Bank.withdrawAll(true, "Earth rune", true);
-                        Rs2Inventory.waitForInventoryChanges(5000);
+                        Rs2Inventory.waitForInventoryChanges(3000);
                     }
                 }
             }
             Rs2Bank.withdrawAll(true, "Cosmic rune", true);
-            Rs2Inventory.waitForInventoryChanges(5000);
+            Rs2Inventory.waitForInventoryChanges(3000);
```
SteffenCarlsen, line 288: You can leverage Rs2Magic.canCast to ensure that we can actually use shadowveil at this point. If it returns false, we could disable the config option and let user know.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/thieving/ThievingScript.java
```diff
@@ -256,35 +256,51 @@ private void bank() {
...
         }
 
         boolean successfullyWithdrawFood = Rs2Bank.withdrawX(true, config.food().getName(), config.foodAmount(), true);
         if (!successfullyWithdrawFood) {
             Microbot.showMessage(config.food().getName() + " not found in bank");
-            sleep(5000);
+            sleep(3000);
```
SteffenCarlsen: if we don't have food, shouldn't we just shutdown the script?

runelite-client/src/main/java/net/runelite/client/plugins/microbot/thieving/ThievingScript.java
NLG-code: have push the new change to reduce the logic for handleShadowVeil function

runelite-client/src/main/java/net/runelite/client/plugins/microbot/thieving/ThievingScript.java
NLG-code: Have comitted shutdown instead

runelite-client/src/main/java/net/runelite/client/plugins/microbot/thieving/ThievingScript.java
```diff
@@ -256,35 +255,50 @@ private void bank() {
...
 
         boolean successfullyWithdrawFood = Rs2Bank.withdrawX(true, config.food().getName(), config.foodAmount(), true);
         if (!successfullyWithdrawFood) {
             Microbot.showMessage(config.food().getName() + " not found in bank");
-            sleep(5000);
-            return;
+            shutdown();
         }
 
-        Rs2Bank.withdrawDeficit("dodgy necklace", config.dodgyNecklaceAmount());
```
gmason0, line 287: What issues are you seeing with withdrawDeficit? 

```java
Rs2Bank.withdrawDeficit("teleport to house", 20);
Rs2Bank.withdrawDeficit("prayer potion", 8);
```

The code above did as expected & withdrew 10 house teleports & 2 prayer potions based on the number of items I already had in my inventory


https://github.com/user-attachments/assets/f5abfa2d-e1b5-4060-8982-7fe300a6da92

gmason0 (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/thieving/ThievingScript.java
NLG-code: the first access always works , but subsequent accessing of withdrawing the dodgy necklaces withdrawals all instead causing inventory to be filled no matter the config input amount.

NLG-code: so i first changed to withdrawX and had the same issue , after i added Rs2Inventory.waitForInventoryChanges(3000); before the withdrawal code and it got solved , so i think both withdrawdeficit and withdrawX would also work now if that line is swapped interchangeably

NLG-code: both implementations are pretty much the same for this case because all items are banked regardless

PR #1004, opened 2025-05-19
Added check for Event dismiss plugin to disallow starting if Agility plugin is active as it causes occassional bugging

NLG-code:
Bug found is :

When NPC is stuck behind another agility platform the player keeps trying to click it which forms a loop until the event is dismissed by itself i.e. a timeout. This happens a few times a day, therefore i added a check for Event dismiss plugin to disallow starting if Agility plugin is active.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/eventdismiss/EventDismissScript.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/eventdismiss/EventDismissScript.java
@@ -17,7 +17,6 @@ public class EventDismissScript extends Script {
     public boolean run(EventDismissConfig config) {
         List<Plugin> activePlugins = Microbot.getActiveMicrobotPlugins();
         for (Plugin plugin : activePlugins) {
-            Microbot.log(plugin.getName());
             if (plugin.getName().equals("<html>[<font color=#b8f704M>M</font>] Agility")) {
                 Microbot.log("Event Dismiss Start Failed : Cannot use event dismiss with Agility Plugin");
                 return false;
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/microbot/eventdismiss/EventDismissScript.java
```diff
@@ -1,19 +1,27 @@
...
-import java.util.Arrays;
+
+import java.util.List;
 import java.util.concurrent.TimeUnit;
 
 public class EventDismissScript extends Script {
     public static double version = 1.0;
-
     public boolean run(EventDismissConfig config) {
+        List<Plugin> activePlugins = Microbot.getActiveMicrobotPlugins();
```
SteffenCarlsen: Wouldnt it be better to fix the event dismiss plugin, so it can't loop indefinitely if it fails it's interaction? Something like:
1. Try to dismiss
2. If fail (Check distance to event/if event is still there), increase errorCount
3. If errorCount > 3 or something like that, skip dismissing this event and just let a script do its thing with the event following

gmason0: Like I mentioned to you in discord, I think a better fix here would be checking if you have line of sight to the random event npc

gmason0 (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/eventdismiss/EventDismissScript.java
NLG-code: heyo yup thanks! have attempt a fix still testing

gmason0:
Any update on this?

NLG-code:
> Any update on this?

this should fix it havent seen it get stuck yet for the past day

PR #1007, opened 2025-05-19
[F] Birdhouserunner - Reset option

UhWatchMe:
- Makes use of reset option
- Added sleepUntils for crafting since there are 2 exp drops


runelite-client/src/main/java/net/runelite/client/plugins/microbot/birdhouseruns/FornBirdhouseRunsScript.java
```diff
@@ -177,6 +183,23 @@ private boolean interactWithObject(int objectId) {
         return true;
     }
 
+    private void logMissingBankItems(InventorySetup setup) {
```
gmason0: I dont think this is really required, this here could just be modified if you want it to also include the quanity

https://github.com/chsami/Microbot/blob/development/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/Rs2InventorySetup.java#L120

runelite-client/src/main/java/net/runelite/client/plugins/microbot/birdhouseruns/FornBirdhouseRunsScript.java
```diff
@@ -186,19 +209,19 @@ private void seedHouse(WorldPoint worldPoint, states status) {
...
-        Rs2GameObject.interact(worldPoint, "Build");
-        sleepUntil(Rs2Player::isAnimating);
         botStatus = status;
     }
 
     private void dismantleBirdhouse(int itemId, states status) {
-        Rs2GameObject.interact(itemId, "Empty");
-        Rs2Player.waitForXpDrop(Skill.HUNTER);
+        Rs2GameObject.interact(itemId, "Reset");
+        if (!craftedBirdhouses) {
```
gmason0: this doesn't seem to be reset anywhere?, so this logic inside of `if(!craftedBirdHouses)` is only ran once?

gmason0 (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/birdhouseruns/FornBirdhouseRunsScript.java
UhWatchMe: Ah right, seems I still left it in. Basically this was for the interface when it wanted to make the birdhouses. Instead of doing that for each birdhouse spot, I had this variable that was set to true if they were all made.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/birdhouseruns/FornBirdhouseRunsScript.java
UhWatchMe: I agree, this was more of a personal debug since it kept failing after a deposit, I assume it's timing related to the inventorySetup class

runelite-client/src/main/java/net/runelite/client/plugins/microbot/birdhouseruns/FornBirdhouseRunsScript.java
```diff
@@ -186,19 +189,14 @@ private void seedHouse(WorldPoint worldPoint, states status) {
     }
 
     private void buildBirdhouse(WorldPoint worldPoint, states status) {
-        if (!Rs2Inventory.hasItem("bird house") && Rs2Inventory.hasItem(ItemID.CLOCKWORK)) {
```
Gaisberg, line 189: No more building birdhouses eh?

runelite-client/src/main/java/net/runelite/client/plugins/microbot/birdhouseruns/FornBirdhouseRunsScript.java
```diff
@@ -186,19 +189,14 @@ private void seedHouse(WorldPoint worldPoint, states status) {
...
-        }
-        Rs2GameObject.interact(worldPoint, "Build");
-        sleepUntil(Rs2Player::isAnimating);
         botStatus = status;
     }
 
     private void dismantleBirdhouse(int itemId, states status) {
-        Rs2GameObject.interact(itemId, "Empty");
-        Rs2Player.waitForXpDrop(Skill.HUNTER);
+        Rs2GameObject.interact(itemId, "Reset");
```
Gaisberg, line 196: This will work only for birdhouses that aren't finished so effectively this fix will most likely break existing logic of finished birdhouses. A better way to handle this would be to sleep until based on the state of the birdhouse

Gaisberg (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/birdhouseruns/FornBirdhouseRunsScript.java
UhWatchMe: Indeed, which is fine as long there is one. The Reset handles the crafting automatically. Making crafting, before the run, useless. That are my thoughts.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/birdhouseruns/FornBirdhouseRunsScript.java
UhWatchMe: Reset removed the birdhouse, and places the new one if you got the log. It uses the clockworks from the dismantled birdhouse right away. Therefor, gaining Crafting Experience.

https://oldschool.runescape.wiki/w/Bird_house_trapping#Catching_birds

Gaisberg: Nice! I didn't realise.

UhWatchMe: It was quite a sneaky update from Jagex's side. The only flaw the logic has, if there are no birdhouses placed, which would be a rare case of a player just starting, or messing it up legit. I need to update the requirements that I saw!

Edit ; the description seems fine

gmason0:
You need to pull in the latest changs from the development branch, resolve the merge conflict & push those changes so we can further review

gmason0:
any plans to fix the merge conflicts, or can this be closed @UhWatchMe ?

UhWatchMe:
Hi @g-mason0 sorry quite some hectic days. You can close this for now, I will do a new pull en merge :) but that will be next week.

PR #1008, opened 2025-05-19
Calcified rocks script + Various fixes 

SteffenCarlsen:
**Calcified rocks script**

- Mines deposits
- Option to smash deposits at an anvil or drop deposits/gem
- Option to focus on the water deposits when they spawn
- Option to hop if too many players around

**Bugfixes**

- Update Demonic Gorilla plugin description
- Added "Override start state" for Royal Titan script
- Added null-check on Tempoross script for crash prevention
- Make DeadFallHunter plugin schedulable
- Making Karambwan script not get stuck when transporting (Although I'm like 99% sure it should probably be removed at this point?)

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/gabplugs/karambwans/GabulhasKarambwansScript.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/gabplugs/karambwans/GabulhasKarambwansScript.java
@@ -7,6 +7,7 @@ import net.runelite.client.Notifier;
 import net.runelite.client.plugins.microbot.Microbot;
 import net.runelite.client.plugins.microbot.Script;
 import net.runelite.client.plugins.microbot.util.bank.Rs2Bank;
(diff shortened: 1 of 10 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/microbot/gabplugs/karambwans/GabulhasKarambwansScript.java
```diff
@@ -96,16 +97,13 @@ private void fishingLoop() {
 
     private void walkToRingToBank() {
         Rs2GameObject.interact(29495, "Zanaris");
+        Rs2Bank.walkToBank(BankLocation.ZANARIS);
```
gmason0, line 100: Just as an FYI - I do plan on nuking this plugin its kinda redundant now that AutoFishing can catch karambwans & is able to leverage the shortestpath plugin to find the nearest bank

gmason0 (approved):

PR #992, opened 2025-05-17
Motherload mine, Barrows, and Rev killer changes.

Gage307:
MLM will now avoid mining next to other players

Barrows The script will now wait until it's out of brews to tele out. Added Who is tunnel to the overlay.

Rev killer moved death listener

PR #1010, opened 2025-05-20
Revert "Motherload mine, Barrows, and Rev killer changes."

FunkyMonkeyCloud:
Reverts chsami/Microbot#992

No comments on changes made to plugins other than MLM. 

This MLM pull request has ruined the script for upper level mining. Worlds are too busy & there are too many players that use the upper level to include a players.findAny() check like this. My character frequently gets stuck in one position clicking the ground.

I added a log message to these MLM changes:

    Stream<Rs2PlayerModel> players = Rs2Player.getPlayers(it -> it!=null && it.getWorldLocation().distanceTo(wallObject.getWorldLocation()) <= 2);
    if (players.findAny().isPresent()) {
        Microbot.log("Script found another player too close - choosing another vein");
        return false;
    }

And now the subsequent screenshot

image


Gage307:
I see, sorry about that. I'll push a fix no need to revert

gmason0:
We will move forward with the changes on #1017 - if there are still issues doing this only down stairs, we will remove it.

PR #1019, opened 2025-05-20
Add volcanic ash mining script

SteffenCarlsen:
Ironman chore script that's schedulable

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/TaF/VolcanicAshMiner/VolcanicAshMinerScript.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/TaF/VolcanicAshMiner/VolcanicAshMinerScript.java
@@ -16,6 +16,7 @@ import net.runelite.client.plugins.microbot.util.math.Rs2Random;
 import net.runelite.client.plugins.microbot.util.player.Rs2Player;
 import net.runelite.client.plugins.microbot.util.security.Login;
 import net.runelite.client.plugins.microbot.util.walker.Rs2Walker;
+import net.runelite.client.plugins.mining.MiningAnimation;
 
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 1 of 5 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/microbot/TaF/VolcanicAshMiner/VolcanicAshMinerScript.java
```diff
@@ -0,0 +1,126 @@
...
+        if (maxPlayers > 0) {
+            WorldPoint localLocation = Rs2Player.getWorldLocation();
+
+            long nearbyPlayers = Microbot.getClient().getPlayers().stream()
+                    .filter(p -> p != null && p != Microbot.getClient().getLocalPlayer())
+                    .filter(p -> {
+                        return p.getWorldLocation().distanceTo(localLocation) <= 15;
+                    })
+                    //filter if players are using mining animation
+                    .filter(p -> p.getAnimation() != -1)
```
gmason0: There is a list of mining animations available here: https://github.com/chsami/Microbot/blob/main/runelite-client/src/main/java/net/runelite/client/plugins/mining/MiningAnimation.java

Not sure if we want to make this public & leverage this instead of just only checking if a player is animating at all?

gmason0 (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/TaF/VolcanicAshMiner/VolcanicAshMinerScript.java
SteffenCarlsen: Sounds like a good idea. Come to think of it, the animation check might catch nearby players doing crabs/birdhouse runs.

gmason0 (approved):

PR #1020, opened 2025-05-21
Adds "Defence Pure Mode" to AIO Fighter

FireStarter-Instigator:
1. When Monk Killer config is selected uses the same logic that's in the Monk Killer Plugin #944 
2. Disables banking and looting logic of aio fighter to stop "wrestling" of the bot, particularly if it starts in an area that isn't the monastery 
3. Disables NpcAttackScript

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/aiofighter/AIOFighterConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/aiofighter/AIOFighterConfig.java
@@ -987,6 +987,16 @@ public interface AIOFighterConfig extends Config {
         return new WorldPoint(0, 0, 0);
     }
 
+    @ConfigItem(
+            keyName = "monk_killer",
+            name = "Monk killer",
+            description = "Enable defence pure mode, goes to monestary and attacks monks",
+            position = 993
+    )
+    default boolean defencePureMode() {
+        return false;
+    }
+
 }
 
 
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/aiofighter/AIOFighterPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/aiofighter/AIOFighterPlugin.java
@@ -68,7 +68,7 @@ public class AIOFighterPlugin extends Plugin {
     public static int cooldown = 0;
     private final CannonScript cannonScript = new CannonScript();
     private final AttackNpcScript attackNpc = new AttackNpcScript();
-
+    private final MonkKillerScript defencePure = new MonkKillerScript();
     private final FoodScript foodScript = new FoodScript();
     private final LootScript lootScript = new LootScript();
     private final SafeSpot safeSpotScript = new SafeSpot();
@@ -148,6 +148,7 @@ public class AIOFighterPlugin extends Plugin {
         highAlchScript.run(config);
         potionManagerScript.run(config);
         safetyScript.run(config);
+        defencePure.run(config);
         //slayerScript.run(config);
         Microbot.getSpecialAttackConfigs()
                 .setSpecialAttack(true);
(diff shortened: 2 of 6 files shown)


gmason0:
This is exactly what I told you not to do. 

Please generically implement the features you'd like to see added, not a supplementary script that contains all of it in one. 

There are already features to attack npcs, skill etc. leverage those & add the missing features that aren't here like running away while breaking or world hoping.

FireStarter-Instigator:
> This is exactly what I told you not to do.
> 
> Please generically implement the features you'd like to see added, not a supplementary script that contains all of it in one.
> 
> There are already features to attack npcs, skill etc. leverage those & add the missing features that aren't here like running away while breaking or world hoping.

If I would be so bold as to ask why, as either this way or that way achieves the same end?

gmason0:
As this does not fit the definition of AIO fighter, which has a pretty structured code base to be generic as possible where what you've done is essentially stuck a hook into the side of it to keep your existing script there.

Implement the generic features you'd like to see such as running away before a break, world hopping, health ratio re-attack, all of those things can be generically implemented into this plugin

If you're unwilling to, sideloading is still an option

FireStarter-Instigator:
> Implement the generic features you'd like to see such as running away before a break, world hopping, health ratio re-attack, all of those things can be generically implemented into this plugin

1. running away before a break is something requested by the community (not me) which nobody can come to a conclusion about how to implement, it's not a generic feature i'd like to see, it's just a qol for defence pure training which I also use in moss killer, because as it's specific (and not generic) it allows for it. 

2. the particular features for defence pure training are unique. in no other situation do you want to re-attack something that has no health bar, there's no situation where that happens or where it would be useful except for being in combat for >30 mins with something without interacting with the game for all those 30 minutes, which only happens for an afk defence pure. **edit**: although arguably there may be some very particular situations like using guthans at bandit camp, once you've totally run out of potions and have afk'd now for 30 mins 

3. world hopping? that's just so you don't waste time idling waiting for an available monk that will most likely never appear, I honestly can't think where this would be relevant anywhere else. Anywhere else NPC's will come available within 30 seconds to a couple minutes because people kill npc's as fast as possible nowhere do people box npc's so they don't die like in defence pure training.

gmason0:
Then be the difference maker & implement it away that is generic enough to fit AIO fighter ! You don't have to wait around for someone to do it be the difference you'd like to see, within the guidelines of this plugin!

There are ways that I could think of within AttackNPCScript where you could create an option under skilling called "defense pure" that would allow for these specific re-attack mechanisms

Agreed World Hopping could be helpful in a few cases like if you are being crashed like  if you're idle for too long or if there are too many players near by, again by another configurable option that uses a timer or if you can find a way that could solidify this by real game data

FireStarter-Instigator:
and just to confirm what the suggestion is, to add these to the aio fighter config? world hopping config (if no npc's are available) attack when interacting but no health bar (if reachable) and the running away i guess you could put (try to run away before log)... still haven't come to conclusion about that....

could either use any available teleport (if banking is also selected as it may not be desired to go so far from the training spot) or worldpoint to the nearest singles spot if in multi, or blindly run in the opposite direction of the npc that is getinteracting with you 10 tiles at a time until you're not in combat, though that could be risky

FireStarter-Instigator:
> There are ways that I could think of within AttackNPCScript where you could create an option under skilling called "defense pure" that would allow for these specific re-attack mechanisms

it's just defence pures only train on monks as they heal themselves and have low defence 

others train on seagulls, active defence pure training, with pots and etc, like active stun alching compared to splashing but for defence pures, but that can be covered easily by aio fighter

just like aio magic has stun alch and alch mode and other modes, aio fighter should have defence pure mode (defence pures train at monks, that's the culture)

FireStarter-Instigator:
anyway ty for the feedback I respect your decisions and desires

PR #1038, opened 2025-05-25
Kebbithunter working

VIPDO1:
The old pull request messed up by me so this should be the new one and is working. Maybe it has some buggs that I could not find if anyone finds buggs please inform me so I will try to fix them

This pull request adds a new script for Kebbit hunting using the falconry method in the Piscatoris Hunter Area.

The script was written from scratch and handles all aspects of Kebbit hunting: identifying Kebbits, throwing the falcon, retrieving it, and responding to failed catches. It supports all Kebbit types based on the player’s Hunter level.

The plugin and overlay structure are based on the Salamander script by Bee & TaF, to maintain consistency with their existing style. However, all actual hunting logic — including falcon interactions and pathing — is entirely my own implementation.

The script includes settings to select the Kebbit type and configure sleep timings for realistic behavior. The overlay displays useful information like falcon status and hunter progress on screen.

To use this script, the player must go to the Piscatoris Falconry area and rent a falcon from the NPC Matthias.


runelite-client/src/main/java/net/runelite/client/plugins/microbot/hunterKabbits/HunterKabbitsScript.java
```diff
@@ -0,0 +1,249 @@
...
+                sleep(200, 300);
+            }
+
+            if (retrieved) {
+                sleep(config.minSleepAfterCatch(), config.maxSleepAfterCatch());
+                KebbitCaught++;
+                currentState = State.CATCHING;
+            }
+        } else {
+            NPC falcon = Rs2Npc.getNpc(GYR_FALCON_NPC_ID);
```
gmason0: You should be using Rs2NpcModel object type here.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/hunterKabbits/HunterKabbitsScript.java
```diff
@@ -0,0 +1,249 @@
...
+     *
+     * @param config The configuration for the script.
+     */
+    private void handleCatchingState(HunterKebbitsConfig config) {
+        String npcName = getKebbit(config).getNpcName();
+
+        if (Rs2Npc.interact(npcName, "Catch")) {
+            boolean falconActive = false;
+            for (int i = 0; i < 10; i++) {
+                NPC falcon = Rs2Npc.getNpc(1342);
```
gmason0: Magic Number & incorrect object type

runelite-client/src/main/java/net/runelite/client/plugins/microbot/hunterKabbits/HunterKabbitsScript.java
```diff
@@ -0,0 +1,249 @@
...
+     */
+    private void handleRetrievingState(HunterKebbitsConfig config) {
+        final int GYR_FALCON_NPC_ID = 1342;
+        NPC hintNpc = Microbot.getClient().getHintArrowNpc();
+
+        if (hintNpc != null && hintNpc.getId() == GYR_FALCON_NPC_ID) {
+            Rs2NpcModel model = new Rs2NpcModel(hintNpc);
+            boolean retrieved = false;
+            for (int i = 0; i < 5; i++) {
+                if (Rs2Npc.interact(model, "Retrieve")) {
```
gmason0, line 128: there should be a check if the thread is still running when looping inside of the scheduledFuture to ensure that inner-loops are broken when the future is cancelled

```
for (int i = 0; i < 5; i++) 
 If(!isRunning()) break;
```

runelite-client/src/main/java/net/runelite/client/plugins/microbot/hunterKabbits/HunterKabbitsScript.java
```diff
@@ -0,0 +1,249 @@
...
+    private void handleDroppingState(HunterKebbitsConfig config) {
+        KebbitHunting currentKebbit = getKebbit(config);
+        Integer furItemId = getSupportedFurItemId(currentKebbit);
+        if (furItemId == null) {
+            currentState = State.CATCHING;
+            return;
+        }
+
+        while (Rs2Inventory.contains(furItemId) || Rs2Inventory.contains(ItemID.BONES)) {
+            if (Rs2Inventory.contains(furItemId)) Rs2Inventory.drop(furItemId);
```
gmason0, line 102: Similar thing here, add a condition at the top of this loop to break out of the inner loop if the future is cancelled

gmason0 (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/hunterKabbits/HunterKabbitsScript.java
VIPDO1: so would it fix the problem if i would initialize 1342 at the top of my code?
because i found out if you catch a diffrent kebbit then the id changes 
private static final Set<Integer> VALID_FALCON_IDS = Set.of(1342, 1343, 1344, 1345, 1346);
would this fix the magic number problem?

VIPDO1: > Magic Number & incorrect object type

so would it fix the problem if i would initialize 1342 at the top of my code?
because i found out if you catch a diffrent kebbit then the id changes
private static final Set VALID_FALCON_IDS = Set.of(1342, 1343, 1344, 1345, 1346);
would this fix the magic number problem?

gmason0: A magic number is a constant that is used within your code without an explanation, you should avoid using the id directly here & define constants as class members instead

so for example,

```
// Outside of the run method

private final int WHATEVER_THIS_NUMBER_IS = 1342

// Then, where-ever you're using this number
Rs2NpcModel falcon = Rs2Npc.getNpc(WHATEVER_THIS_NUMBER_IS);
```

gmason0: Yes, this would be fine. 

you could leverage the predicate based method to find the NPC using this Set of IDs

runelite-client/src/main/java/net/runelite/client/plugins/microbot/hunterKabbits/HunterKabbitsScript.java
VIPDO1: > there should be a check if the thread is still running when looping inside of the scheduledFuture to ensure that inner-loops are broken when the future is cancelled
> 
> ```
> for (int i = 0; i < 5; i++) 
>  If(!isRunning()) break;
> ```

Can you say it a little bit in noob language so I can understand what you are saying? Should I add break; in my loops?

gmason0: Sure! So, let's just start at the beginning. Inside of Microbot the 'script' structure uses 2 primary pieces. A plugin class & a script class. There are other classes such as overlay panels, general overlays/highlighting objects a config interface, but won't get too much in-detail about those sorts of things.

Your Plugin class extends the abstract class `Plugin`, which contains baked-in methods & class members that can be used in any concrete class that extends from the abstract class. Microbot introduces an abstract class called `Script`. Inside of this script class, there is a couple extremely important things.

```java
// Scheduled Executor Service that creates a pool of threads to scheduled jobs on, in this case, 10 threads are available to be scheduled.
protected ScheduledExecutorService scheduledExecutorService = Executors.newScheduledThreadPool(10);

// Scheduled Futures, these are jobs that can be ran on the Scheduled Executor Service
protected ScheduledFuture<?> scheduledFuture;
protected ScheduledFuture<?> mainScheduledFuture;
```

Moving onto the concrete classes that you would create, in this case, `HunterKabbitsScript.java`, this class extends `Script` so you inherit all methods & class members from the abstract class. You create an implementation of the run() method & the shutdown() method, inside of the run() method, all scripts start with `mainScheduledFuture = scheduledExecutorService.scheduleWithFixedDelay()` so here we are saying, keep the result of computation of the `Runnable` that's being ran on Executor Service inside of this variable. scheduleWithFixedDelay is an infinite loop until the one of the following things happen.

- The task is explicitly cancelled via the returned future
- The executor terminates, also resulting in task cancellation.
- An execution of the task throws an exception. In this case calling get on the returned future will throw ExecutionException, holding the exception as its cause.

Now we have some base understanding, moving on the issue at hand.

When we call the shutdown method, we effectively are canceling mainScheduledFuture, which will cause the the infinite loop to stop once it has reached the end of its current iteration, BUT when we have these inner loops, these have to be completed before continuing on to the rest of the mainScheduledFuture's runnable task. For example, lets take something super simple like: 

```java
for (int i = 0; i < 20; i++) {
	System.out.println("Running " + i);
	sleep(100);
}
```

If we quickly start & turn off the plugin/script, we will see the following

```text
Running 0
Running 1
Running 2
Running 3
...
Running 18
Running 19
Total time for loop 1957
```
This shows that this inner loop needs to complete BEFORE we can actually shutdown the script. But by adding `if (!isRunning()) break; at the top of your inner loops, we will see something like this

```text
Running 0
Running 1
Running 2
Total time for loop 257
```
Essentially, here we are saying if the thread is cancelled or isDone, then we are breaking out of this inner loop, as there as the desired thread state is to be cancelled or done.

There is so much more to concurrency & the scheduledExecutorService that I haven't explained here, but hopefully this sheds light onto why adding this if statement to the top of your inner-loops is important.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/hunterKabbits/HunterKabbitsScript.java
VIPDO1: > Similar thing here, add a condition at the top of this loop to break out of the inner loop if the future is cancelled

So I did the changes you asked for I hope I did not oversee anything. I also added the function that you can decide if you want to bury the bones or not. I was asked for that in my old Pull request

VIPDO1: They are committed in my fork on my main

PR #1040, opened 2025-05-25
extend sleepUntil time limit from 5000ms to 8000ms

crispycritter99:
adjustment to account for quetzal whistle, which has an additional animation that otherwise causes webwalker to fail with the 5000ms time limit


gmason0 (changes requested):
Shouldn't we be raising this value in `handleInventoryTeleports` & `handleWearableTeleports` ?

crispycritter99:
yup it should really be in handleInventoryTeleports, i edited the wrong line for the PR

gmason0 (approved):

PR #1041, opened 2025-05-26
Motherload Mining: Specific mining location added with dropdown config

DylanMartino:
Added Mining location selection in config to mine at a specific location rather than random, with the choice of (ANY) spots if the user chooses to select.

Updated mining location coordinates as well.


.run/Runelite.run.xml
```diff
@@ -1,13 +1,13 @@
 <component name="ProjectRunConfigurationManager">
-    <configuration default="false" name="Runelite" type="Application" factoryName="Application" singleton="false">
```
gmason0, line 2: please check out an older version of this file, we should not make any changes to the run config that is in the repo.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/mining/motherloadmine/MotherloadMineScript.java
```diff
@@ -337,31 +338,74 @@ private void bankItems()
...
-    private void selectRandomMiningSpot()
-    {
-        // Randomly decide which spot to go to
-        // More variety can be added if needed
-        miningSpot = (Rs2Random.between(1, 5) == 2)
-                ? (config.mineUpstairs() ? MLMMiningSpot.WEST_UPPER : MLMMiningSpot.SOUTH)
-                : (config.mineUpstairs() ? MLMMiningSpot.EAST_UPPER : MLMMiningSpot.WEST_LOWER);
-        Collections.shuffle(miningSpot.getWorldPoint());
+//    private void selectRandomMiningSpot()
+//    {
```
gmason0, line 342: Can we just remove this method if we are no longer going to use it? The method can still be found in the commit history if we did need to revert

runelite-client/src/main/java/net/runelite/client/plugins/microbot/mining/motherloadmine/MotherloadMineScript.java
```diff
@@ -337,31 +338,74 @@ private void bankItems()
...
+            // 🔁 Random selection based on upstairs config
+            if (config.mineUpstairs()) {
+                miningSpot = Rs2Random.between(0, 1) == 0 ? MLMMiningSpot.WEST_UPPER : MLMMiningSpot.EAST_UPPER;
+            }
+            else {
+                miningSpot = Rs2Random.between(0, 1) == 0 ? MLMMiningSpot.WEST_LOWER : MLMMiningSpot.WEST_MID;
+                miningSpot = Rs2Random.between(0, 1) == 0 ? MLMMiningSpot.SOUTH_WEST : MLMMiningSpot.SOUTH_EAST;
+            }
+        } else {
+            // ✅ Use the user-specified mining area
```
gmason0, line 364: Remove ChatGPT emojis

runelite-client/src/main/java/net/runelite/client/plugins/microbot/mining/motherloadmine/MotherloadMineScript.java
```diff
@@ -337,31 +338,74 @@ private void bankItems()
...
     }
 
     private boolean walkToMiningSpot()
     {
         WorldPoint target = miningSpot.getWorldPoint().get(0);
-        if (config.mineUpstairs() && !isUpperFloor())
-        {
+
+        // Navigates to correct floor based on selected mining area
+        if ((miningSpot == MLMMiningSpot.EAST_UPPER || miningSpot == MLMMiningSpot.WEST_UPPER) && !isUpperFloor()) {
```
gmason0, line 392: I think something that would be beneficial here would be to add a parameter to the MLMMiningSpot enum called

`isUpstairs`, when doing so, you'd just be able to check 

`miningSpot.isUpstairs() && !isUpperFloor()`

Doing something like this would remove the complexity from these conditions

gmason0 (changes requested):

FunkyMonkeyCloud:
FYI I have some fulsome commits in the works for the MLM script. I haven't touched mining areas, but my request would be to keep an option for random selection. With upper level mining, keeping it random feels more human like, rather than sitting in front of the same 4 veins for hours

DylanMartino:
I kept an option to mine at random spots, using the option ANY in the
drop down which it is defaulted to 👍

PR #1042, opened 2025-05-26
InvokeMenu check if bank is open

Sorries:


FireStarter-Instigator:
yeah isOpen keeps returning true for me even when nowhere near a bank, i've had to just used modified entries of openBank and walkToBank that don't have an isOpen check

Sorries:
my issue was that Rs2Inventory.invokeMenu would call Rs2Tab.switchToInventoryTab();. It would close out my bank as my tab was on magic.

PR #1048, opened 2025-05-28
Refactor Rs2Equipment

Kryaex:


subinznz:
what is this supposed to improve / do

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/equipment/Rs2Equipment.java
```diff
@@ -1,80 +1,69 @@
...
+            Item item = itemContainer.getItems()[i];
+            if (item.getId() == -1) continue;
+            ItemComposition itemComposition = Microbot.getClient().getItemDefinition(item.getId());
+            _equipmentItems.add(new Rs2ItemModel(item, itemComposition, i));
         }
+        equipmentItems = _equipmentItems;
     }
 
+    @Deprecated(since = "Use interact", forRemoval = true)
     public static boolean useCapeAction(int itemId, String action) {
```
gmason0, line 53: As a side note, I feel this method should just be nuked, it was used early on in the walk then we ended up just using the Rs2Equipment.invokeMenu directly.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/equipment/Rs2Equipment.java
```diff
@@ -1,80 +1,69 @@
...
             return false;
         }
 
-        Microbot.doInvoke(new NewMenuEntry(-1, 25362448, MenuAction.CC_OP.getId(), identifier, -1, action),
-                new Rectangle(1, 1, Microbot.getClient().getCanvasWidth(), Microbot.getClient().getCanvasHeight()));
+        invokeMenu(item, action);
         return true;
     }
 
     public static boolean useRingAction(JewelleryLocationEnum jewelleryLocationEnum) {
```
gmason0, line 65: Same thing with this, its really no longer needed

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/equipment/Rs2Equipment.java
```diff
@@ -84,7 +73,7 @@ public static boolean useRingAction(JewelleryLocationEnum jewelleryLocationEnum)
     }
 
     public static boolean useAmuletAction(JewelleryLocationEnum jewelleryLocationEnum) {
```
gmason0, line 76: Same here

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/equipment/Rs2Equipment.java
```diff
@@ -93,25 +82,17 @@ public static boolean useAmuletAction(JewelleryLocationEnum jewelleryLocationEnu
         return true;
     }
 
-
     public static Rs2ItemModel get(EquipmentInventorySlot slot) {
-        return equipmentItems.stream().filter(x -> x.slot == slot.getSlotIdx()).findFirst().orElse(null);
+        return get(x -> x.slot == slot.getSlotIdx());
```
gmason0: I would like to recommend using the getters from the Rs2ItemModel class instead of using the class members directly. Ideally these values should be private so they can not be modified

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/equipment/Rs2Equipment.java
```diff
@@ -93,25 +82,17 @@ public static boolean useAmuletAction(JewelleryLocationEnum jewelleryLocationEnu
...
-        if (exact) {
-            return equipmentItems.stream().filter(x -> x.name.equalsIgnoreCase(name))
-                    .findFirst()
-                    .orElse(null);
-        }
-        return equipmentItems.stream().filter(x -> x.name.toLowerCase().contains(name.toLowerCase()))
-                .findFirst()
-                .orElse(null);
-
+        return exact ? get(x -> x.name.equalsIgnoreCase(name)) :
```
gmason0: Can we get the ternary expression inside of the get method so

`return get (x -> exact ? x.getName().equalsIgnoreCase(name) : x.getName().toLowerCase().contains(name.toLowerCase())`

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/equipment/Rs2Equipment.java
```diff
@@ -219,35 +161,36 @@ public static boolean isEquipped(String name, EquipmentInventorySlot slot) {
...
 
     public static boolean isEquipped(String name, EquipmentInventorySlot slot, boolean exact) {
         final Rs2ItemModel item = get(slot);
-        if (exact) {
-            return item != null && item.name.equalsIgnoreCase(name);
-        } else {
-            return item != null && item.name.toLowerCase().contains(name.toLowerCase());
-        }
+        if (item == null) return false;
+        return exact ? item.name.equalsIgnoreCase(name) : item.name.toLowerCase().contains(name.toLowerCase());
```
gmason0: Same as before, I would prefer us to use the getters from Rs2ItemModel & moving the ternary expression

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/equipment/Rs2Equipment.java
```diff
@@ -1,80 +1,69 @@
...
-                int finalI = i;
-                Optional<EquipmentInventorySlot> equipmentSlot = Arrays.stream(EquipmentInventorySlot.values()).filter(x -> x.getSlotIdx() == finalI).findFirst();
-                if (equipmentSlot.isEmpty()) continue;
-                int slot = equipmentSlot.get().getSlotIdx();
-                ItemComposition itemComposition = Microbot.getClientThread().runOnClientThreadOptional(() ->
-                        Microbot.getClient().getItemDefinition(item.getId())).orElse(null);
-                _equipmentItems.add(new Rs2ItemModel(item, itemComposition, slot));
-            }
-            equipmentItems = _equipmentItems;
+        assert Microbot.getClient().isClientThread();
```
gmason0, line 36: I am not sure how I feel about using assertions here. Us developers have assertions enabled when running the client from the IDE, but this is not something that is enabled by default when running the jar from the launcher

could we just introduce a early return such as - or possibly @chsami what are your thoughts on this?

```
if (!Microbot.getClient().isClientThread()) {
    Microbot.log("This must be ran from the client thread!", Level.WARN);
    return;
}

File: runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/equipment/Rs2Equipment.java [hunk 8]
```diff
@@ -1,80 +1,69 @@
...
 public class Rs2Equipment {
     public static ItemContainer equipment() {
-        return Microbot.getClient().getItemContainer(InventoryID.EQUIPMENT);
+        return Microbot.getClient().getItemContainer(InventoryID.WORN);
     }
 
-    public static List<Rs2ItemModel> equipmentItems = new ArrayList<>();
-    
+    // TODO: make private? why does the vorkath script need access to this?
+    public static List<Rs2ItemModel> equipmentItems = Collections.emptyList();
```
gmason0: agreed, this should probably be private & just use the items() method

gmason0 (changes requested):
I like the looks of this so far, but there are a few syntax nit-picks that I would like to see adjusted & I am not entirely sure how I feel about using assertions in this method

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/equipment/Rs2Equipment.java
chsami: Just use early returns

gmason0 (approved):

PR #1052, opened 2025-05-31
Wilderness Script goes live?

oversized93:
Think this is ready to get sent out to the world.  Let me know if there's anything in here you'd want removed before pushing.


FireStarter-Instigator:
gl cranny, ik a lot of vennies gonna be pleased with this

PR #1057, opened 2025-06-04
Refactor Rs2Inventory

Kryaex:


gmason0:
@Kryaex - Is this ready to merge? If so, can you resolve the merge conflicts & take this out of draft?

gmason0 (approved):
There are some breaking changes here, as far as changing the data-types that some of the API methods take, but I see that it does bring down some of the duplicate functionality, I would just ask for us to try to stray away from this in the future.

I am not sure what differences it makes using primitive types vs reference types in this sort of scenario. But I appreciate the standardization towards primitives here.

PR #1064, opened 2025-06-05
Fixed issue where GEcooker is stuck at Cooldown Active

NLG-code:
Fixed issue where its stuck at Cooldown Active after bank and 1st inventory cooked.
Removing the antiban check fixed the issue


NLG-code:
for some reason original works on my other pc will close this for now

PR #1065, opened 2025-06-06
Fixed Prayer Helper for defence requirements

FunkyMonkeyCloud:
Added a character defence check for the prayers Piety & Chivalry

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/zerozero/zeroprayer/AttackTimerMetronomePlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/zerozero/zeroprayer/AttackTimerMetronomePlugin.java
@@ -653,11 +653,13 @@ public class AttackTimerMetronomePlugin extends Plugin
             case CONTROLLED:
             case DEFENSIVE:
                 if (Microbot.getVarbitValue(CAMELOT_TRAINING_ROOM_STATUS) == 8 &&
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/microbot/zerozero/zeroprayer/AttackTimerMetronomePlugin.java
```diff
@@ -653,11 +653,13 @@ private Rs2PrayerEnum determineOffensivePrayer(AttackStyle attackStyle) {
             case CONTROLLED:
             case DEFENSIVE:
                 if (Microbot.getVarbitValue(CAMELOT_TRAINING_ROOM_STATUS) == 8 &&
```
gmason0: I think the better thing to do here is add these requirements to `Rs2Prayer.getBestMeleePrayer()` & use that here

gmason0 (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/zerozero/zeroprayer/AttackTimerMetronomePlugin.java
gmason0: Oh they are already included! lets just use that

FunkyMonkeyCloud: Ahhh good spotting. I'll put the fix in now

gmason0 (approved):

PR #1069, opened 2025-06-07
fixed walkTo distance so doesn't get stuck at Trollheim patch

subinznz:
…added  a retry on inventory setup on failure when banking with high ping


runelite-client/src/main/java/net/runelite/client/plugins/microbot/herbrun/HerbrunScript.java
```diff
@@ -65,10 +65,15 @@ public boolean run() {
                 }
                 var inventorySetup = new Rs2InventorySetup(config.inventorySetup(), mainScheduledFuture);
                 if (!inventorySetup.doesInventoryMatch() || !inventorySetup.doesEquipmentMatch()) {
+                    Microbot.log("Starting Inventory Setup");
                     Rs2Walker.walkTo(Rs2Bank.getNearestBank().getWorldPoint(), 20);
-                    if (!inventorySetup.loadEquipment() || !inventorySetup.loadInventory()) {                        
-                        plugin.reportFinished("Failed to load inventory setup",false);
-                        return;
+                    if (!inventorySetup.loadEquipment() || !inventorySetup.loadInventory()) {
```
gmason0, line 72: This is just overall a bad idea, please let me know what specific issues you're having with inventory setups so they can be diagnosed

gmason0 (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/herbrun/HerbrunScript.java
subinznz: I'll remove this from this pull request, you happy with the other changes ?

gmason0: Thats fine lets remove that

runelite-client/src/main/java/net/runelite/client/plugins/microbot/herbrun/HerbrunScript.java
```diff
@@ -255,7 +257,9 @@ private static String getHerbPatchState(TileObject rs2TileObject) {
 
     @Override
     public void shutdown() {
-        super.shutdown();
-        initialized = false;
+        if (isRunning()) {
```
gmason0, line 262: We also dont need isRunning() wrapped here

runelite-client/src/main/java/net/runelite/client/plugins/microbot/herbrun/HerbrunPlugin.java
```diff
@@ -60,7 +61,11 @@ protected void shutDown() {
     @Subscribe
     public void onPluginScheduleEntrySoftStopEvent(PluginScheduleEntrySoftStopEvent event) {
         if (event.getPlugin() == this) {
-            Microbot.stopPlugin(this);
+            if (herbrunScript != null && herbrunScript.isRunning()) {
```
gmason0, line 64: `Microbot.stopPlugin(this)` is effectively going to call the shutdown() method

gmason0 (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/herbrun/HerbrunScript.java
subinznz: Was a recommendation by Vox since stopPlugin did not actually stop and wait for the plugin to finish action on soft stop event.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/herbrun/HerbrunPlugin.java
subinznz: Same as above.  Happy to revert if this won’t cause the switching issues

runelite-client/src/main/java/net/runelite/client/plugins/microbot/herbrun/HerbrunScript.java
gmason0: Let me think on it & i'll get back to you

Voxsylvae: The scheduler plugin will  actually wait until the plugin stop's.  Just when the onPluginScheduleEntrySoftStopEvent is send to the herbrunner and we stop the script there we should check in the stop plugin method if the script was already stopped. Because the shutdown of the plugin will be invoked later and the script could just be stopped earlier. Hope these make sense

gmason0: @Voxsylvae - Why are we delaying the plugin stopping using invoke later though? Wouldn't we want to just stop the plugin whenever a `onPluginScheduleEntrySoftStopEvent` is sent? Then once the plugin's shutdown method is called, it shuts down the script? I think we should keep considering a plugin & the script as one, not shutting them down seperately

gmason0: @Voxsylvae - Still waiting for a response here, just curious as to why we are using invokeLater to stop the plugin? Shouldn't we be relying just on Microbot.stopPlugin() to stop it?

Voxsylvae: In the current state of the HerbRun plugin.these both should not be necessary. 
When you want to shutdown the script in onPluginScheduleEntrySoftStopEvent is if you have some scheduled future, In which you perform post schedule task/clean ups. Regarding when stoping with microbot.stopplugin not on the client thread you should Invoke the stop on the client thread. 
I hope it makes sense, in the current state of the herbrun Plugin both should not be necessary. 
Please reach out to me in discord, if something should be discussed even further. Sorry for the late reply I missed the notification..

gmason0: This is my thought process as well, that we do not need delay the stopPlugin using invokeLater..

@subinznz please clean up accordingly.

subinznz: i'm going to close this, will do another pull request

PR #1073, opened 2025-06-07
Just look at MahoganyHomesScript.java 

VIPDO1:
Added support for bank teleport tabs:
The script now uses teleport tabs (Falador, Ardougne, Varrock) to get to the bank faster if available.

Xeric's Talisman fallback:
If no tab is available, the script uses Xeric's Talisman and automatically picks the best available destination (priority order).

Anti-ban settings applied:
applyAntiBanSettings() is called at script startup to enable all anti-ban features.

I dont know how to take the rest out I commited it on my main but I cant take it down I dont know how


runelite-client/src/main/java/net/runelite/client/plugins/microbot/mahoganyhomez/MahoganyHomesScript.java
```diff
@@ -351,9 +359,15 @@ private void bank() {
                 && isMissingItems()) {
             ShortestPathPlugin.getPathfinderConfig().setIgnoreTeleportAndItems(true);
             BankLocation bankLocation = Rs2Bank.getNearestBank(currentHome.getLocation());
+
+            if (useTeleportTabForBank(bankLocation)) {
```
gmason0, line 363: I am confused here, is the issue not related to the lines above this 
```
            ShortestPathPlugin.getPathfinderConfig().setIgnoreTeleportAndItems(true);
            BankLocation bankLocation = Rs2Bank.getNearestBank(currentHome.getLocation());
```

gmason0 (changes requested):

.run/Runelite.run.xml
```diff
@@ -1,13 +1,13 @@
...
-        <option name="MAIN_CLASS_NAME" value="net.runelite.client.RuneLite"/>
-        <module name="client"/>
-        <option name="PROGRAM_PARAMETERS" value="-developer-mode -clean-jagex-launcher"/>
-        <option name="VM_PARAMETERS" value="-ea"/>
-        <method v="2">
-            <option name="Make" enabled="true"/>
-        </method>
-    </configuration>
-</component>
+  <configuration default="false" name="Runelite" type="Application" factoryName="Application" singleton="false">
```
gmason0, line 2: please checkout an older version of this file, we should not see any modficiations here

runelite-client/src/main/java/net/runelite/client/plugins/microbot/mahoganyhomez/MahoganyHomesScript.java
SteffenCarlsen, line 1: I am not sure what the issue is with the current script? It uses the webwalker, so doesn't care about how you decide to change-up your inventory with runes/tablets

VIPDO1: So I was running this script but it was always walking to the houses although I had the teleport tabs to the Cities so now I added when the script locates the nearest bank it teleports to the city so that I dont have to walk that much


> I am not sure what the issue is with the current script? It uses the webwalker, so doesn't care about how you decide to change-up your inventory with runes/tablets

gmason0:
Do we plan on implementing any of the requests? I am also still confused about this teleport thing, we want to include teleports in the nearest bank to current house location, I think we should just add a toggle for the line I mentioned 

` ShortestPathPlugin.getPathfinderConfig().setIgnoreTeleportAndItems(true);`

PR #1077, opened 2025-06-09
Lunar tablets dream mentor check added

Gage307:
We'll no longer spam click the bank that's unavailable to people who haven't complete Dream Metor


gmason0:
We do not want a random booth, we want to ensure we are always interacting with the correct booth.

I have created a fix for this here: https://github.com/chsami/Microbot/commit/40ae40adf6de1b7cf3b29329cb3a99095d81daa3

Now any script that eventually finds its way to the openBank/findBank method it will consider which booths are usable

Gage307:
> We do not want a random booth, we want to ensure we are always interacting with the correct booth.
> 
> I have created a fix for this here: [40ae40a](https://github.com/chsami/Microbot/commit/40ae40adf6de1b7cf3b29329cb3a99095d81daa3)
> 
> Now any script that eventually finds its way to the openBank/findBank method it will consider which booths are usable

Just to clarify it's random of two booths that will work for someone who doesn't have Dream Mentor because it's tied to the booth's worldpoints

Oooo! Thank you for the fix appreciate that!

PR #1100, opened 2025-06-12
fixed farm tree run to start without spirit mode turned on

subinznz:
forgot to remove this spirit tree check


gmason0:
I have pushed the change to include your fix here, also just completely removing spirit mode. If you would like to rely on spirit tree for transportation, ideally we should work on two things here.

1. Finding the proper varbits/varplayers for player grown spirit trees to be grown.
2. Re-working this entire plugin to find the distance between each of the enabled patches to find the most efficient patch order, the way this is implemented now forces users to make modifications to the code base.

PR #1111, opened 2025-06-16
Fixed issues with Vyrewatchkiller plugin

NLG-code:
1. Added automatic banking when inventory is full or close to full i.e. when left with 1 inventory slot
2. Fixed issue where it paths to statue wierdly and clicks around the statue before interacting with statue
3. Fixed issue where it stops at the statue for prolonged period before going back to initial position
4. Changed looting system to allow telegrab so character doesnt keep moving around also fixes the issue where vampyre dust is looted by the old looting system if item value is low i.e. <5k

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/virewatch/PAlcher.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/virewatch/PAlcher.java
@@ -16,7 +16,8 @@ public class PAlcher extends Script {
             "Adamant platelegs",
             "Adamant platebody",
             "Rune full helm",
-            "Rune kiteshield"
+            "Rune kiteshield",
+            "Rune 2h sword"
     };
 
     public boolean run(PVirewatchKillerConfig config, PVirewatchKillerPlugin plugin) {
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/virewatch/PAlcher.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/virewatch/PAlcher.java
@@ -1,45 +1,40 @@
 package net.runelite.client.plugins.microbot.virewatch;
-
 import net.runelite.client.plugins.microbot.Microbot;
 import net.runelite.client.plugins.microbot.Script;
 import net.runelite.client.plugins.microbot.util.inventory.Rs2Inventory;
-import net.runelite.client.plugins.microbot.util.keyboard.Rs2Keyboard;
 import net.runelite.client.plugins.microbot.util.magic.Rs2Magic;
 import net.runelite.client.plugins.microbot.util.tabs.Rs2Tab;
-import net.runelite.client.plugins.microbot.util.widget.Rs2Widget;
-
-import java.awt.event.KeyEvent;
+import net.runelite.client.plugins.skillcalculator.skills.MagicAction;
 import java.util.concurrent.TimeUnit;
 
 public class PAlcher extends Script {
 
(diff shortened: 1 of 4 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/microbot/virewatch/PAlcher.java
```diff
@@ -51,25 +47,16 @@ public boolean run(PVirewatchKillerConfig config, PVirewatchKillerPlugin plugin)
     private void alchItem(String item) {
         plugin.alchedItems++;
         plugin.alchingDrop = true;
-        Rs2Magic.alch(item);
+        if (Rs2Magic.canCast(MagicAction.HIGH_LEVEL_ALCHEMY)){
```
gmason0, line 50: Please review the implementation of Rs2Magic.alch, it will use High Alch or Low Alch depending on magic level. We should be considering that here as well.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/virewatch/PLooter.java
```diff
@@ -1,64 +1,102 @@
...
-            if(config.lootRunes()) lootItemByName(config, "rune");
-            if(config.lootCoins()) lootItemByName(config, "coins");
+            try{
+                if (!super.run()) return;
 
+                if (Rs2Inventory.isFull() || Rs2Inventory.getEmptySlots() <= 1) {
+                    bank(config);
+                    return;
+                }
+                System.out.println("toggleLootItems: " + config.toggleLootItems());
```
gmason0, line 32: Usage of System.out.printLn, should use Microbot.log or SL4J logging.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/virewatch/PLooter.java
```diff
@@ -1,64 +1,102 @@
...
+        Rs2Bank.depositAll("Dragonstone bolt tips");
+
+        Rs2Inventory.waitForInventoryChanges(1800);
+        Rs2Bank.closeBank();
+        sleepUntil(() -> !Rs2Bank.isOpen());
+
+        Rs2Walker.walkTo(currentLocation);
+
+        // Re-equip combat gear from config
+        Rs2Inventory.equip(config.top());
```
gmason0, line 73: We should be storing the equipment in memory instead of using configuration options to define gear. Requires less of the user.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/virewatch/PLooter.java
```diff
@@ -1,64 +1,102 @@
...
+                .collect(Collectors.toList());
+
+        if (config.teleGrabLoot()) {
+            for (var item : itemsToPickup) {
+                int distance = Microbot.getClient().getLocalPlayer().getWorldLocation().distanceTo(item.getTile().getWorldLocation());
+                if (distance > 2) {
+                    Rs2Magic.cast(MagicAction.TELEKINETIC_GRAB);
+                }
+                Rs2GroundItem.interact(item);
+                sleepUntil(() -> Rs2Inventory.waitForInventoryChanges(1800));
```
gmason0, line 91: Improper usage of Rs2Inventory.waitForInventoryChanges(). waitForInventoryChanges is a sleepUntil the size of the inventory changes or the size of a stackable item in the inventory changes.

gmason0 (changes requested):

PR #1116, opened 2025-06-17
64smrr codex/create ge flipper with overlay and antiban

KSPOG:


runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/grandexchange/Rs2GrandExchange.java
```diff
@@ -214,87 +212,6 @@ private static void setPrice(int price) {
 
     }
 
-    public static boolean buyItemAbove5Percent(String itemName, int quantity) {
-        return buyItemAbove5Percent(itemName, quantity, 1);
```
gmason0, line 218: Why are removing a large amount of methods from the Rs2GrandExchange class?

gmason0 (changes requested):
We should also look to clean up the commit history, adding a new plugin shouldn't require > 100 commits.

gmason0:
I am going to close this, lets clean up the commit history & the Rs2GrandExchange class methods before re-opening..

PR #1125, opened 2025-06-23
mosskillerplugin + wildyscript updates

FireStarter-Instigator:
- adds new process walk (boolean supplier) to rs2walker
- cleans up some gpt dross in plugin class
- updates tick reader for pvp target
- changes plugin class calls for local calls for target
- drinks the energy potion withdrawn
- destroys or drops untradeables to not get into a banking loop
- withdraws nature runes if alching is selected

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/bee/MossKiller/MossKillerPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/bee/MossKiller/MossKillerPlugin.java
@@ -571,7 +571,7 @@ public class MossKillerPlugin extends Plugin implements SchedulablePlugin {
 
             // Increment tick count if the player is interacting and performing combat animation
             if (player.getInteracting() == localPlayer && !isNonCombatAnimation(player)) {
-                tickCount += 3;
+                tickCount += 2;
                 attackerTickMap.put(player, tickCount);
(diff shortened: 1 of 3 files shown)


gmason0 (changes requested):
Let's remove this new walkTo method, this sort of thing is already possible by setting  target to null based on if a condition is met, otherwise the other non-walker changes look fine to me

FireStarter-Instigator:
Rs2Walker.walkTo(MOSS_GIANT_SPOT, 6, () -> mossKillerPlugin.currentTarget != null);

gmason0:
I see the changes you made, but I don't feel this is required or necessary. The walker is already interruptible by setting the target to null. 

Please make the proper modifications to revert the changes that were made to the Rs2Walker class.

FireStarter-Instigator:
> I see the changes you made, but I don't feel this is required or necessary. The walker is already interruptible by setting the target to null.
> 
> Please make the proper modifications to revert the changes that were made to the Rs2Walker class.

I tried every other alternative, doesn't work for this use case, not without some serious over-engineering and back bending. Just leave the plugin as it is, I won't be reverting the changes. Just thought I'd post my updates for the lads who have been using and enjoying the plugin.

gmason0:
Sounds good! - I will get this closed then considering we are at a standstill.

PR #1132, opened 2025-06-25
[F] Birdhouse - Use Reset button to build birdhouses

UhWatchMe:
Make use of the "Reset" option of the birdhouses.


UhWatchMe:
Oops I forgot to edit buildbirdhouses function, will be adjusting that and testing. Since we make use of the Reset it is not needed.

UhWatchMe:
Alright that is it.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/birdhouseruns/FornBirdhouseRunsScript.java
```diff
@@ -204,19 +205,14 @@ private void seedHouse(WorldPoint worldPoint, states status) {
     }
 
     private void buildBirdhouse(WorldPoint worldPoint, states status) {
```
chsami, line 207: is this method still necessary? it seems to only set the botstatus

UhWatchMe: Necessary, not really but niec to keep it in there if in some cases there needs to be an update regarding it, knowing Jagex. For now it indeed is just for setting the bot status since it actually building so it's accurate on its action.

gmason0 (changes requested):
I think if we want to support the "reset" action, we should be a bit more conditional. What if the player hasn't setup any birdhouses at all? 

The script appears to use getGameObject via WorldPoint. Possibly we check the object id to see if its empty, build the bird house & seed it, otherwise use the reset option like you have implemented here?

UhWatchMe:
> I think if we want to support the "reset" action, we should be a bit more conditional. What if the player hasn't setup any birdhouses at all?
> 
> The script appears to use getGameObject via WorldPoint. Possibly we check the object id to see if its empty, build the bird house & seed it, otherwise use the reset option like you have implemented here?

I personally don't feel this is needed, since once you complete Bone Voyage, you set up the birdhouses manually anyways and they will always be there.

gmason0:
I do feel that there could be a chance of it not being there - we should consider this & conditionally handle if it is there/not there or we just keep it the way it is to build the birdhouse manually.

UhWatchMe:
> I do feel that there could be a chance of it not being there - we should consider this & conditionally handle if it is there/not there or we just keep it the way it is to build the birdhouse manually.

I have been running this local change for a long long time, and never had issue's. The only way this could happen if someone removes it manually. Even playing legit, birdhouses are always present, so I really find it useless for the rare case someone might have manually removed it.

gmason0:
Considering we are at standstill on the design aspect, I am going to close this - i will be happy to merge this once we have the requested functionality

PR #1136, opened 2025-06-28
Qol update

oversized93:
Update to the QOL plugin to add auto-prayer features.  detects a large portion of available weapons (and some animations) to call for the player to pray against that attack type automatically.  can also automatically enable protect item when attacked, and has a more/less aggressive mode based on user preference.


gmason0:
@See1Duck - Thoughts on this being an addition to QOL ?

See1Duck:
> @See1Duck - Thoughts on this being an addition to QOL ?

Seems fine to me

PR #1137, opened 2025-06-29
Rework: thieving plugin adapted to vyres and better organized

jesufh:
Hello again, again tinkering with the thiev plug as a hobby. I did a rework, organized it and added a few things to make it adapt to Darkmeyer and his damn vyrewatchs. 💗

### Checking if the NPC is inside the room or not (darkmeyer):
thanks duck for your snippet. Now check if the npc is inside the room, being you inside it. In case you are inside and the npc is outside, it will wait for a few seconds before it will hop world.

### Close the doors of the room:
already positioned inside the room, close the nearby doors. I haven't added an anti-ratkids system to avoid the character to close the doors every time.


runelite-client/src/main/java/net/runelite/client/plugins/microbot/mining/motherloadmine/MotherloadMineScript.java
jesufh, line 1: this is a likely fix for the following problem. It will only return that the vein ran out when the vein stored in memory is not found.
https://github.comhttps://github.com/chsami/Microbot/issues/1072/chsami/Microbot/issues/1072

runelite-client/src/main/java/net/runelite/client/plugins/microbot/thieving/ThievingConfig.java
```diff
@@ -11,172 +11,157 @@
 @ConfigGroup("Thieving")
 public interface ThievingConfig extends Config {
 
-    @ConfigItem(
```
gmason0, line 14: We should create a `@ConfigInformation` to add some information on how to use the features of this plugin.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/thieving/ThievingScript.java
```diff
@@ -1,353 +1,385 @@
...
     }
 
-    private void bank() {
-        Microbot.status = "Getting food from bank...";
-
-        BankLocation nearestBank = Rs2Bank.getNearestBank();
+    private void closeNearbyDoor(int radius) {
+        Rs2GameObject.getAll(
+            o -> {
+                ObjectComposition comp = Rs2GameObject.convertToObjectComposition(o);
```
gmason0, line 252: This is fine here as the radius is very small that we are searching around, but just understand that doing this very expensive so please be mindful of this in the future.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/mining/motherloadmine/MotherloadMineScript.java
gmason0: I will commit this for now, but will look to see if we can improve this further

gmason0 (changes requested):
Looks good! I think we should just add a proper Config description to the thieving considering you've modified it vastly - functionality is still there, we just want to explain how to use it

jesufh:
@g-mason0 all ready, check it and sorry for the delay

runelite-client/src/main/java/net/runelite/client/plugins/microbot/thieving/ThievingScript.java
jesufh: if you mean for kidrats that can damage the bot by opening the doors, yes, but then with more time I plan something for that, or someone to give me a hand helps.

gmason0 (approved):

PR #1143, opened 2025-07-01
Rs2Magic rework

Kryaex:


runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/magic/Rs2Magic.java
```diff
@@ -416,601 +361,252 @@ public static boolean isShadowVeilActive() {
...
-        // Remove runes provided by equipped tomes
-        Rs2ItemModel equippedShield = Rs2Equipment.get(EquipmentInventorySlot.SHIELD);
-        if (equippedShield != null) {
-            Rs2Tome equippedTome = Rs2Magic.getRs2Tome(equippedShield.getId());
-            if (equippedTome != Rs2Tome.NONE) {
-                equippedTome.getRunes().forEach(requiredRunes::remove);
-            }
-        }
-        return getRequiredRunes(requiredRunes, checkRunePouch);
+    public static final Map<Integer, int[]> COMBO_RUNES = ImmutableMap.of(
```
gmason0: Maybe these contants better belong within `Runes` ?

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/magic/Rs2Spells.java
```diff
@@ -676,31 +673,35 @@ public enum Rs2Spells {
...
-        this.action = action;
-        this.requiredRunes = requiredRunes;
+        this.magicAction = action;
+        this.requiredRunes = Spell.convertRequiredRunes(requiredRunes);
         this.spellbook = spellbook;
-        this.name = action.getName();
-        this.requiredLevel = action.getLevel();
+    }
+
+    public String getName() {
```
gmason0: Is there any reason why we would remove these from the constructor & create getter methods instead?

gmason0 (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/magic/Rs2Magic.java
Kryaex: That would require an Id -> Runes lookup for all runes in this method.
https://github.com/chsami/Microbot/blob/f8f8908a38fa8628c1d3ec18639e82a4897eac25/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/magic/Rs2Magic.java#L482-L492
I can change all Rs2Magic getXRunes() -> Map<Runes, Integer>, but for consistency that should probably also include the Rs2RunePouch getRunes().

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/magic/Rs2Spells.java
Kryaex: It was just a duplicate reference; this is minimally more memory efficient. But mostly I just thought this was easier to understand where the value is coming from. If you don't like it, I can change it back

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/magic/Rs2Magic.java
gmason0: I think it makes more sense we keep the Rune, Quantity signature

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/magic/Rs2Spells.java
gmason0: Let's revert if you don't mind

gmason0 (approved):

PR #1144, opened 2025-07-01
Moons of Peril Plugin - Beta release

FunkyMonkeyCloud:


runelite-client/src/main/java/net/runelite/client/plugins/microbot/moonsOfPeril/moonsOfPerilConfig.java
```diff
@@ -0,0 +1,233 @@
...
+    @ConfigItem(
+            keyName  = "enableEclipse",
+            name     = "Fight Eclipse Moon",
+            description = "Untick to skip Eclipse Moon runs",
+            position = 0,
+            section  = eclipseMoonSection
+    )
+    default boolean enableEclipse() { return true; }
+
+    @ConfigItem(
```
gmason0, line 128: I would like to see usage of Rs2InventorySetups here instead, having a large list of configuration options is something we would like to move away from.

A better approach would be to define setups for each phase & switching between those setups. One of them can be used as a "main" setup to check against to verify the required items you are looking for exist

gmason0 (changes requested):
Great Work! Would like to see inventory setups be implemented as it will vastly reduce the end-user plugin configuration

runelite-client/src/main/java/net/runelite/client/plugins/microbot/moonsOfPeril/enums/GameObjects.java
Kryaex, line 1: Why aren't you using `net.runelite.api.gameval` values?
`ObjectID.PMOON_ENTRY_STATUE_BLOOD`
`ObjectID.PMOON_ENTRY_STATUE_BLUE`
`ObjectID.PMOON_ENTRY_STATUE_ECLIPSE`
`NpcID.PMOON_BOSS_BLOOD_MOON_VIS`
`NpcID.PMOON_BOSS_ECLIPSE_MOON_VIS`
`NpcID.PMOON_BOSS_BLUE_MOON_VIS`
`NpcID.PMOON_BOSS_THREAT_CIRCLE`

gmason0 (approved):

PR #1147, opened 2025-07-01
Added Daeyalt essence mine feature

translibrius:


runelite-client/src/main/java/net/runelite/client/plugins/microbot/mining/AutoMiningScript.java
```diff
@@ -36,8 +36,34 @@ public class AutoMiningScript extends Script {
     public static final String version = "1.4.4";
     private static final int GEM_MINE_UNDERGROUND = 11410;
     private static final int BASALT_MINE = 11425;
+    private static final int DAEYALT_MINE = 39095;
     State state = State.MINING;
 
+    private void mineDaeyalt(AutoMiningConfig config) {
+        GameObject minableEssence = Rs2GameObject.getGameObject(DAEYALT_MINE);
```
gmason0, line 43: Does a name search not work for this at all?

gmason0 (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/mining/AutoMiningScript.java
SteffenCarlsen: (A little hijack) There are both mineable and unmineable daeyalt with the same name

translibrius: Yes, both the mineable (id `39095`) and unmineable (id `39094`) are both named "Daeyalt Essence".

translibrius: Now that i thinmk about it, we should ignore the full inventory check if mining daeyalt, since you can fill your inventory with gems slowly over time

gmason0: I believe something like this would suffice.

```java
List<TileObject> rocks = Rs2GameObject.getAll(nameAndActionMatches(config.ORE().getName(), "mine"), initialPlayerLocation, config.distanceToStray());
TileObject rock = rocks.stream().findFirst().orElse(null);
```

gmason0:
I am going to close this for now, let me know once you have the opportunity to implement & we can revisit

PR #1150, opened 2025-07-02
Refactor BlastoiseFurnaceScript for maintainability

TheHaliax:
Initial stage of schedulable transition
Refactored BlastoiseFurnaceScript to use new gameval ItemID and VarbitID constants, consolidated ore run logic, and replaced hardcoded values with constants. Improved hybrid and standard bar handling with dispatch methods, added coffer management logic, and cleaned up redundant code. Updated Bars enum to use new constants and improved clarity.


runelite-client/src/main/java/net/runelite/client/plugins/microbot/blastoisefurnace/BlastoiseFurnacePlugin.java
```diff
@@ -45,6 +44,7 @@ BlastoiseFurnaceConfig provideConfig(ConfigManager configManager) {
 
     @Override
     protected void startUp() throws AWTException {
+        BlastoiseFurnaceScript.init = false;
```
gmason0: We could make this var private & non-static & set it to false within shutdown() method of the script class

runelite-client/src/main/java/net/runelite/client/plugins/microbot/blastoisefurnace/BlastoiseFurnaceScript.java
```diff
@@ -229,73 +237,55 @@ private void handleDispenserLooting() {
...
+            Rs2Bank.withdrawAll(ore);
+            sleep(500, 1200);
+            return;
+        }
+        if (!Rs2Inventory.interact(coalBag, "Fill"))
+            return;
 
+        sleep(500, 1200);
+        Rs2Bank.closeBank();
+        sleep(500, 1200);
```
gmason0: Shouldn't this be `sleepUntil(() -> !Rs2Bank.isOpen())` ?

runelite-client/src/main/java/net/runelite/client/plugins/microbot/blastoisefurnace/BlastoiseFurnaceScript.java
```diff
@@ -693,11 +528,92 @@ private void applyAntiBanSettings() {
...
 
-
     public void shutdown() {
         state = State.BANKING;
         primaryOreEmpty = false;
         secondaryOreEmpty = false;
         super.shutdown();
     }
+
+    public int evaluateCofferDeposit()
```
gmason0, line 558: I don't think using assertions here is the right move , this is going to require the SchedulerPlugin to be enabled in-order to use the BF plugin

gmason0 (changes requested):

TheHaliax:
ready for Plugin Scheduler

runelite-client/src/main/java/net/runelite/client/plugins/microbot/blastoisefurnace/BlastoiseFurnaceScript.java
Voxsylvae: I still have significant concerns (as I mentioned in the discord) regarding the current usage of the plugin time estimation methods getTimeBeforeNextScheduledPlugin() and getNextScheduledPlugin(). While these functions provide basic timing information, there are fundamental issues with their current implementation and reliability that will be addressed in my upcoming PR.
### Current Limitations:
The primary issue is that these methods only consider the main time-based start condition of the next scheduled plugin. They completely ignore the current plugin's stop conditions and any additional start conditions the upcoming plugin might have. This creates a false sense of predictability when the actual scheduling behavior is far more complex.
### Method Scope and Naming Issues:
The current method names are confusing and will be updated in the next PR. More importantly, these methods only provide information about when a plugin is theoretically "due to run" based on time conditions alone. However, being "due to run" does not guarantee that the plugin will actually be scheduled, which is a critical distinction.
### Complex Conditional Logic:
My major concern centers around user-defined stop conditions, which can be incredibly diverse and complex. Users can define stop conditions based on relative experience gained, level progression, gold accumulated, number of items processed, time-based triggers, or any logical combination of these factors. Similarly, the next plugin's start conditions can involve multiple criteria beyond simple timing.
The scheduler cannot reliably predict whether the current plugin will be stopped when the next plugin becomes "due to run" because of these complex logical condition structures. Both the start conditions for upcoming plugins and the user-defined stop conditions for the current plugin create too many variables for accurate prediction.
### Recommendation Against Critical Usage:
Given these limitations, I strongly advise against relying on these time estimation methods for critical plugin decisions, such as determining whether to deposit gold to the coffer or making other resource management choices. The uncertainty introduced by conditional logic makes these estimates unreliable for anything beyond basic informational purposes.
### Future Considerations:
I'm considering  on providing functions which gives a better estimate on the next time a plugin will be scheduled, including the nested logical start and stop condition structure. So, I'm working on developing more sophisticated estimation methods that can provide better predictions by accounting for the full complexity of the conditional system, though even these will have inherent limitations due to the dynamic nature of user-defined conditions.


**The bottom line is that while these methods can provide a rough estimate of upcoming plugin timing, they should not be trusted for critical decision-making in plugin logic due to the complex and unpredictable nature of the scheduler's conditional systems**.

TheHaliax: I absolutely understand, and have implemented many fallbacks to ensure that the coffer is never filled by a margin that would cause significant personal loss.
This includes removing and deposing excess gold, refills as required, etc.
I spent several hours trying to find edge cases where the current implementation would fail.
Would be more than happy to work with you to develop a method that dynamically calculates a time based on user conditions, what do you think?

gmason0: @TheHaliax - This still doesn't address my concern of - topping off the coffer now requires the scheduler plugin to be running?

gmason0 (changes requested):

TheHaliax:
this will handle coffer regardless of what plugin is enabled or disabled
if plugin scheduler is enabled it will grab the time to next plugin or if that is not available it will grab break time from the plugin scheduler
if the scheduler is not enabled it will grab break in from the breakhandler
and if neither is true it will deposit 72000 to allow one hour of runtime
and if the coffer  goes empty and it is for some reason still running it will fall back and redeposit
if the plugin get's shutdown and  restarted with a lower expected run time it will withdraw the excess from coffer and continue
if the value is higher it will add  to the coffer and continue
i have tested every possible scenario to the best of my ability and i don't use the plugin scheduler at this time, so it would be impossible for this to work in my general use case if this relied on the plugin scheduler being enabled.
this simply grabs a value and falls through to backup logic if that value is null
same for breakhandler.

please let me know if any further changes are required

gmason0 (approved):

PR #1153, opened 2025-07-03
Rs2Equipment - Overload Methods, Deprecate Old Methods

Kryaex:


runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/inventory/Rs2ItemModel.java
```diff
@@ -312,4 +315,23 @@ public String toString() {
         sb.append("}");
         return sb.toString();
     }
+
+    public static boolean matches(Rs2ItemModel item, boolean exact, String... names) {
```
gmason0: I think these could be better suited as predicate type, instead of returning a boolean.

gmason0 (changes requested):

gmason0 (approved):

PR #1154, opened 2025-07-03
fix (Spell): provide a mutable copy of req runes

Kryaex:


gmason0 (approved):
Don't think we need the @Override annotation consider we are simply providing an implementation of a method, not overriding a default method - either way fine by me

PR #1157, opened 2025-07-04
Expand shop container ID detection with more keywords

TheHaliax:
The getShopContainerIds method now detects container IDs with field names containing 'shop', 'store', 'merchant', 'bazaar', 'stall', or 'trader'. This broadens support for identifying shop-related inventories by matching additional relevant keywords.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/MicrobotPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/MicrobotPlugin.java
@@ -201,32 +201,43 @@ public class MicrobotPlugin extends Plugin
 
 	/**
 	 * Retrieves all container IDs from {@link net.runelite.api.gameval.InventoryID}
-	 * that contain the word "shop" in their field names.
-	 * <p>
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/microbot/MicrobotPlugin.java
```diff
@@ -201,32 +201,43 @@ else if (Arrays.stream(getShopContainerIds()).anyMatch(sid -> Objects.equals(eve
...
 		for (Field field : fields)
 		{
-			if (field.getType() != int.class) continue;
+			if (field.getType() != int.class)
+				continue;
+
+			String fieldName = field.getName().toLowerCase();
 
-			if (field.getName().toLowerCase().contains("shop"))
+			for (String keyword : keywords)
```
gmason0: Instead of another inner for loop, we should be able to use `Arrays.stream(keywords).anyMatch(k -> fieldName.contains(k))`

gmason0 (changes requested):

gmason0 (approved):

PR #1158, opened 2025-07-05
Rs2Tile - Rework

Kryaex:


gmason0:
@Kryaex - It would be helpful to include a brief overview of the changes and the reasoning behind them for clarity.

Kryaex:
> @Kryaex - It would be helpful to include a brief overview of the changes and the reasoning behind them for clarity.

`dangerousTiles: List<MutablePair<WorldPoint, Integer>> -> ConcurrentHashMap<WorldPoint, Integer>` 
Map allows lookup of WorldPoints in O(1), and ConcurrentHashMap makes it Thread Safe.
Made the getter view immutable to restrict modifications to it.

Other changes like getFlags() replace the old, deprecated ones.
`isFlagSet(int data, int flag)` is a faster and functionally identical to the previous check.

gmason0 (approved):

PR #1159, opened 2025-07-05
Add Blessed Wine prayer automation plugin

TheHaliax:
Introduces the Blessed Wine plugin for automating prayer training using blessed wines. Includes configuration, overlay, main plugin logic, state management, and the core script to handle banking, walking, blessing, and prayer restoration cycles.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/hal/blessedwine/BlessedWinePlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/hal/blessedwine/BlessedWinePlugin.java
@@ -63,12 +63,6 @@ public class BlessedWinePlugin extends Plugin implements SchedulablePlugin {
     }
 
     @Subscribe
-    public void onGameTick(GameTick tick) {
-        if (!Microbot.isLoggedIn()) return;
-        // Future overlay or reaction logic
-    }
-
-    @Subscribe
     public void onPluginScheduleEntrySoftStopEvent(PluginScheduleEntrySoftStopEvent event) {
         if (event.getPlugin() == this) {
             Microbot.stopPlugin(this);
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/hal/blessedwine/BlessedWinePlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/hal/blessedwine/BlessedWinePlugin.java
@@ -1,0 +1,81 @@
+package net.runelite.client.plugins.microbot.hal.blessedwine;
+
+import com.google.inject.Provides;
+import lombok.extern.slf4j.Slf4j;
+import net.runelite.api.events.GameTick;
+import net.runelite.client.config.ConfigManager;
+import net.runelite.client.eventbus.Subscribe;
+import net.runelite.client.plugins.Plugin;
+import net.runelite.client.plugins.PluginDescriptor;
+import net.runelite.client.plugins.microbot.Microbot;
+import net.runelite.client.plugins.microbot.pluginscheduler.api.SchedulablePlugin;
+import net.runelite.client.plugins.microbot.pluginscheduler.condition.logical.AndCondition;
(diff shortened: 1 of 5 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/microbot/hal/blessedwine/BlessedWinePlugin.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    @Override
+    protected void shutDown() {
+        blessedWineScript.shutdown();
+        overlayManager.remove(blessedWineOverlay);
+    }
+
+    @Subscribe
+    public void onGameTick(GameTick tick) {
```
gmason0: It would be best to remove events that are unused.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/hal/blessedwine/BlessedWineScript.java
```diff
@@ -0,0 +1,256 @@
...
+
+import javax.inject.Inject;
+import java.util.Arrays;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+@Slf4j
+public class BlessedWineScript extends Script {
+    @Inject
+    private ConfigManager configManager;
```
gmason0: Instead of injecting your own instance of the clientThread or config manager, lets use Microbot.getConfigManager() or Microbot.getClientThread().

runelite-client/src/main/java/net/runelite/client/plugins/microbot/hal/blessedwine/BlessedWineScript.java
```diff
@@ -0,0 +1,256 @@
...
+        mainScheduledFuture = scheduledExecutorService.scheduleWithFixedDelay(() -> {
+            if (!Microbot.isLoggedIn()) return;
+            int maxPrayerLevel = Microbot.getClient().getRealSkillLevel(Skill.PRAYER);
+            int currentPrayerPoints = (Rs2Player.getPrayerPercentage() * maxPrayerLevel) / 100;
+
+            switch (state) {
+                case INITIALIZING:
+                    BlessedWinePlugin.status = "Initializing run...";
+                    initialize();
+                    if (!super.run()) return;
```
gmason0: This is not proper usage of the super.run() method. This should be placed as the first line within your future.

With how you currently have it, this will only run when in the initialize state.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/hal/blessedwine/BlessedWineScript.java
```diff
@@ -0,0 +1,256 @@
...
+                    } else {
+                        state = BlessedWineState.WALK_TO_BOWL;
+                    }
+                    break;
+
+                case TELEPORT_TO_BANK:
+                    BlessedWinePlugin.status = "Teleporting back to Cam Torum Bank...";
+                    if (Rs2Player.getWorldLocation().getPlane() != 2) {
+                        for (int id : CALCIFIED_MOTH) {
+                            if (Rs2Inventory.hasItem(id)) {
```
gmason0, line 122: Inside of any inner loops, we should be first checking `if (!isRunning()) break;`, this ensures that when the scheduled future is cancelled that we break out of the for loop, instead of continuing.

gmason0 (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/hal/blessedwine/BlessedWinePlugin.java
TheHaliax: my bad i had planned on doing more but realized most of it wasn't required and forgot to delete the extras

TheHaliax:
my bad i had planned on doing more but realized most of it wasn't required and forgot to delete the extras

gmason0 (approved):

PR #1160, opened 2025-07-05
update house thieving to thief wealthy citizens outside of aurelia ev…

simpmaxxin:
…ent, added food and dodgy necklace support, attempt fix for idling at door after entering house, attempt wait for when npc could be leaving house

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/maxxin/housethieving/HouseThievingScript.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/maxxin/housethieving/HouseThievingScript.java
@@ -1,6 +1,5 @@
 package net.runelite.client.plugins.microbot.maxxin.housethieving;
 
-import net.runelite.api.GameState;
 import net.runelite.api.Quest;
 import net.runelite.api.QuestState;
 import net.runelite.api.TileObject;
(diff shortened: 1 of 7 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/microbot/maxxin/housethieving/HouseThievingScript.java
```diff
@@ -40,13 +41,18 @@ enum State {
...
     private Long lastThievingSearch = null;
     private ThievingHouse currentThievingHouse = null;
+    private Rs2NpcModel pickpocketNpc = null;
+    private final static String HOUSE_KEYS = "House keys";
+    private final static String DODGY_NECKLACE = "Dodgy necklace";
+    private final static String COIN_POUCH = "Coin pouch";
+    private final static String WEALTHY_CITIZEN = "Wealthy citizen";
 
     public boolean run(HouseThievingConfig config) {
-		Microbot.pauseAllScripts.compareAndSet(true, false);
```
gmason0, line 49: Why was this changed to use set() instead of compareAndSet() ? 

We should always be checking if pauseAllScripts is in the desired state (in this case, its true) then we should set it to false.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/maxxin/housethieving/HouseThievingScript.java
```diff
@@ -55,6 +61,7 @@ public boolean run(HouseThievingConfig config) {
             try {
                 if (!super.run()) return;
                 if (!Microbot.isLoggedIn()) return;
+                if (Microbot.pauseAllScripts.get()) return;
```
gmason0: This is not required, look @ the implementation of super.run()

runelite-client/src/main/java/net/runelite/client/plugins/microbot/maxxin/housethieving/HouseThievingScript.java
```diff
@@ -95,143 +102,204 @@ public boolean run(HouseThievingConfig config) {
...
+        var targetWealthyCitizen = distractedWealthyCitizen != null ? distractedWealthyCitizen : pickpocketNpc;
+        if( targetWealthyCitizen != null ) {
+            if( distractedWealthyCitizen == null ) { // Regular pickpocketing
+                if( Rs2Inventory.getInventoryFood().isEmpty() ) {
+                    state = State.BANKING;
                     return;
                 }
+                if( Rs2Player.getHealthPercentage() <= 50.0 ) {
+                    Rs2Player.useFood();
+                    Rs2Inventory.waitForInventoryChanges(600);
```
gmason0, line 165: May want to extend this, I have had issues with setting this to a lower value such as what you have here - can be an issue with people with slower computers

gmason0 (changes requested):

PR #1169, opened 2025-07-07
Game Chat Logging

Kryaex:
Adds support to log all Slf4j logs in game chat


chsami (approved):

gmason0 (commented):
Shouldn't we tie this into Microbot.log to make this a bit more uniform?

Kryaex:
> Shouldn't we tie this into Microbot.log to make this a bit more uniform?

Could probably also deprecate the Microbot.log(...) methods since now they're pretty much equivalent to Slf4j logging.

gmason0:
> Could probably also deprecate the Microbot.log(...) methods since now they're pretty much equivalent to Slf4j logging.

I think we should just leave it how it is, considering the amount of times Microbot.log is used

gmason0 (approved):

PR #1176, opened 2025-07-08
Detection logic for Imcando Hammer (off-hand)

ghost:
TemporossPlugin.java
Detection logic for Imcando hammer added


gmason0:
This is not required for the Imcando to function correctly.

PR #1181, opened 2025-07-08
tempoross

ghost:
Added Infernal harpoon uncharged for self use. Gave up on add all harpoons, kept fking up the code.

Imcando support in invent and offhand equip version.


ghost:
installed and tested this time. it works. man im too dumb for this. just trying to put a few more QoL in this plug lol.

Toggle for Imancodo hammer and stuff is there. Just can't get it to repair mast/totem. If someone else could look at that. other than that script is working fine. ffs

PR #1186, opened 2025-07-09
fix(EventBus): handle method not found gracefully

Kryaex:


gmason0:
@Kryaex - what does this fix exactly? I presume this is to catch an exception if a Subscribe annotation is added to a method that doesn't have an event?

Kryaex:
> @Kryaex - what does this fix exactly? I presume this is to catch an exception if a Subscribe annotation is added to a method that doesn't have an event?

It should catch all the AbstractMethodErrors resulting from switching how projectiles are created. And instead of crashing the ClientThread simply log an error
`java.lang.AbstractMethodError: Receiver class cy does not define or inherit an implementation of the resolved method 'abstract int getX1()' of interface net.runelite.api.Projectile.
    at com.suppliestracker.SuppliesTrackerPlugin.blowpipeDartCheck(SuppliesTrackerPlugin.java:1286)
    at com.suppliestracker.SuppliesTrackerPlugin.onProjectileMoved(SuppliesTrackerPlugin.java:1271)
    at net.runelite.client.eventbus.EventBus$Subscriber.invoke(EventBus.java:70)
    ...`

Kryaex:
It can be reverted once all plugins have been updated. Just a temporary fix.

gmason0:
I appreciate the explanation - these sort of things should be included in the initial message when opening a pull request.

PR #1188, opened 2025-07-10
Calcified rock miner bugfixes

SteffenCarlsen:
### Changes:
* Added logic to retry mining only if sufficient time has passed since the last XP gain.
* Modified the crushing logic to continuously process deposits until the inventory is empty.
* Updated the plugin's configuration description to provide detailed instructions on starting conditions and recommended gear.

Changed commit messages too, according to your suggestions, hope this is how you wanted it :)


gmason0 (approved):
love you long time

PR #1189, opened 2025-07-10
chore(Rs2Tab): consolidate tab switching

Kryaex:


runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/tabs/Rs2Tab.java
```diff
@@ -52,171 +53,128 @@ public static InterfaceTab getCurrentTab() {
...
     public static boolean switchToMusicTab() {
-        if (getCurrentTab() == InterfaceTab.MUSIC) {
-            return true;
-        }
-        Rs2Keyboard.keyPress(getKeyBind(Microbot.getVarbitValue(4688), InterfaceTab.MUSIC));
-        return getCurrentTab() == InterfaceTab.MUSIC;
+        return switchTo(InterfaceTab.MUSIC);
     }
 
+    // TODO: make private and/or incorporate widget finding for all switchTo?
```
gmason0: I don't think we need to incorporate widget finding for all of them. By default, the logout hot key is not configured.

Kryaex: > I don't think we need to incorporate widget finding for all of them. By default, the logout hot key is not configured.

Yeah, was mostly just for when I or someone else removes all the other deprecated methods that they know why that one specifically is still called

Kryaex: I think I saw some widget ids in the `InterfaceTab` class though so it might not be to hard

gmason0 (approved):

PR #1193, opened 2025-07-10
chore(FixedSizeQueue): refactor

Kryaex:
- Removes unnecessary code
- Allows generic `Queue<T> queue = new FixedSizeQueue<>();`


SteffenCarlsen:
Doesn't this potentially allow breaking of the principle of the fixed size queue? Extending LinkedList (and not overriding all underlying methods to add) should expose methods such as addAll that doesn't have any checks on the size.

Kryaex:
> Doesn't this potentially allow breaking of the principle of the fixed size queue? Extending LinkedList (and not overriding all underlying methods to add) should expose methods such as addAll that doesn't have any checks on the size.

I added checks to addAll & add(index, obj). Push, offer etc. all rely on add, addFirst, addLast as far as I can see.

SteffenCarlsen (approved):
LGTM

gmason0 (approved):

PR #1207, opened 2025-07-12
feat(agility): add werewolf agility course, update agility course han…

simpmaxxin:
…dler with handle ground/game object interface to try and move more course specific logic into course handler classes, expose stick tile in agility plugin

wouldn't mind some feedback for adding those interfaces, wanted to handle all the werewolf edge cases within the AgilityCourseHandler implementation itself

also could use testing for failing the course, as i think you stop failing at 75 agility

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/bee/chaosaltar/ChaosAltarScript.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/bee/chaosaltar/ChaosAltarScript.java
@@ -1,8 +1,8 @@
 package net.runelite.client.plugins.microbot.bee.chaosaltar;
 
+import lombok.extern.slf4j.Slf4j;
 import net.runelite.api.GameObject;
 import net.runelite.api.Skill;
-import net.runelite.api.TileObject;
 import net.runelite.api.coords.WorldArea;
 import net.runelite.api.coords.WorldPoint;
 import net.runelite.client.plugins.microbot.Microbot;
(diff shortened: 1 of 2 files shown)


gmason0 (changes requested):
Considering we have a large majority of the courses here already & we are only implementing these two new methods for one course, i think we should place these within the werewolf course class & implement them into the script similar to how the priff portal is handled

simpmaxxin:
> Considering we have a large majority of the courses here already & we are only implementing these two new methods for one course, i think we should place these within the werewolf course class & implement them into the script similar to how the priff portal is handled

could we go in the direction of overriding getCurrentObstacle inside WerewolfCourse instead and removing the new interfaces?
the core of the issue lies in the Predicate<TileObject> only matches based on ObjectID but that's not enough to select the correct obstacle -
image
image
the call to Rs2GameObject.getAll with validObjectPredicate returns multiple matched TileObject/GameObject but we only get the first one, where in this case we need to determine the correct one from the multiple found for at least 2 scenarios

can also put the logic within AgilityCourseHandler under the instanceof GroundObject/GameObject switch statements instead

gmason0:
I believe these instances you've shown those far can be filtered out if you specify more strict x & y operations?

simpmaxxin:
> I believe these instances you've shown those far can be filtered out if you specify more strict x & y operations?

unfortunately matching a specific AgilityObstacleModel doesn't help in this case since it only provides what ObjectID we are looking for later on in the Predicate<TileObject> - 
image
adjusting the canReach doesn't help either as we are on a TileObject with ID 11643 and need to get to the next one with 11643 but have no way of knowing which one is correct - without an adjustment to that Predicate<TileObject> logic we just repeatedly click the same TileObject we are standing on

gmason0:
I don't see an issue with overriding the getCurrentObstacle method, I just dont think we need to add new methods to the interface that only get used for 1 course.

PR #1213, opened 2025-07-14
chore(Rs2Bank): refactor

Kryaex:
- changes return values: void -> boolean
- changes withdrawX(boolean checkInv, ..., int amount) to check if inv_amount >= amount
- adds early returns

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/bank/Rs2Bank.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/bank/Rs2Bank.java
@@ -128,23 +128,12 @@ public class Rs2Bank {
     }
 
     /**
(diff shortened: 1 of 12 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/bank/Rs2Bank.java
```diff
@@ -1907,6 +1898,10 @@ public static boolean hasCachedBankData() {
         return !rs2BankData.isEmpty();
     }
 
+    private static final String[] DIGIT_INSTRUCTIONS = {
```
gmason0: Let's consider scope here, i don't see how/why this would ever be re-used.

gmason0 (commented):
Looks good thus far

gmason0 (approved):

PR #1220, opened 2025-07-15
fix(FishingTrawler): Change overlay text to be correct

ghost:
Fixed the overlay text in the FishingTrawlerOverlay, which still showed "Micro Example V1.0.0".


gmason0:
This is already being handled in a different pull, thank you for the effort

PR #1221, opened 2025-07-15
feat(AutoWoodcutting): Do not drop crystal shards when woodcutting in Prifddinas

ghost:
(Suggestion from discord by ajs74)

When woodcutting in Prifddinas, you gain crystal shards which are normally not dropped, take one inventory slot and are very valuable. People who do woodcutting in Prifddinas usually also do it for the crystal shards, so they should probably not be dropped.


gmason0 (changes requested):
Please rebase so conventional commits check passes

PR #1222, opened 2025-07-16
feat(MicrobotMouseOverlay): overlay set as naughty to not appear on screenshots

Bolado:
To avoid this happening on certain achievements:

image


gmason0 (changes requested):
Please rebase so the conventional commits pass

PR #1225, opened 2025-07-16
Fix(DemonicGorillaScript): General bugfixes

SteffenCarlsen:
**Changes:**

- Warn user if their inventory setups are not set correctly
- Disable QoLScript to stop intervention with banking/restocking logic
- Stop LooterScript from accidently disabling all scripts
    - This is a sideffect of 6163a4b2e944408e47342e81a4ec39ea9eaafeaa
- Clean-up of unused variables
- Add some extra failsafes
- Version bump

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/TaF/DemonicGorillaKiller/DemonicGorillaLooterScript.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/TaF/DemonicGorillaKiller/DemonicGorillaLooterScript.java
@@ -19,9 +19,12 @@ public class DemonicGorillaLooterScript extends Script {
     public boolean run(DemonicGorillaConfig config) {
         mainScheduledFuture = scheduledExecutorService.scheduleWithFixedDelay(() -> {
             try {
+                if (DemonicGorillaScript.BOT_STATUS.equals(DemonicGorillaScript.State.BANKING) || DemonicGorillaScript.BOT_STATUS.equals(DemonicGorillaScript.State.TRAVEL_TO_GORILLAS)) {
+                    Microbot.pauseAllScripts.set(false);
+                    return;
+                }
                 if (!super.run()) return;
                 if (!Microbot.isLoggedIn()) return;
-                if (DemonicGorillaScript.BOT_STATUS.equals(DemonicGorillaScript.State.BANKING) || DemonicGorillaScript.BOT_STATUS.equals(DemonicGorillaScript.State.TRAVEL_TO_GORILLAS)) return;
                 if (Rs2Inventory.isFull() || Rs2Inventory.getEmptySlots() <= minFreeSlots) return;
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/microbot/TaF/DemonicGorillaKiller/DemonicGorillaLooterScript.java
```diff
@@ -19,9 +19,12 @@ public DemonicGorillaLooterScript() {
     public boolean run(DemonicGorillaConfig config) {
         mainScheduledFuture = scheduledExecutorService.scheduleWithFixedDelay(() -> {
             try {
+                if (DemonicGorillaScript.BOT_STATUS.equals(DemonicGorillaScript.State.BANKING) || DemonicGorillaScript.BOT_STATUS.equals(DemonicGorillaScript.State.TRAVEL_TO_GORILLAS)) {
+                    Microbot.pauseAllScripts.set(false);
```
gmason0: can we use 

`Microbot.pauseAllScripts.compareAndSet(true,  false)`

runelite-client/src/main/java/net/runelite/client/plugins/microbot/TaF/DemonicGorillaKiller/DemonicGorillaScript.java
```diff
@@ -134,7 +122,42 @@ public boolean run(DemonicGorillaConfig config) {
         return true;
     }
 
+    private void preflightChecks(DemonicGorillaConfig config) {
+        if (config.useMagicStyle() && config.magicGear() == null) {
```
gmason0, line 126: For these checks, can we use Rs2InventorySetup.isInventorySetup() ?

gmason0 (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/TaF/DemonicGorillaKiller/DemonicGorillaScript.java
SteffenCarlsen: No, config.magicGear is an InventorySetup, while Rs2InventorySetup.isInventorySetup looks for a string representation of the name of an inventorysetup. This is to catch the case where a user blindly assumes that the UI showing a selected inventorysetup, also means that the inventorysetup is selected, but in reality, the inventorysetup is null.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/TaF/DemonicGorillaKiller/DemonicGorillaLooterScript.java
SteffenCarlsen: Changed

gmason0 (approved):

PR #1227, opened 2025-07-17
fix(agility):prif course

Sorries:


runelite-client/src/main/java/net/runelite/client/plugins/microbot/agility/courses/PrifddinasCourse.java
```diff
@@ -26,7 +27,7 @@ public class PrifddinasCourse implements AgilityCourseHandler
 		ObjectID.PRIF_AGILITY_SHORTCUT_PORTAL_6
 	);
 
-	private final WorldArea prifFallArea = new WorldArea(3260, 6103, 15, 17, 0);
+	private final WorldArea prifFallArea = new WorldArea(3260, 6103, 15, 9, 0);
```
gmason0, line 30: Will this not cause someone to get stuck next to the dark hole obstacle? atleast this was the reason why i made the world area bigger

gmason0 (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/agility/courses/PrifddinasCourse.java
Sorries: the thing is that the dark hole should not be included as it is part of the obstacle course. this prif area only accounts for the first 2 failure points or else it would try to go back to the starting point instead of to the next obstacle .

gmason0: We do still need to account for it as getCurrentObstacleIndex() returns 4 when standing outside of that area, then you will be stuck spam clicking it

Sorries:
```
if (prifFallArea.contains(playerWorldLocation))
		{
			Rs2Walker.walkTo(getStartPoint(), 5);
			Microbot.log("Going back to course's starting point");
			return true;
		}

		if (getCurrentObstacleIndex() > 0)
		{
			return false;
		}
```

this should allow it to walk back to the beginning and not spam click it

PR #1233, opened 2025-07-20
fix(Rs2InventorySetup): Broken Barrows items

Gage307:
Added a check to the withdraw section for barrows items; the old code would withdraw a broken barrows item and try to equip it

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/Rs2InventorySetup.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/Rs2InventorySetup.java
@@ -237,6 +237,10 @@ public class Rs2InventorySetup {
 			}
 		} else {
 			if (useName) {
+				if (isBarrowsItem(item.getName())) {
+					item.setName(item.getName().replaceAll("\\s+[1-9]\\d*$", ""));
+				}
+
 				Rs2Bank.withdrawItem((String) identifier);
 			} else {
 				Rs2Bank.withdrawItem((int) identifier);
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/Rs2InventorySetup.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/Rs2InventorySetup.java
@@ -237,10 +237,6 @@ public class Rs2InventorySetup {
 			}
 		} else {
 			if (useName) {
-				if (isBarrowsItem(item.getName())) {
-					item.setName(item.getName().replaceAll("\\s+[1-9]\\d*$", ""));
-				}
-
 				Rs2Bank.withdrawItem((String) identifier);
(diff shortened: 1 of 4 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/Rs2InventorySetup.java
```diff
@@ -237,6 +237,10 @@ private void withdrawItem(InventorySetupsItem item, int quantity) {
 			}
 		} else {
 			if (useName) {
+				if (isBarrowsItem(item.getName())) {
```
chsami: normalization should happen at the top of the method, it's a bit cleaner
```
private void withdrawItem(InventorySetupsItem item, int quantity) {
	boolean useName = item.isFuzzy();

	// Normalize item name if needed
	String itemName = item.getName();
	if (useName && isBarrowsItem(itemName)) {
		itemName = normalizeBarrowsName(itemName);
		item.setName(itemName); // Optional: if other logic depends on the updated name
	}

	Object identifier = useName ? itemName.toLowerCase() : item.getId();

	// Withdraw logic
	if (quantity > 1) {
		if (useName) {
			Rs2Bank.withdrawX((String) identifier, quantity);
		} else {
			Rs2Bank.withdrawX((int) identifier, quantity);
		}
	} else {
		if (useName) {
			Rs2Bank.withdrawItem((String) identifier);
		} else {
			Rs2Bank.withdrawItem((int) identifier);
		}
	}

	sleepUntil(() -> Rs2Inventory.hasItemAmount(itemName, item.getQuantity()));
}

private String normalizeBarrowsName(String name) {
	// Removes trailing number suffix from Barrows items (e.g. "Karil's crossbow 75" → "Karil's crossbow")
	return name.replaceAll("\\s+[1-9]\\d*$", "");
}
```

chsami (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/Rs2InventorySetup.java
gmason0: I think it may make more sense to include this in the InventorySetupItem class within the getName() method ?

chsami: > withdrawItem(InventorySetupsItem item, int quan

Yeah I agree

chsami: > I think it may make more sense to include this in the InventorySetupItem class within the getName() method ?

Yeah I agree

Gage307: You got it!

PR #1250, opened 2025-07-23
Feat/scheduler walker

Kayo-b:
Extra feature added to plugin scheduler that allows for the option to activate the Rs2Walker to a target location before triggering the scheduled plugin.


gmason0:
Hello, first I want to say - thank you for making this pull request.

unfortunately, this currently does not fall in-line with our roadmap with the scheduler. There is an upcoming feature I have been discussing with @Voxsylvae to add scheduled pre & post tasks. These tasks would overlap with this feature you have implemented here. 

Feel free to discuss this with Vox to get a better idea of the timeline of this implementation.

Thanks!

https://github.com/chsami/Microbot/blob/main/docs/scheduler/roadmap.md

PR #1253, opened 2025-07-24
feat(Auto Mining): added functionality for barronite rocks and bank

devxyz-blue:


runelite-client/src/main/java/net/runelite/client/plugins/microbot/mining/AutoMiningScript.java
```diff
@@ -102,15 +103,28 @@ public boolean run(AutoMiningConfig config) {
...
+                            if (rock != null) {
+                                if (Rs2GameObject.interact(rock)) {
+                                    Rs2Player.waitForXpDrop(Skill.MINING, true);
+                                    Rs2Antiban.actionCooldown();
+                                    Rs2Antiban.takeMicroBreakByChance();
+                                }
                             }
                         }
+                        else {
+                            GameObject rock = Rs2GameObject.findReachableObject(config.ORE().getName(), true, config.distanceToStray(), initialPlayerLocation);
```
gmason0, line 117: Instead of continuing to deprecated usages of the Rs2GameObject class, heres a better alternative. This would also avoid having to create a whole separate if statement within the script.

```java
List<TileObject> rocks = Rs2GameObject.getAll(nameAndActionMatches(config.ORE().getName(), "mine"), initialPlayerLocation, config.distanceToStray());
TileObject rock = rocks.stream().findFirst().orElse(null);
```

gmason0 (changes requested):

gmason0:
Any further clarification needed?

devxyz-blue:
none needed. I can fix these, but need some time. I am moving cross country and preparing for all of that.

SteffenCarlsen:
> none needed. I can fix these, but need some time. I am moving cross country and preparing for all of that.

FYI, AIO Camdozal is a thing and pretty much does everything you could like to do in that area :) So not sure if this change is needed tbh

PR #1254, opened 2025-07-24
Rev killer & Rs2Dialogue update

Gage307:
Rev killer: moved drinkStam into the EatFood future

Rs2Dialogue: widget child id for isInDialogue() updated to 566.


runelite-client/src/main/java/net/runelite/client/plugins/microbot/inventorysetups/InventorySetupsItem.java
gmason0, line 1: I did just push some changes in #1255 that align with your changes here, lets just revert this - I appreciate it though!

gmason0 (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/inventorysetups/InventorySetupsItem.java
Gage307: Wow what a coincidence I think we pushed it the same minute. Will do!

PR #1267, opened 2025-07-25
Fix WorldArea constructor usage for proper centered area calculations

Voxsylvae:
## Problem

The `WorldArea(WorldPoint, width, height)` constructor treats the WorldPoint parameter as the south-west corner, not the center. This was causing several issues:

1. **Proximity Detection Failures**: Players standing directly adjacent to targets weren't being detected as "near" the target
2. **Reachability Calculation Errors**: Objects were incorrectly marked as unreachable due to improper area intersection calculations  
3. **Pathfinding Issues**: Area intersection logic was failing in walker methods
4. **Cannon Positioning Problems**: 3x3 centered areas for cannon operations were incorrectly calculated

Multiple utility methods across the codebase were passing center points directly to the WorldArea constructor, which expects south-west corner coordinates.

### Files Modified:

**1. Rs2Player.java**
* `isNearArea(WorldPoint target, int distance)`: Fixed to calculate south-west corner for centered area creation

**2. Rs2GameObject.java**
* `isReachable(WorldPoint worldPoint, int sizeX, int sizeY)`: Added null-safety and proper centered area calculation
* `canReach(WorldPoint worldPoint, int sizeX, int sizeY)`: Fixed area calculation logic

**3. Rs2Walker.java**
* `canReach(WorldPoint worldPoint, int sizeX, int sizeY, int pathSizeX, int pathSizeY, boolean useBankedItems)`: Fixed both object and path area calculations
* `getTotalTilesFromPath(List<WorldPoint> path, WorldPoint destination)`: Fixed area intersection logic for path validation

**4. Rs2Cannon.java**
* `repair()`: Fixed 3x3 centered area calculation for cannon positioning
* `refill()`: Fixed 3x3 centered area calculation for cannon positioning
* Updated deprecated method calls to current API

## Technical Details

The fix involves calculating the south-west corner coordinate using the formula:
```java
WorldPoint southWestCorner = new WorldPoint(
    centerPoint.getX() - (width / 2),
    centerPoint.getY() - (height / 2),
    centerPoint.getPlane()
);
WorldArea centeredArea = new WorldArea(southWestCorner, width, height);
```
This ensures proper area intersection calculations for proximity detection, reachability checks, and pathfinding operations across the entire system.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/player/Rs2Player.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/player/Rs2Player.java
@@ -1117,10 +1117,113 @@ public class Rs2Player {
      * @param worldPoint The {@link WorldPoint} to check proximity to.
      * @param distance   The radius (in tiles) around the {@code worldPoint} to check.
      * @return {@code true} if the player is within the specified distance, {@code false} otherwise.
+     * @deprecated Since 1.9.6, use {@link #isInArea(WorldPoint, int)} for better naming consistency.
      */
-    public static boolean isNearArea(WorldPoint worldPoint, int distance) {
-        WorldArea worldArea = new WorldArea(worldPoint, distance, distance);
-        return worldArea.contains(getWorldLocation());
+    @Deprecated(since = "1.9.6", forRemoval = true)
+    public static boolean isNearArea(WorldPoint worldPoint, int radius) {
+        return isInArea(worldPoint, radius);
+    }
+
+    /**
+     * Checks if the player is within a specified distance of a given {@link WorldPoint}.
+     *
+     * @param worldPoint The {@link WorldPoint} to check proximity to.
+     * @param distance   The radius (in tiles) around the {@code worldPoint} to check.
+     * @return {@code true} if the player is within the specified distance, {@code false} otherwise.
+     */
+    public static boolean isInArea(WorldPoint worldPoint, int radius) {
+        return isInArea(worldPoint, radius, radius);
+    }
+
+    /**
+     * Checks if the player is within a specified area around a given {@link WorldPoint}.
+     *
+     * @param worldPoint The {@link WorldPoint} to check proximity to.
+     * @param xRadius    The horizontal radius (in tiles) around the {@code worldPoint}.
+     * @param yRadius    The vertical radius (in tiles) around the {@code worldPoint}.
+     * @return {@code true} if the player is within the specified area, {@code false} otherwise.
+     */
+    public static boolean isInArea(WorldPoint worldPoint, int xRadius, int yRadius) {
+        // Null check for world point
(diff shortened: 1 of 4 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/player/Rs2Player.java
```diff
@@ -1117,7 +1117,14 @@ public static Rs2WorldPoint getRs2WorldPoint() {
...
     public static boolean isNearArea(WorldPoint worldPoint, int distance) {
-        WorldArea worldArea = new WorldArea(worldPoint, distance, distance);
+        // Create a WorldArea centered on the worldPoint by calculating the south-west corner
+        WorldPoint southWest = new WorldPoint(
+            worldPoint.getX() - distance, 
+            worldPoint.getY() - distance, 
+            worldPoint.getPlane()
+        );
+        // Create area with size (distance * 2 + 1) to include the center tile
+        WorldArea worldArea = new WorldArea(southWest, distance * 2 + 1, distance * 2 + 1);
```
Kryaex: Is there a reason you don't just use `return worldPoint.distanceTo(getWorldLocation()) <= distance` + potential null checks if you want

Voxsylvae: Thanks for the review! You're absolutely this would be suitable here. I adapted this from the previous implementation, but I'll fix that here by adding a function which takes x and y spans for more precise area checking and overload it.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/player/Rs2Player.java
```diff
@@ -1115,10 +1115,117 @@ public static Rs2WorldPoint getRs2WorldPoint() {
...
+        // Null check for player location
+        if (playerLocation == null) {
+            return false;
+        }
+        
+        // Ensure both points are on the same plane
+        if (worldPoint.getPlane() != playerLocation.getPlane()) {
+            return false;
+        }
+
```
Kryaex, line 1168: `(int) Math.floor(xSpan / 2.0)` is equivilant to `xSpan / 2` for non-negative numbers. Also, this is still a distance check albeit now x & y separately. 
`int dx = Math.abs(worldPoint.getX()-playerLocation.getX());
int dy = Math.abs(worldPoint.getY()-playerLocation.getY());
return dx <= xSpan / 2 && dy <= ySpan / 2;` 
Additionally, personally I would prefer using xSpan -> xRadius so the overloads are similar

Voxsylvae: Right, it's the same.

gmason0 (approved):

PR #1268, opened 2025-07-25
new(flipper) automate the flipping copilot plugin

celatus:


runelite-client/src/main/java/net/runelite/client/plugins/microbot/flipper/FlipperConfig.java
```diff
@@ -0,0 +1,26 @@
...
+@ConfigGroup("Flipper Config")
+public interface FlipperConfig extends Config {
+
+        @ConfigItem(
+            keyName = "guide",
+            name = "How to use",
+            description = "How to use this plugin",
+            position = 0
+    )
+    default String GUIDE() {
```
gmason0, line 17: This should be refactored to use `@ConfigInformation` annotation

runelite-client/src/main/java/net/runelite/client/plugins/microbot/flipper/FlipperPlugin.java
```diff
@@ -0,0 +1,43 @@
...
+@PluginDescriptor(
+        name = PluginDescriptor.Choken + "Flipper",
+        description = "Flipping copilot automation",
+        tags = {"flip", "ge", "grand", "exchange", "automation"},
+        enabledByDefault = false
+)
+public class FlipperPlugin extends Plugin {
+    
+    @Inject
+    private Client client;
```
gmason0, line 22: This is an un-used inject, this can be safely removed.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/flipper/FlipperScript.java
```diff
@@ -0,0 +1,309 @@
...
+        // Fallback to existing name-matching approach
+        // Get the item name to abort
+        Field nameField = currentSuggestion.getClass().getDeclaredField("name");
+        nameField.setAccessible(true);
+        String itemName = (String) nameField.get(currentSuggestion);
+        
+        Microbot.log("Using fallback abort method for item: " + itemName);
+        
+        // Abort only the specific item, not all offers
+        boolean success = Rs2GrandExchange.abortOffer(itemName, false);
```
gmason0: Does this below not work? I dont believe we need to reach into the flipping co-pilot plugin for the entire widget if we are able to just find the name of the offer.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/flipper/FlipperScript.java
```diff
@@ -0,0 +1,309 @@
...
+
+        if (highlightController == null) {
+            return;
+        }
+
+        // Execute widget checks on client thread
+        Microbot.getClientThread().invokeLater(() -> {
+            try {
+                // Get the list of highlighted overlays from FlippingCopilot
+                Field highlightOverlaysField = highlightController.getClass().getDeclaredField("highlightOverlays");
```
gmason0: This is repeated quite often - possibly we create a method that returns the highlightOverlays object?

gmason0 (changes requested):
Some minor TLC - but I like the direction thus far.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/flipper/FlipperScript.java
ghost: the fallback method was my original method, but i had to change it because copiliot is weird about the name of some items, it sends sometimes in weird formats, so instead i did a catch-all method of reaching directly to copilot so i can directly check the widget and i would check if its suggestion is abort i would call the abort method

list of some items i got confirmation it doesnt work with:
ardy tele, house tele, chugging barrel, antidote++, superantipoison

i only got 1 screenshot of the debug and that was for the ardy tele, it was sending data: ardougne teleport (tablet), im guessing there might be a way of getting the correct name but i couldnt figure out and i didnt want to add special parsing for each item

ghost: also one more note: i kept the fallback method in case because i hadnt tested the new code thouroughly enough but after multiple days i havent had any problems with it and ill be removing it

gmason0 (approved):

PR #1269, opened 2025-07-25
new(TheMess): Automate The Mess cooking minigame

Bolado:
Good XP without items/gp required :)

squashed commits:

feat(TheMess): new overlay, better logic for combineItems, etc

fix(TheMess): many fixes and improvements
- fixed overlay position by fixing on top of chat widget
- remade the item organization on the last slots
- fixed the last condition of the combine items
- added a lot of more items that can be dropped
- fixed dropping logic
- using a specific nearby bank because doesn't make sense to teleport when one is so close proper reset of things when shutting down

chore(TheMess): cleaning of code, some touchups

chore(TheMess): change PluginDescriptor and touch up on the config

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/mess/TheMessConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/mess/TheMessConfig.java
@@ -1,0 +1,97 @@
+package net.runelite.client.plugins.microbot.mess;
+
+import net.runelite.client.config.Config;
+import net.runelite.client.config.ConfigGroup;
+import net.runelite.client.config.ConfigItem;
+import net.runelite.client.config.ConfigSection;
+import net.runelite.client.plugins.microbot.mess.TheMessScript.Dish;
+
+@ConfigGroup("the_mess")
+public interface TheMessConfig extends Config {
+    @ConfigSection(
+            name = "",
+            description = "General information about the activity and requirements.",
+            position = 0
+    )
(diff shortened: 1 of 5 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/microbot/mess/TheMessConfig.java
```diff
@@ -0,0 +1,97 @@
...
+
+    @ConfigItem(
+            keyName = "info_box",
+            name = "-- ABOUT THE MESS PLUGIN --",
+            description = "Some information about the usage and requirements for the script",
+            position = 0,
+            section = aboutSection
+    )
+    default String info_box() {
+        return "• Wiki:\n" +
```
gmason0: I would try not place information of the plugin within a ConfigItem. If you update this, users would have to reset their config in-order to see the changes.

It would be better if you created a `@ConfigInformation` annotation with the pertinent information - then create a discord channel within the community-plugins section of the discord to provide any of this verbose explaination

gmason0 (changes requested):
Lets just clean up the config / guide

runelite-client/src/main/java/net/runelite/client/plugins/microbot/mess/TheMessConfig.java
Bolado: My bad, totally skipped @ConfigInformation 😄

gmason0 (approved):

PR #1270, opened 2025-07-25
Rs2GrandExchange minor tweaks

Gage307:
Set hasWidgetText(searchName, exact) to false as it would always fail when isExact was true even if it matches.

Added a sleepUntil instead of a generic sleep.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/grandexchange/Rs2GrandExchange.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/grandexchange/Rs2GrandExchange.java
@@ -209,7 +209,7 @@ public class Rs2GrandExchange
 				}
 				Rs2Keyboard.typeString(request.getItemName());
 
-				if (!Rs2Widget.hasWidgetText(searchName, 162, 43, false)) break;
+				if (!Rs2Widget.sleepUntilHasWidgetText(searchName, 162, 43, false, 5000)) break;
 
 				sleepUntil(() -> {
 					Widget widget = Rs2Widget.getWidget(162, 51);
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/grandexchange/Rs2GrandExchange.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/grandexchange/Rs2GrandExchange.java
@@ -209,9 +209,22 @@ public class Rs2GrandExchange
 				}
 				Rs2Keyboard.typeString(request.getItemName());
 
-				if (!Rs2Widget.sleepUntilHasWidgetText(searchName, 162, 43, request.isExact(), 5000)) break;
+				if (!Rs2Widget.hasWidgetText(searchName, 162, 43, false)) break;
 
-				sleep(1800); // TODO: make this conditional.
+				sleepUntil(() -> {
+					Widget widget = Rs2Widget.getWidget(162, 51);
+					if (widget == null) return false;
+
+					Widget[] children = widget.getDynamicChildren();
+					if (children == null || children.length == 0 || children[0] == null || !children[0].getText().isEmpty()) return false;
+
+					String text = Rs2UiHelper.stripColTags(children[0].getName());
+					if (text == null || text.isEmpty()) return false;
+
+					String itemName = request.getItemName();
+					return text.toLowerCase().contains(itemName.toLowerCase()) || text.equalsIgnoreCase(itemName);
+				}, 2200);
+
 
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/grandexchange/Rs2GrandExchange.java
```diff
@@ -209,9 +209,22 @@ public static boolean processOffer(GrandExchangeRequest request) {
 				}
 				Rs2Keyboard.typeString(request.getItemName());
 
-				if (!Rs2Widget.sleepUntilHasWidgetText(searchName, 162, 43, request.isExact(), 5000)) break;
+				if (!Rs2Widget.hasWidgetText(searchName, 162, 43, false)) break;
```
gmason0: I made a slight modification to show the return value of `Rs2Widget.sleepUntilHasWidgetText` - this appears fine to me.
Can you provide an example GrandExchangeRequest where this does fail?
```java
...
Rs2Keyboard.typeString(request.getItemName());

boolean hasSearchResult = Rs2Widget.sleepUntilHasWidgetText(searchName, 162, 43, request.isExact(), 5000);
Microbot.log("Search result: " + hasSearchResult);
if (!hasSearchResult) break;

sleep(1800); // TODO: make this conditional.

Pair<Widget, Integer> itemResult = getSearchResultWidget(request.getItemName(), request.isExact());
...
```

```java
...
GrandExchangeRequest request = GrandExchangeRequest.builder()
	.action(GrandExchangeAction.BUY)
	.itemName("fire rune")
	.percent(6)
	.quantity(1000)
	.build();

Rs2GrandExchange.processOffer(request);
...
```

image

Gage307: ...
GrandExchangeRequest request = GrandExchangeRequest.builder()
	.action(GrandExchangeAction.BUY)
	.itemName("fire rune")
	.percent(6)
	.quantity(1000)
	.build();

Rs2GrandExchange.processOffer(request);
...



Add .exact(true) and this is where I believe the problem lies

GrandExchangeRequest request = GrandExchangeRequest.builder()
	.action(GrandExchangeAction.BUY)
	.itemName("fire rune")
	.percent(6)
        .exact(true)
	.quantity(1000)
	.build();

Rs2GrandExchange.processOffer(request);

gmason0: Hmm, you're right - but I am not sure what you're recommending would be best..

What if we just force this to false as this is the widget text it searches against. this is simply verifying if the string that is typed in the search box had completed. Then `itemResult = getSearchResultWidget(request.getItemName(), request.isExact());` should provide us with correct widget.

idea64_1EpqRBX1Al

Gage307: They say great minds think alike! That's exactly what I meant "Set hasWidgetText(searchName, exact) to false as it would always fail"

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/grandexchange/Rs2GrandExchange.java
```diff
@@ -209,9 +209,22 @@ public static boolean processOffer(GrandExchangeRequest request) {
 				}
 				Rs2Keyboard.typeString(request.getItemName());
 
-				if (!Rs2Widget.sleepUntilHasWidgetText(searchName, 162, 43, request.isExact(), 5000)) break;
+				if (!Rs2Widget.sleepUntilHasWidgetText(searchName, 162, 43, false, 5000)) break;
+
+				sleepUntil(() -> {
```
gmason0: I think this works just as well to replace the static sleep with something conditional 

```
if (!Rs2Widget.sleepUntilHasWidgetText(searchName, 162, 43, false, 5000)) break;

sleepUntil(() -> getSearchResultWidget(request.getItemName(), request.isExact()) != null);
```

gmason0 (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/grandexchange/Rs2GrandExchange.java
Gage307: Yeah that is much cleaner!

gmason0 (approved):

PR #1275, opened 2025-07-26
 AUBURN_SAWMILL option added to PlankRunner

ghost:
AUBURN_SAWMILL option added

- added  new Sawmill Location - AUBURNVALE(new WorldPoint(1395, 3368, 0) 
- editted BankLocation at AUBURNVALE to WorldPoint(1413, 3353, 0) to make it closest to Sawmill
- added NpcID:AUBURN_SAWMILL_OPERATOR to support sawmillOperator at AUBURNVALE


runelite-client/src/main/java/net/runelite/client/plugins/microbot/plankrunner/PlankRunnerScript.java
```diff
@@ -132,6 +132,10 @@ public boolean run() {
                         }
 
                         var sawmillOperator = Rs2Npc.getNpc(NpcID.POH_SAWMILL_OPP);
+                        if (sawmillOperator == null) {
```
gmason0: I think a slightly better approach would be to create a collection of ids & look for all of them @ once.

```java
Set<Integer> sawmillNpcs = Set.of(NpcID.POH_SAWMILL_OPP, NpcID.AUBURN_SAWMILL_OPERATOR);
var sawmillOperator = Rs2Npc.getNpcs(n -> sawmillNpcs.contains(n.getId())).findFirst().orElse(null);
```

gmason0 (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/plankrunner/PlankRunnerScript.java
Kryaex: Shouldn't we just change `getNpcs(int id) -> getNpcs(int... ids)`. If you want to improve lookup times with sets a lot of other code would need to be updated since afaik they almost all use `anyMatch`

gmason0: anyMatch would be perfectly fine in this situation, no? We are just looking for either of these npcs?

Kryaex: He just did what you commented guess I'll change Rs2Npc myself if I ever need to use it. And, yeah anyMatch works you just mentioned Set so I didn't know if you were worried about the performance of it for large sets.

gmason0: I think the performance improvement using a reference type to a primitive type is negligible here - the purpose of using a set was simply to perform a single look up compared to having to look each one up individually by checking if null again

vi3tcn: Set<Integer> sawmillNpcs = Set.of(NpcID.POH_SAWMILL_OPP, NpcID.AUBURN_SAWMILL_OPERATOR);
var sawmillOperator = Rs2Npc.getNpcs(NpcID.POH_SAWMILL_OPP, NpcID.AUBURN_SAWMILL_OPERATOR)
    .findFirst()
    .orElse(null);
    
    
    @Kryaex 
    so you would prefer it like this?

gmason0 (approved):

PR #1279, opened 2025-07-26
chore(Rs2Bank): add util methods, refactor

Kryaex:
- Adds getAll(...) utility methods
- Replaces deprecated value

This could probably use more work I just need to getAll methods for the InvSetup changes I'm working on

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/bank/Rs2Bank.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/bank/Rs2Bank.java
@@ -3,10 +3,21 @@ package net.runelite.client.plugins.microbot.util.bank;
 import com.google.gson.Gson;
 import com.google.gson.JsonSyntaxException;
 import lombok.extern.slf4j.Slf4j;
-import net.runelite.api.*;
+import net.runelite.api.GameObject;
+import net.runelite.api.InventoryID;
+import net.runelite.api.ItemComposition;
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/bank/Rs2Bank.java
```diff
@@ -1536,6 +1551,7 @@ private static AbstractMap.SimpleEntry<List<WorldPoint>, BankLocation> getPathAn
                                 .min(Comparator.comparingInt(b -> Rs2WorldPoint.quickDistance(obj.getWorldLocation(), b.getWorldPoint())))
                                 .orElse(null);
 
+                        assert closestBank != null;
```
gmason0, line 1554: closetBank can be null here if we are not remotely close to a bank object?

Kryaex: It can't `BankLocation closestBank = allBanks.stream()
                                .min(Comparator.comparingInt(b -> Rs2WorldPoint.quickDistance(obj.getWorldLocation(),b.getWorldPoint())))
                                .orElse(null);` is a min comparison not a filter for distance 
unless allBanks is empty which imo should throw an assertion error
could probably change `.orElse(null)` to `orElseThrow(...)`

gmason0 (approved):

PR #1281, opened 2025-07-26
fix(Rs2Bank): speed up required filtering 100x

Kryaex:
image

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/bank/enums/BankLocation.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/bank/enums/BankLocation.java
@@ -2,6 +2,7 @@ package net.runelite.client.plugins.microbot.util.bank.enums;
 
 import lombok.Getter;
 import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
 import net.runelite.api.Quest;
 import net.runelite.api.QuestState;
 import net.runelite.api.Skill;
@@ -13,8 +14,11 @@ import net.runelite.client.plugins.microbot.Microbot;
 import net.runelite.client.plugins.microbot.util.equipment.Rs2Equipment;
 import net.runelite.client.plugins.microbot.util.player.Rs2Player;
 
+import java.util.Arrays;
+
 @Getter
 @RequiredArgsConstructor
+@Slf4j
 public enum BankLocation {
     ALDARIN(new WorldPoint(1398, 2927, 0), true),
     AL_KHARID(new WorldPoint(3270, 3166, 0), false),
@@ -258,5 +262,14 @@ public enum BankLocation {
             default:
                 return true;
         }
+    }
+
+    private static final BankLocation[] EMPTY_ARRAY = new BankLocation[0];
+
+    public static BankLocation[] values(boolean hasRequirements) {
+        if (!hasRequirements) return values();
+        return Microbot.getClientThread().runOnClientThreadOptional(
+                () -> Arrays.stream(values()).filter(BankLocation::hasRequirements).toArray(BankLocation[]::new)
+        ).orElse(EMPTY_ARRAY);
     }
 }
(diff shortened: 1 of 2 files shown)


Voxsylvae:
Wasn't it already that fast, with my recent changes, based on the caching system? What was the performance before on your Rig?

Kryaex:
> Wasn't it already that fast, with my recent changes, based on the caching system? What was the performance before on your Rig?
image
This is using the fix mango posted since I could not login otherwise

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/bank/enums/BankLocation.java
```diff
@@ -259,4 +263,13 @@ public boolean hasRequirements() {
                 return true;
         }
     }
+
+    private static final BankLocation[] EMPTY_ARRAY = new BankLocation[0];
+
+    public static BankLocation[] values(boolean hasRequirements) {
+        if (!hasRequirements) return values();
+        return Microbot.getClientThread().runOnClientThreadOptional(
```
gmason0, line 271: I dont think running on client thread would be required here as all the requirements above leverage utility classes in-order to call the client thread where needed

gmason0 (commented):
Nice work! would you mind also fixing merge conflicts?

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/bank/enums/BankLocation.java
Kryaex: Running all (atleast non-cached) requirements on the ClientThread is the only reason this improves performance. Otherwise, it would revert back to _wait for ClientThread -> resolve Value -> wait for ClientThread -> ..._ which caused the long wait time in the first place. I'm sure it could be improved by finding out all requirements that aren't cached and batch resolving them but that's too much effort for me

Voxsylvae: This one of the reason why I introduced the caching. After an initial caching run, it should be potentially faster. The only non cached values are the equipment. As far as I remember it was already in similar magnitudes after the caching.🤔 I can't double check it until tomorrow. 
But at some point I also has a second level cache in place here. So it can be, that it only was that fast with the second level cach, which was also caches the equipment.

Kryaex: Well as you can see above, I ran it twice with about 50s between calls. Both times I did not run it on the ClientThread it took about 1500ms. Cache Invalidation between calls should be impossible because the non-client thread one ran directly after my ClientThread check which means it should've been up to date.

Voxsylvae: Screenshot from 2025-07-28 18-29-08
Screenshot from 2025-07-28 19-01-19
Screenshot from 2025-07-28 19-01-47


to follow it up. I had on my branch a similar change like you, which I forgot to push.

```java
Set<BankLocation> accessibleBanks = Microbot.getClientThread().runOnClientThreadOptional(() ->
    allBanks.stream()
        .filter(BankLocation::hasRequirements)
        .collect(Collectors.toSet())
).orElse(Collections.emptySet());
```
I think the change to the BankLocation enum is not necessary.  Just adding the code above to getPathAndBankToNearestBank is enough

Scitur: This does not use caching either tho. Obviously, the precise location of the client thread call can be moved. But this essentially does what my PR does. I put it in the BankLocation class in case it needs to be reused somewhere else.

Scitur: I would be more interested as to why caching fails here. Still, this should be a band-aid fix until someone can debug it.

Voxsylvae: I'm about to look into, to improve the caching here.

PR #1283, opened 2025-07-26
fix(Rs2Bank): Update VarbitCache after withdrawing-x.

zacharyh211:
Otherwise we can have a sequence like:
1. Cached value is 4.
2. Withdraw 40 of something.
3. Attempt to withdraw 4 of something.
4. Cached value is still 4, so use withdraw-x.
5. Withdraws 40 instead.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/bank/Rs2Bank.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/bank/Rs2Bank.java
@@ -30,6 +30,7 @@ import net.runelite.client.plugins.microbot.shortestpath.ShortestPathPlugin;
 import net.runelite.client.plugins.microbot.shortestpath.pathfinder.Pathfinder;
 import net.runelite.client.plugins.microbot.util.antiban.Rs2AntibanSettings;
 import net.runelite.client.plugins.microbot.util.bank.enums.BankLocation;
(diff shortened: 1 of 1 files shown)


chsami:
@Voxsylvae Isn't the varbit updated by the event?

gmason0:
> @Voxsylvae Isn't the varbit updated by the event?

From my understanding yes, this value does get updated in the cache when its changed.

How?

Within the `Rs2VarbitCache.java` class, this class is registered to the event bus to listen for the `onVarbitChanged` - this uses `VarbitUpdateStrategy.java` to update the varbit id's value within the cache.

I will let Vox chime in, but this is my overall understanding of the cache system. We could probably improve the method by pulling the varbit from cache if its available to avoid unnecessary client thread calls

zacharyh211:
Got it, thanks for the explanation! I tried reverting my change and adding some logging to the cache to see what was going wrong before, but now I can't reproduce it.. :)

For context of my specific instance, I was using the Birdhouse Runner plugin and had 40 seeds, 4 logs. The inventory loading step was withdrawing 40 seeds correctly but then tried to withdraw 40 logs. From dumping the local vars, configuredX was still 4 (from the initial state) even though X was now 40.

Thanks for the comments! Closing this pull request.

Voxsylvae:
As George explained, it will be updated by the event. So this shouldn't not happen. So the update should not be necessary.

PR #1288, opened 2025-07-28
Web Walker Changes

gmason0:
### Fixes
- Remove scroll requirement from Ardounge Tablet & Watchtower tablet as this requirement is strictly used for the teleport spell for these locations, not tablets.
- Shilo Village Cart now uses the Pay fare option to streamline getting to Shilo Village & Brimhaven


oversized93:
rejected.

PR #1303, opened 2025-08-01
fix(ArceuusRc): rewrote state machine

Kryaex:
- Rewrote state machine
- Adds REACHED_DISTANCE incase user has their walker reached distance set to something else
- Adds an update message to inform users of new features
- Adds break handler support
- Adds error recovery


gmason0 (commented):
@Kryaex - Is this in a decent state to where we can merge it? I would like to get as many as these closed as possible. 

I understand the other on-going issue with this plugin doesn't appear to be plugin related, but more pathfinder/shortestpath related where SOME users are given the wrong path back to the runestone. But I dont think that is something that needs to be covered here.

Kryaex:
Should be fine, I ran it for like 2hours a couple days ago trying to get it to crash or take the wrong path.

gmason0 (approved):

PR #1306, opened 2025-08-02
new(AnonymousMode): plugin to anonymize your character

Bolado:
### An example of what it does:

https://github.com/user-attachments/assets/e886a162-3c69-4c24-8b5d-7d5df7d4402e

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/anonymous/AnonymousConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/anonymous/AnonymousConfig.java
@@ -1,0 +1,141 @@
+package net.runelite.client.plugins.microbot.anonymous;
+
+import net.runelite.client.config.*;
+
+@ConfigGroup("anonymous")
+@ConfigInformation("<b>-- ANONYMOUS MODE --</b><br /><br />" +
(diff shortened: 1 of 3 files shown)


gmason0 (commented):
Great work on this, sorta neat!

Just curious - how does this affect utility classes such as Rs2Bank, Rs2Player's skill requirement methods, things of that nature - or is this strictly visual ?

Bolado:
> Great work on this, sorta neat!
> 
> Just curious - how does this affect utility classes such as Rs2Bank, Rs2Player's skill requirement methods, things of that nature - or is this strictly visual ?

I hope people like it 😄

Before I started this one I tried to find heavy dependencies of widget's text values, and I couldn't find anything that would really be impacted, although due my limited knowledge about Java language I couldn't find for example the logic behind functions inside the Widget interface, so I couldn't really say if in there something depends heavily on text field or instead use varbits or memory values.

The plugin itself is very simple, it just replace the text field value of the specified widget, so if the functions inside important interfaces like Item and Widget isn't dependent on such simple value, I think it's fine, but just in case I made everything optional, so if the user disables the masking of item quantities, on the first widget refresh call the changed text fields goes back to what it should look like initially.

gmason0 (approved):

PR #1308, opened 2025-08-02
feat(AutoWoodcutting): Add infected root and don't drop demon tears while woodcutting

ghost:
**NOTE**: As of right now, the Infected Root ObjectID is not in the gameval/ObjectID class from RuneLite, so the tree game object can not be found yet. This probably should be merged _after_ RuneLite adds the ObjectID into the class.

This PR adds the new infected root woodcutting method which is afk and profitable. It also adds demon tears to the items that should not be dropped when emptying the inventory for the method to make sense.

More info on the new method: https://oldschool.runescape.wiki/w/Infected_root

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/gameobject/ObjectID.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/gameobject/ObjectID.java
@@ -12,6 +12,7 @@ public class ObjectID {
     public static final int YEW_TREE_GE = 10823;
     public static final int DENSE_RUNESTONE_1 = 10796;
     public static final int DENSE_RUNESTONE_2 = 8981;
+	public static final int INFECTED_ROOT = 56780;
 
 	/*
 	 * Abyss objects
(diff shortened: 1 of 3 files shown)


gmason0 (changes requested):
If you have the object id manually - we also have an object id class for these situations where the runelite API doesn't have the constants defined. 

https://github.com/chsami/Microbot/blob/main/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/gameobject/ObjectID.java

ghost:
> If you have the object id manually - we also have an object id class for these situations where the runelite API doesn't have the constants defined.
> 
> https://github.com/chsami/Microbot/blob/main/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/gameobject/ObjectID.java

Oh, that's good to know! Any pointers on how I get the object id then? I normally always got them through the RuneLite Dev Tools -> Game Objects, but these Objects don't show any ID or similar. Thanks :)

gmason0:
Under the Microbot Plugin - you can enable `Menu Entry Logging` - this will start to log each menu entry option that is clicked

Optionally, you can enable the Game Chat Logs - this could be helpful if you do not want to dig through the client.log file or the console of the application

Once doing this, interact with the object once & then look for the `getIdentifier` within the menu entry, this is your object id. 

Simply add the constant as a friendly name, then search based on that name e.g

```
// In our ObjectId static class
public static final int INFESTED_ROOT_1 = 123
public static final int INFESTED_ROOT_2 = 345

// Retrieving the Game Object 
Rs2GameObject.getGameObject("Infested root"); 
```

ghost:
I have added the object id to the microbot ObjectID class! Thanks for the pointer @g-mason0 :)

gmason0 (approved):

PR #1316, opened 2025-08-04
feat(Logging): log java & microbot version

Kryaex:


gmason0 (approved):
Thanks for this! I was thinking of something like this - I think it would be nice to include commit SHA as well when its generated via Github action to better debug when someone is using the nightly version.

Kryaex:
Yeah, that sounds like a good idea. I just saw you were getting ready to update to the next runelite version and wanted to add something quickly.

PR #1337, opened 2025-08-09
feat(agility): Add efficient alching mode with improved obstacle completion detection

runsonmypc:
This PR introduces an efficient alching mode for the agility plugin along with improved obstacle completion detection to prevent mid-obstacle interactions, and introduces XP Drops as an early signal for obstacle completion.

## Features Added:

- **Efficient Alching Mode**: New config option that changes alching behavior for obstacles 5+ tiles away
  - When enabled: Click obstacle first → Alch during movement → Click obstacle again
  - Maximizes efficiency by utilizing movement time for alching
- **Skip Inefficient Option**: New config to only alch when obstacles are 5+ tiles away
  - Prevents suboptimal alching on close obstacles to maintain maximum XP rates
- **XP Drop Detection**: Uses agility XP drops to instantly detect obstacle completion
  - Allows immediate action on next obstacle without waiting for animations to finish
  - Significantly faster reactions compared to movement-based detection alone for some obstacles
- **Alch Skip Chance**: Configurable percentage chance (0-100%) to randomly skip alching on any obstacle
  - Applies to all obstacles and all alching modes

## Bug Fixes:

- **Mid-Obstacle Interaction Prevention**: Added XP-based completion detection and timing delays to prevent mid-obstacle clicking during animation pauses (e.g., obstacles that have a small pause before the character jumps)
- **Colossal Wyrm Support**: Disabled XP checks for Colossal Wyrm courses which have obstacles with multiple XP drops
- **Rs2Walker Alching Fix**: Prevents alching immediately after Rs2Walker returns to course start

## Technical Improvements:

- Added XP tracking to detect obstacle completion immediately upon XP gain for faster XP rates
- Special handling for first obstacle (index 0) to prevent issues after lap completion
- **Refactored obstacle completion detection** into course-specific `isObstacleComplete()` method
- **Reduced code nesting** by extracting alching logic into dedicated helper methods
- **Improved code maintainability** through better separation of concerns

## Testing Notes:

Please let me know if you know of any other courses that have obstacles with XP drops before they end

## Summary by CodeRabbit

- New Features
  - Added configurable alching options: efficient alching, skip inefficient alchs (nearby obstacles), and per-obstacle skip chance (0–100%).
  - Integrated alching into the agility loop with distance-aware and randomized skip behavior.

- Refactor
  - Introduced XP-aware gating for obstacle clicks to avoid premature interactions.
  - Replaced blocking waits with polling timeouts and improved movement/animation/plane/health detection.
  - Consolidated course-specific handling for more reliable runs across courses (Colossal Wyrm, Prifddinas, Werewolf).


runelite-client/src/main/java/net/runelite/client/plugins/microbot/agility/AgilityScript.java
```diff
@@ -96,10 +101,24 @@ public boolean run()
 					return;
 				}
 
+				// Check if we're still completing an obstacle
 				if (plugin.getCourseHandler().getCurrentObstacleIndex() > 0)
 				{
-					if (Rs2Player.isMoving() || Rs2Player.isAnimating())
-					{
+					// Disable XP checks for Colossal Wyrm courses (they have multi-XP drop obstacles)
+					boolean isColossalWyrm = config.agilityCourse().name().contains("COLOSSAL_WYRM");
```
gmason0: I think the real issue here is the waitForCompletion method from the CourseHandler interface.

This method should be holding this future until the obstacle is completed.


We can create exceptions for specfic courses by overriding this method in the course's class, or globally change it by changing it in the interface.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/agility/AgilityScript.java
```diff
@@ -170,11 +184,69 @@ else if (!(plugin.getCourseHandler() instanceof GnomeStrongholdCourse))
 					Rs2Walker.walkMiniMap(gameObject.getWorldLocation());
 				}
 
-				if (Rs2GameObject.interact(gameObject))
-				{
-					plugin.getCourseHandler().waitForCompletion(agilityExp, Microbot.getClient().getLocalPlayer().getWorldLocation().getPlane());
+				// Handle alchemy if enabled
+				if (config.alchemy()) {
```
gmason0: Can this be refactored to not be so, nested? I think we could create a method on the interface & override it if there is going to be exceptions for specfic courses.

gmason0 (changes requested):

runsonmypc:
@g-mason0 Thanks for the review! I've addressed both of your concerns:

### Changes Made:

1. **Refactored `waitForCompletion` method**:
   - Created `isObstacleComplete()` method in AgilityCourseHandler interface
   - Courses can now override this for custom completion detection
   - Colossal Wyrm courses override it to handle multi-XP drop obstacles

2. **Reduced code nesting**:
   - Extracted alching logic into `performEfficientAlch()` and `performNormalAlch()` helper methods
   - Created `handleCourseSpecificActions()` to consolidate course-specific handling
   - Moved completion checks into the new interface method

runelite-client/src/main/java/net/runelite/client/plugins/microbot/agility/AgilityScript.java
```diff
@@ -97,10 +102,26 @@ public boolean run()
 					return;
 				}
 
+				// Check if we're still completing an obstacle
 				if (plugin.getCourseHandler().getCurrentObstacleIndex() > 0)
 				{
-					if (Rs2Player.isMoving() || Rs2Player.isAnimating())
+					boolean obstacleComplete = plugin.getCourseHandler().isObstacleComplete(
```
gmason0: Shouldn't the functionality be at the end of the loop? seems like you wanted to replace waitForCompletion.

gmason0 (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/agility/AgilityScript.java
runsonmypc: @g-mason0 Thanks for the feedback! You're right - the logic was scattered. I've refactored to centralize everything in the course handler:

- Added shouldClickObstacle() method that handles the "don't click while animating unless XP gained" logic
- Kept waitForCompletion() for post-click blocking
- Removed duplicate XP tracking and animation checks from the main script

Now the handler controls all timing decisions. The script just asks "should I click?" and the handler decides based on animation/XP state. Much cleaner than having these checks spread across multiple places.

The XP drop detection for early action still works - if we get XP while animating, shouldClickObstacle() returns true and we can click the next obstacle immediately.

runsonmypc:
Addressed coderabbit comments.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/agility/courses/ColossalWyrmBasicCourse.java
```diff
@@ -1,35 +1,100 @@
...
+				return true;
+			}
+			
+			// Sleep before next check
+			Global.sleep(100);
+		}
+		
+		// Timeout reached
+		return false;
+	}
```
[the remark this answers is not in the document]
runsonmypc: Further changes to other courses are out of scope.

gmason0 (approved):
there is still some heavy nesting here, but otherwise this seems to be alot better than before.

We can move forward with this - thank for your being on top of this despite the long review period.

PR #1340, opened 2025-08-10
fix(aiofighter): Fix 'Eat food for space' feature in AIO Fighter

runsonmypc:
## Summary
Fixes the broken "Eat food for space" feature in the AIO Fighter plugin that was not working as intended.

## Problem
The "Eat food for space" setting was supposed to eat food when inventory space dropped below the configured minimum free slots threshold, preventing unnecessary banking trips. However, it was completely
broken due to:

1. **Incorrect execution order** - The eating logic was inside the banking check, so it only ran after already deciding to bank
2. **Wrong trigger condition** - It only activated when inventory was completely full rather than at `Min. free slots`
3. **No coordination between scripts** - Banking could trigger before the eat food logic had a chance to run
4. **Didn't work with banking disabled** - Feature only worked when banking was enabled

## Solution
- Created a dedicated `EatForSpaceScript` that runs independently every second to handle eating food for space
- Added coordination logic to `BankerScript.needBanking()` to prevent premature banking when food is available to eat
- Fixed the threshold logic to:
  - When banking enabled: eat at `minFreeSlots` threshold
  - When banking disabled: eat when inventory is full
- Ensures proper execution order - banking decisions now check if food can be eaten first
- Added debug logging to track when the feature activates

## Testing
- ✅ Plugin now correctly eats food when inventory slots drop to or below the configured threshold
- ✅ Works with both banking enabled and disabled modes
- ✅ Banking is deferred when food is available to eat for space
- ✅ Feature properly prevents unnecessary banking trips when food is available

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/aiofighter/AIOFighterPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/aiofighter/AIOFighterPlugin.java
@@ -20,6 +20,7 @@ import net.runelite.client.plugins.microbot.Microbot;
 import net.runelite.client.plugins.microbot.aiofighter.bank.BankerScript;
 import net.runelite.client.plugins.microbot.aiofighter.cannon.CannonScript;
 import net.runelite.client.plugins.microbot.aiofighter.combat.*;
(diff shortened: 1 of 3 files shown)


gmason0 (commented):
@See1Duck - What are your thoughts around this feature?

gmason0 (approved):
I think this is fine

PR #1346, opened 2025-08-10
region filter added for Autologin and BreakHandler! img/vid included

ghost:
breakworld

world

region option filter added for Autologin /BreakHandler-

UK
US
GER
AUS

Tested, it works for all switches :)

https://github.com/user-attachments/assets/f23201eb-f883-4fa6-8469-937166512881


chsami:
Thanks for the contribution!

It seems you forked of from main instead of development. Make sure to fork from development and get the latest development and then make a PR, this will avoid having branch conflicts

ghost:
@chsami  Will just do another commit for autologin for now then as lots of people are working on break handler still

PR #1361, opened 2025-08-13
Fix karambwan script

beta-ms:
Improved the karambwan script. When done fishing it uses the fairy ring correctly, stores clues and uses micro brakes. Also some code cleanup.


gmason0 (approved):
Im surpised this script hasn't died yet considering AutoFishing exists. 

Looks good to me though.

PR #1364, opened 2025-08-14
TEMPOROSS CHANGES - SOLO  (mostly) - Will work with all harpoons/bare-hand

T-edit:
### **_TEMPOROSS CHANGES - SOLO_** - Infernal Harpoon not needed no more!!! Most changes are for solo mode only, equipment/area changes works for all.

Why solo only? because it more xp/hr and points than mass. I tested Mass, still works as it did. (Solo a lot better)

Solo is very RNG dependent on catch rate of double/normal fish and cloud/fire spawn timer. Higher level/better harpoon will help points as less Emergency_catch/fill mode, the better the points.
19 free slots minimum. Will work with extra free slots! so more points! Recommended GPU on so web walker does not have to first calculate path. 

I tested on 2 accounts lv80+ fishing with barbtail harpoon/bare-hand fishing and had 90%-99% win rate. 

Updated all Old Id values to use gameval IDs

Config description updated

**Overlay changes**
- All overlays have toggle option to enable/Disable
- Overlay opacity reduced to be more pleasant
- The state transition logic was moved from TemporossProgressionOverlay to TemporossScript, allowing the script to function correctly when the overlay is hidden.
- Added Stat Tracking Overlay - Tracks: Games/Win/Lose/Reward Permits gained/Reward p/Hr/Total fishing xp gained/Fishing xp/h/runtime
- Added new dock areas to the work area overlay.

**Equipment support**

- Added support for all Variants of Harpoon to be detected under main harpoon.
- Imcando Hammer(Off-hand) support added
- 6 Buckets minimum recommended for Solo mode (default set)

**Change and improved how STATE is handled Solo config overall.** 

- Added new EMERGENCY_CATCH state for SOLO mode and new EnergyStateManager.jar to handle EMERGENCY_CATCH. It essentially keeps Tempoross Energy at 10% or under by catching the exact amount of fish needed asap. When Tempoross Energy is 10% or low he can not use special attacks. This state is active only after Initial_fill state has been finished. EMERGENCY_CATCH will always activated if at THIRD_CATCH state and Tempoross energy >= 11%. 
Set threshold of raw harpoon fish to catch on EMERGENCY_CATCH is set in EnergyStateManager which will grab the fish values to get tempoross to 10% energy  or lower.

- Edited SECOND_FILL to have it own solo method to use getAllFish() == 0 instead of cooked as due to emergency storm intensity of >=93 and slow cooking it sometimes won't have enough fish to send all cooked fishes so better to get points from raw aswell. All fishes send all cooked and raw fishes to ammobox. previous code would only dump cooked fishes.

THIRD_CATCH state has option to go to SECOND_FILL when Tempoross energy is  92% or more - added this as if player gets bad catch RNG/low lvl fishing, this enables it to fill without needing to go cook state. Very rarely needed but I have noticed it does happen. Prevents you from losing game before first attack tempoross state.

Initial_fill state changes for solo - This is the most important phase for beating Tempoross in solo mode
These changes help with the catch RNG of INITIAL_FILL state , so with faster harpoon/catch rate it will still be optimal and quick, however safety measures is added for lower level fishing, to still handle the fire before filling if not quick enough.

- Implement Cloud dodge if filling ammo, if cloud is on same tile with player when filling on ammocrate it will switch to next ammocrate.
- Script will focus on filling ammo if already started filling ammocrate before clouds appears, if player has 9 or less fish in inventory it will fill and then handle fire.
- if cloud has spawn before player has started filling ammocrate, it will now wait at a safe distance to handle fire. This is safety measure to prevent fire from ruining the game by accidently walking into it/prevents running to fill ammocrate when fire is about to spawn. 
- At lower level fishing/bad catch rate RNG then the clouds will spawn as you finish cooking or before, when the script goes to INITIAL_FILL state it will sometimes run through fire, this change is a prevention measure for slow catch rate to handlefire first

**Enhance Wave handling logic**

- Modified Totem/Mast detection to be quicker. Prioritized repairing totem when player is nearby now.
- Added safety measures for when player would get hit by wave to recover back to previous state.
- Adjusted script to get buckets in solo mode, When buckets is less or equal to 2 , it will now grab 3 more , safety measure in solo mode to fight fire, rarely happens.

**WorkArea changes**

- Moved safepoint up 1 tile and added separate safepoint to mirror west side safe point as well. This allows the script to detect it easier as it will be 1 tile closer
- Added Dock water pump and bucket to WorkArea. Also added them to be supported in script. Script will now grab/fill buckets at dock area when fire is blocking path, prevents forfeits when its not needed.
- Added new dock areas to the workarea overlay.

**Finishing game ** 

- when Tempoross is defeated, to also grab bucket and fill up water to match the config amount, previously this only filled up bucket in inventory. This enables faster restarts without having to get config buckets at start of minigame. 

**Fire Handling**

- reduced handlefires interact speed to deal with fire quicker
- added prevention measures to prevent bot from fighting fires when it spreads near the end phase as this happened sometimes

**Camera angles**

- Set camera zoom/angles so bot can click interact with certain areas easier. Every time it has to spin the camera it loses a tick. Very important in initial phase.

**ATTACK_TEMPOROSS state**

- Updated logic to interact faster with spirit pool after SECOND_FILL state ends. Noticed it stalled/lagged sometimes causing tempoross to enrage and lose the game(rarely before).

- RS2 mouse hover method added over Spirit pool to allow faster transitions

- If Tempoross Energy is above 26% before last Attack phase, script will only catch 19 fish, if he has less than 26% , the script will catch all available inventory slots. This is to ensure the boss gets killed in time

popopopo

overlaystats


gmason0 (commented):
@See1Duck - I am going to need you to review this if possible. There is just too much changed here for me to feel comfortable reviewing.

PR #1365, opened 2025-08-14
fix(birdhouse): use non-blocking error messages and improve banking

runsonmypc:
## Summary
This PR fixes client freezing issues caused by error popup dialogs and improves the banking workflow for the Birdhouse Runner plugin.

## Changes
- **Fixed error handling**: Replaced `reportFinished(false)` calls with `Microbot.showMessage()` to avoid freezing the client
- **Always empty bird nests**: Bird nests are now emptied regardless of the "Return to bank" configuration setting
- **Improved bank location**: Changed from using nearest bank to specifically using the Fossil Island shipwreck bank (north of the island) for more convenience (chaining seaweed runs)

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/birdhouseruns/FornBirdhouseRunsScript.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/birdhouseruns/FornBirdhouseRunsScript.java
@@ -344,7 +344,7 @@ public class FornBirdhouseRunsScript extends Script {
         
         // If we get here, no seed type had 40+ available
         setupErrorMessage = "Need 40 seeds but no birdhouse seed type has 40+ in bank";
-        Microbot.log(setupErrorMessage);
+        log.error(setupErrorMessage);
         return false;
     }
 }
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/birdhouseruns/FornBirdhouseRunsScript.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/birdhouseruns/FornBirdhouseRunsScript.java
@@ -1,5 +1,6 @@
 package net.runelite.client.plugins.microbot.birdhouseruns;
 
+import lombok.extern.slf4j.Slf4j;
 import net.runelite.api.gameval.ItemID;
 import net.runelite.api.Quest;
 import net.runelite.api.QuestState;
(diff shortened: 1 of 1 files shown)


Voxsylvae:
Please revert the changes you made for: 
- 'Fixed error handling: Replaced reportFinished(false) calls with Microbot.showMessage() to avoid freezing the client.'
These changes would lead to the plugin no longer being schedulable. The function reports back to the scheduler that the plugin has finished its main task or failed to finish its main task, so the PluginScheduler can stop the plugin and know it can schedule another plugin.

runsonmypc:
> Please revert the changes you made for:
> 
>     * 'Fixed error handling: Replaced reportFinished(false) calls with Microbot.showMessage() to avoid freezing the client.'
>       These changes would lead to the plugin no longer being schedulable. The function reports back to the scheduler that the plugin has finished its main task or failed to finish its main task, so the PluginScheduler can stop the plugin and know it can schedule another plugin.

Done!

gmason0 (approved):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/birdhouseruns/FornBirdhouseRunsScript.java
```diff
@@ -55,7 +58,8 @@ public boolean run() {
                 if (!initialized) {
                     if (Rs2Player.getQuestState(Quest.BONE_VOYAGE) != QuestState.FINISHED) {
                         Microbot.log("You need to finish the quest 'BONE VOYAGE' to use this script");
-                        plugin.reportFinished("Birdhouse run failed, you need to finish the quest 'BONE VOYAGE'",false);
+                        log.error("Birdhouse run failed, you need to finish the quest 'BONE VOYAGE'");
```
gmason0, line 60: This is redundant logging. Microbot.log -> Slf4j logging. we should be using one or the other.

gmason0 (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/birdhouseruns/FornBirdhouseRunsScript.java
runsonmypc: Sorted.

gmason0: There are still multiple areas of this script that use microbot#log - considering you are adding the sl4j logging in this pull request - either we remove that or replace all instances of microbot#log to use sl4j logging.

runsonmypc: Oops, my bad! That should be all of them now.

gmason0 (approved):

PR #1368, opened 2025-08-14
Adds Piety Option to NMZ

Badn1nja:
also renames     
boolean togglePrayerPotions() to boolean togglePrayProtect() to reflect it's function better

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/nmz/NmzScript.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/nmz/NmzScript.java
@@ -54,7 +54,7 @@ public class NmzScript extends Script {
 
     public boolean canStartNmz() {
         return Rs2Inventory.count("overload (4)") == config.overloadPotionAmount() ||
-                (Rs2Inventory.hasItem("prayer potion") && (config.togglePrayProtect() || config.togglePrayPiety()));
+                (Rs2Inventory.hasItem("prayer potion") && (config.togglePrayProtect() || config.toggleOffensivePrayer()));
     }
 
     @Inject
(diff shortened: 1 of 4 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/microbot/nmz/NmzScript.java
```diff
@@ -118,8 +117,10 @@ public boolean isOutside() {
 
     public void handleOutsideNmz() {
         boolean hasStartedDream = Microbot.getVarbitValue(3946) > 0;
-        if (config.togglePrayerPotions())
+        if (config.togglePrayProtect())
             Rs2Prayer.toggle(Rs2PrayerEnum.PROTECT_MELEE, false);
+        if (config.togglePrayPiety())
```
SteffenCarlsen, line 122: Would probably make more sense to use `Rs2Prayer.getBestMeleePrayer()` over hardcoding piety. We could add the same support for range/magic.

SteffenCarlsen: You can use the `WeaponId` enum to get the attackstyle of your currently equipped weapon

SteffenCarlsen (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/nmz/NmzScript.java
Badn1nja: I'll look into it, I just made this PR as the work was done already (It didn't work out for me but nearer max it should be better points per hour)

gmason0 (changes requested):
1. Rebase this to be against the development branch 
2. I think we should consider just overall attack style & use the offensive prayer for that style

PR #1373, opened 2025-08-15
fix(motherlodemine) deposit pay-dirt and inventory when possible

Krulvis:
- Handle infernal pickaxe
 - click crate with 100% hammer
 - drop paydirt when hammer is needed but inventory is full
 - deposit paydirt when possible
 - deposit ore when sack is empty and inventory is not full


runelite-client/src/main/java/net/runelite/client/plugins/microbot/mining/motherloadmine/MotherloadMineScript.java
```diff
@@ -622,27 +634,22 @@ private boolean obtainHammer() {
...
-				plugin.getBlacklistedCrates().add(crate.getWorldLocation());
-				Rs2Inventory.drop("bronze pickaxe");
-				Rs2Inventory.waitForInventoryChanges(5_000);
-			} else {
-				pickedUpHammer = true;
-				break;
-			}
+        while (!Rs2Inventory.hasItem("hammer") && isRunning()) {
+            //The crate at this point ALWAYS gives the player a hammer
+            Rs2GameObject.interact(new WorldPoint(3752, 5674, 0), "Search");
```
gmason0, line 647: This solution ignores that you can get bronze pickaxes from the crate, it is not ALWAYS certain that you will get a hammer. If you look at the previously implementation - this would add the worldpoint for the crate into a collection within the plugin that would clear when you change worlds or logged out. 

These locations appears to be static, not change unless you change worlds 

[This is how I found this information](https://oldschool.runescape.wiki/w/Motherlode_Mine#:~:text=Hammers%20can%20be%20found%20in%20three%20of%20the%20crates%20surrounding%20the%20machine.%20The%20other%20three%20crates%20have%20bronze%20pickaxes.), based on this information, it appears that the three crates will randomly have bronze pickaxe & three crates will randomly have a hammer.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/mining/motherloadmine/MotherloadMineScript.java
```diff
@@ -622,27 +634,22 @@ private boolean obtainHammer() {
...
-			Optional<GameObject> closestCrate = Rs2GameObject.pickClosest(crates, GameObject::getWorldLocation, Microbot.getClient().getLocalPlayer().getWorldLocation());
-
-			if (closestCrate.isEmpty()) {
-				log.error("Unable to find any crates to search for a hammer.");
-				break;
-			}
-
-			GameObject crate = closestCrate.get();
+        if (Rs2Inventory.isFull()) {
+            if (Rs2Inventory.interact("pay-dirt", "drop")) {
```
gmason0, line 638: What if there are multiple pieces of paydirt, I think we should just use Rs2Inventory#dropAll

gmason0 (changes requested):
Everything else looks good

runelite-client/src/main/java/net/runelite/client/plugins/microbot/mining/motherloadmine/MotherloadMineScript.java
Krulvis: It is always certain you get a hammer from this crate. I have used that crate and only that crate in my script since forever and have never gotten a bronze pickaxe from it

runelite-client/src/main/java/net/runelite/client/plugins/microbot/mining/motherloadmine/MotherloadMineScript.java
Krulvis: You only need to drop one to pickup a hammer

gmason0: I see, then we eventually end up back in this spot. when emptying the sack to deposit the remaining!

runelite-client/src/main/java/net/runelite/client/plugins/microbot/mining/motherloadmine/MotherloadMineScript.java
gmason0: Seems legit, If it becomes an issue, i'll know where to look - but this was just my interpretation of the wiki

gmason0 (approved):

PR #1390, opened 2025-08-19
Development

gmason0:
### Chores
- Remove Thieving plugin
- Remove All STTS plugins & PluginDescriptor
- Bump to 1.9.8.4


Some0ne396:
why is thieving plugin removed ?

PR #1402, opened 2025-08-19
feat(AIOFighter): Add wait-for-loot feature

runsonmypc:
## Summary
  This PR adds a configurable wait-for-loot feature to AIOFighter that pauses combat after killing an NPC to allow loot to appear before attacking the next target (useful for NPCs with long death animations such as dragons).

  ## Features
  - **Configurable toggle**: Enable/disable via "Wait for Loot" option in the Loot section
  - **Configurable timeout**: Adjustable wait time from 1-10 seconds (default 6) via "Loot Wait Timeout" option
  - **Smart detection**: Continuously monitors NPC health during combat for reliable death detection
  - **Loot prioritization**: Prevents new combat from interrupting loot collection, even when attacked

  ## Implementation Details
  1. **Death detection**:
     - Continuously tracks the NPC being fought throughout combat
     - Detects death when health reaches 0 or isDead() returns true
     - Avoids false positives from eating, walking, or running out of ammo

  2. **Loot coordination**:
     - Respects `pauseAllScripts` flag to prevent interrupting active looting
     - Clears wait state when loot is found or timeout expires
     - Prevents target switching while waiting to ensure loot priority
     - Status display shows countdown timer while waiting

  3. **Files changed**:
     - `AIOFighterConfig.java`: Added toggleWaitForLoot and configurable timeout options
     - `AIOFighterPlugin.java`: Added wait state management variables
     - `AttackNpcScript.java`: Implemented continuous NPC tracking and death detection
     - `LootScript.java`: Added wait state clearing when loot is found

  ## Use Case
  Essential for NPCs with:
  - Long death animations (e.g., dragons take 3-4 seconds)
  - Delayed loot drops
  - Valuable loot that justifies waiting

## Summary by CodeRabbit

* **New Features**
  * Option to wait for loot to appear before attacking the next NPC.
  * Configurable loot wait timeout (1–10 seconds).

* **Improvements**
  * Combat pauses while waiting for loot and resumes on timeout or after first pickup.
  * Looting more robust: per-item pickup attempts, continue on failures, and ensure scripts resume.
  * Optional automatic food consumption when inventory is full.
  * State resets and clearer status/log messages across lifecycle.


runelite-client/src/main/java/net/runelite/client/plugins/microbot/aiofighter/AIOFighterPlugin.java
```diff
@@ -69,6 +69,15 @@ public class AIOFighterPlugin extends Plugin {
     @Getter
     @Setter
     public static int cooldown = 0;
+    
+    @Getter @Setter
+    private static long lastNpcKilledTime = 0;
+    
+    @Getter @Setter
+    private static boolean waitingForLoot = false;
```
Krulvis: Why is there a waiting For Loot bool and a timer that you keep track of? It seems like just the time since kill should be enough to determine whether you're waiting for loot or not

runelite-client/src/main/java/net/runelite/client/plugins/microbot/aiofighter/combat/AttackNpcScript.java
```diff
@@ -99,11 +99,51 @@ public void run(AIOFighterConfig config) {
...
+                    return; // Don't attack while looting
+                }
+                
+                // Check if our current target just died and we should wait for loot
+                if (config.toggleWaitForLoot() && !AIOFighterPlugin.isWaitingForLoot()) {
+                    // Check if we were recently in combat but no longer interacting (NPC just died)
+                    Actor currentInteracting = Rs2Player.getInteracting();
+                    
+                    // If we're not interacting but were recently, the NPC probably just died
+                    if (currentInteracting == null && Rs2Player.isInCombat()) {
```
Krulvis: This is a check that can trigger for a lot of reasons and should not be used to say that the npc died and thus start a waitforLoot timer.
You lose interacting when you're eating, walking, out of arrows. It generally doesn't work well with anyone that  has auto retaliate off

runelite-client/src/main/java/net/runelite/client/plugins/microbot/aiofighter/AIOFighterPlugin.java
```diff
@@ -429,6 +438,19 @@ public void onNpcDespawned(NpcDespawned npcDespawned) {
         try {
             if(config.togglePrayer())
                 flickerScript.onNpcDespawned(npcDespawned);
+            
+            // Handle wait for loot feature
+            if (config.toggleWaitForLoot()) {
+                NPC npc = npcDespawned.getNpc();
+                if (npc != null && npc.isDead()) {
```
Krulvis: I feel like this PR can be reduced to:
- cache npc were fighting
- check if it dies through the NpcDespawned event
- trigger a delay for the AttackNpcScript
- end timeout early when loot is spawned

Looting script priority should automatically make sure that loot is picked up when the AttackNpcScript is delayed while no loot is spawned.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/aiofighter/AIOFighterPlugin.java
```diff
@@ -69,6 +69,15 @@ public class AIOFighterPlugin extends Plugin {
...
     @Setter
     public static int cooldown = 0;
+    
+    @Getter @Setter
+    private static long lastNpcKilledTime = 0;
+    
+    @Getter @Setter
+    private static boolean waitingForLoot = false;
+    
+    public static final int LOOT_WAIT_TIMEOUT = 6000; // 6 seconds
```
Krulvis: 6 seconds to wait for loot seems like a lot. Doesn't loot spawn immediately after the Npc is de spawned?

runsonmypc: It's a timeout in case loot does **not** spawn. This is intended for use with NPCs with long death animations as per the PR desc

runelite-client/src/main/java/net/runelite/client/plugins/microbot/aiofighter/combat/AttackNpcScript.java
runsonmypc: You're right this is a mistake, should just check if NPC hp hits 0.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/aiofighter/AIOFighterPlugin.java
runsonmypc: The problem is that the NPC takes too long to despawn, and in that time the Fighter will go attack the next NPC.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/aiofighter/AIOFighterPlugin.java
runsonmypc: This is to avoid the Fighter interrupting the loot.

Krulvis: Yes, but you don't need both.
waitingForLoot can just be a function that checks if lastNpcKilledTime is within LOOT_WAIT_TIME

runelite-client/src/main/java/net/runelite/client/plugins/microbot/aiofighter/AIOFighterPlugin.java
Krulvis: > The problem is that the NPC takes too long to despawn, and in that time the Fighter will go attack the next NPC.

Out of curiosity, which npcs take 6 seconds to spawn loot?

runsonmypc: It will stop waiting as soon as the loot spawns, 6 seconds is just a safeguard. Dragons take like 3 seconds idk, have not measured.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/aiofighter/AIOFighterPlugin.java
Krulvis: Let's say you're fighting an NPC that doesn't have a guaranteed drop (no bones or ashes). In this scenario you'll find yourself waiting 6 seconds whenever nothing spawns.

runsonmypc: Yes.

runsonmypc: If you can figure out the longest delay for something to die/loot to spawn I will set it to that. 6 seconds is to avoid missing edge cases. I did not foresee this being used on anything that does not have guaranteed drops (it is mainly intended for dragons and the like).

runsonmypc:
Thanks for the feedback! I've addressed the concerns raised:

####  @Krulvis's suggestions addressed:

  1. "Just use timer, not boolean" - I kept both as they serve different purposes. The boolean provides clear state management while the timer tracks duration. Using only timestamps would require constant calculations throughout the code, making
  it less readable and more error-prone.
  2. Death detection improvements - Removed the problematic "lost target while in combat" check that was triggering falsely on eating/walking/running out of ammo. Now using continuous NPC tracking during combat for reliable detection.
  3. "6 seconds too long?" - Made the timeout configurable (1-10 seconds, default 6). Dragons need 3-4 seconds for death animation plus network delays, so 6 is a reasonable default that users can adjust.
  4. NpcDespawned not useful - Removed this approach entirely as it fires too late for NPCs with long death animations.

####  Key improvements:
  - Death detection now continuously tracks the NPC during combat rather than checking only on interaction loss
  - Added configurable timeout so users can adjust based on the NPCs they're fighting
  - Fixed issue where getting attacked while waiting would interrupt loot collection - now properly prioritizes looting
  - Cleaned up redundant code and simplified the overall implementation

runelite-client/src/main/java/net/runelite/client/plugins/microbot/aiofighter/combat/AttackNpcScript.java
```diff
@@ -103,6 +104,51 @@ public void run(AIOFighterConfig config) {
...
+                // Check if we should pause while looting is happening
+                if (Microbot.pauseAllScripts.get()) {
+                    return; // Don't attack while looting
+                }
+
+                // Check if we need to update our cached target (but not while waiting for loot)
+                if (!AIOFighterPlugin.isWaitingForLoot()) {
+                    Actor currentInteracting = Rs2Player.getInteracting();
+                    if (currentInteracting instanceof net.runelite.api.NPC) {
+                        net.runelite.api.NPC npc = (net.runelite.api.NPC) currentInteracting;
```
gmason0: Instead of using the NPC object, we should be comparing using Rs2NpcModel. You can compare an NPC & an Rs2NpcModel by using the getIndex() method on each of the objects.

gmason0 (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/aiofighter/combat/AttackNpcScript.java
runsonmypc: Done!

runelite-client/src/main/java/net/runelite/client/plugins/microbot/aiofighter/combat/AttackNpcScript.java
```diff
@@ -45,6 +45,7 @@ public class AttackNpcScript extends Script {
     public static Actor currentNpc = null;
     public static AtomicReference<List<Rs2NpcModel>> filteredAttackableNpcs = new AtomicReference<>(new ArrayList<>());
     public static Rs2WorldArea attackableArea = null;
+    public static int cachedTargetNpcIndex = -1;
```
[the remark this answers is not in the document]
runsonmypc: Addressed these comments.

runsonmypc:
Addressed the new coderabbit comments

runelite-client/src/main/java/net/runelite/client/plugins/microbot/aiofighter/loot/LootScript.java
```diff
@@ -58,26 +58,37 @@ public boolean run(AIOFighterConfig config) {
                 if (config.toggleDelayedLooting()) {
                     groundItems.sort(Comparator.comparingInt(Rs2GroundItem::calculateDespawnTime));
                 }
-                //Pause other scripts before looting
+                // Defer clearing wait-for-loot until we successfully pick at least one item
+                //Pause other scripts before looting and always release
                 Microbot.pauseAllScripts.getAndSet(true);
```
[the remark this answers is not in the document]
runsonmypc: Addressed.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/aiofighter/combat/AttackNpcScript.java
```diff
@@ -1,206 +1,255 @@
...
+                attackableArea = new Rs2WorldArea(config.centerLocation().toWorldArea());
+                attackableArea = attackableArea.offset(config.attackRadius());
+                List<String> npcsToAttack = Arrays.stream(config.attackableNpcs().split(","))
+                        .map(x -> x.trim().toLowerCase())
+                        .collect(Collectors.toList());
+                filteredAttackableNpcs.set(
+                        Rs2Npc.getAttackableNpcs(config.attackReachableNpcs())
+                                .filter(npc -> npc.getWorldLocation().distanceTo(config.centerLocation()) <= config.attackRadius())
+                                .filter(npc -> npc.getName() != null && !npcsToAttack.isEmpty() && npcsToAttack.stream().anyMatch(npc.getName()::equalsIgnoreCase))
+                                .sorted(Comparator.comparingInt((Rs2NpcModel npc) -> npc.getInteracting() == Microbot.getClient().getLocalPlayer() ? 0 : 1)
```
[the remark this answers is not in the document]
runsonmypc: out of scope

runelite-client/src/main/java/net/runelite/client/plugins/microbot/aiofighter/loot/LootScript.java
```diff
@@ -1,142 +1,157 @@
...
+        //TODO("Coal bag, Herb Sack, Seed pack")
+        return Rs2Inventory.contains(groundItem.getItemId());
+    }
+
+    private boolean shouldLootBasedOnValue(GroundItem groundItem, AIOFighterConfig config) {
+        if (config.looterStyle() != DefaultLooterStyle.GE_PRICE_RANGE && config.looterStyle() != DefaultLooterStyle.MIXED)
+            return false;
+        int price = groundItem.getGePrice();
+        return config.minPriceOfItemsToLoot() <= price && price / groundItem.getQuantity() <= config.maxPriceOfItemsToLoot();
+    }
```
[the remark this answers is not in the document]
runsonmypc: out of scope

runelite-client/src/main/java/net/runelite/client/plugins/microbot/aiofighter/loot/LootScript.java
```diff
@@ -1,142 +1,157 @@
...
+            itemNamesToLoot.add("coins");
+        }
+        if (config.toggleLootUntradables()) {
+            itemNamesToLoot.add("untradeable");
+            itemNamesToLoot.add("scroll box");
+        }
+        if (config.looterStyle().equals(DefaultLooterStyle.MIXED) || config.looterStyle().equals(DefaultLooterStyle.ITEM_LIST)) {
+            itemNamesToLoot.addAll(Arrays.asList(config.listOfItemsToLoot().trim().split(",")));
+        }
+        return itemNamesToLoot.toArray(new String[0]);
```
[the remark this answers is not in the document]
runsonmypc: out of scope

runelite-client/src/main/java/net/runelite/client/plugins/microbot/aiofighter/loot/LootScript.java
```diff
@@ -38,7 +38,7 @@ public boolean run(AIOFighterConfig config) {
                 if (AIOFighterPlugin.getState().equals(State.BANKING) || AIOFighterPlugin.getState().equals(State.WALKING)) {
                     return;
                 }
-                if (Rs2Player.isInCombat() && !config.toggleForceLoot()) {
+                if (Rs2Player.isInCombat() && !config.toggleForceLoot() && !AIOFighterPlugin.isWaitingForLoot()) {
```
gmason0, line 41: So, if we dont have toggleForceLoot enabled & we dont have isWaitingForLoot, we just aren't looting at all?

gmason0 (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/aiofighter/loot/LootScript.java
runsonmypc: It's:
  - Skip looting if: in combat AND force loot is off AND NOT waiting for loot
  - This means: Allow looting when waiting for loot, even if in combat

chsami (approved):

PR #1410, opened 2025-08-21
Feature: House Thieving improvements

Sunny-P:
Added BreakHandler state locking; No longer is able to take breaks while thieving a house.
Checks if house thieving objects are on screen before interacting; No longer interacts with off-screen objects.


Sunny-P:
Added a check in the statement it uses to go from BANKING to FINDING_HOUSE.
Would previously get stuck between BANKING & FINDING_HOUSE states after reaching less than 5 empty inventory spaces.

Now successfully banks and continues.

gmason0 (approved):

PR #1412, opened 2025-08-22
feat(TaskScript): Ported Dreambot's task node architecture to microbo…

Frozen-tuna:
…t for more structured scripting.

Glass Blowing Plugin is included as an example of this style of architecture. Tasks require accept() and execute() overrides. Plugins create a list of nodes to be iterated over prioritizing top to bottom. Every loop runs the first node where accept() === true. This allows for complex bots while maintaining clean structure. I'm open to any suggestions for cleaning up how the config is passed from the plugin to the tasks.

## Summary by CodeRabbit

- New Features
  - Introduces a Glass Blowing plugin for streamlined crafting.
  - Adds an on-screen overlay with status information while the plugin runs.
  - Provides a configuration option to select which glass product to craft (e.g., vials, lenses, orbs).
  - Automates banking and crafting actions when starting near a bank with a glassblowing pipe and molten glass.
  - Includes sensible defaults and runs only when enabled by the user.


gmason0 (changes requested):
Hey thank you for your contribution!

1. Can you change this pull request to be merged into development ? We never merge directly to main. 
2. Can you remove you plugin example from this & maybe create a md file on how the implementation works? This way it can be better explained to other microbot developers. We are also no longer accepting new plugins into the native client repo anymore. Any new plugins must be implemented into our plugin hub. https://github.com/chsami/microbot-hub

Frozen-tuna:
Summary

This PR adds a class called TaskScript that extends Script that manages the task-node architecture of OSRS scripting. 

    Clean Architecture: Allows script logic to be abstracted into smaller tasks based on simple accept() functions.
    Priority List: Nodes are checked and executed from top to bottom. The first task that returns true will execute.
    Separate responsibilities: Tasks can be broken up based on desired state.

Implementation Details

    TaskScript:
        Extends Script and overrides run() and shutdown(). Shutdown acts as normal but clears the node list. Run now executes tasks in the node list.

    Task:
        New class that is meant contain smaller pieces of bot logic based on state. accept() checks the state and return true or false if that task is ready to run. execute() contains logic for running the task.

Example
    A simple bot to blow glass would include a task for banking and a task for blowing. A more complex bot like Air Orbs would have the following tasks:
    BankTask
    RunToObeliskTask
    EnchantOrbsTask
    ReturnToBankTask

Frozen-tuna:
> Hey thank you for your contribution!
> 
>     1. Can you change this pull request to be merged into development ? We never merge directly to main.
> 
>     2. Can you remove you plugin example from this & maybe create a md file on how the implementation works? This way it can be better explained to other microbot developers. We are also no longer accepting new plugins into the native client repo anymore. Any new plugins must be implemented into our plugin hub. https://github.com/chsami/microbot-hub

Sure, I can do that. Where should I put an md file? The two important new classes are actually in the root of the microbot plugin. I could create a sub-folder for the new script type but then it would like like all the plugins. Are those getting phased out? I'm very open to suggestions for a java project.

Frozen-tuna:
Closing until I create a pr on the plugin hub demonstrating the new architecture.

PR #1441, opened 2025-08-29
Update quest helper to 4.9.4

vznnnnn:


gmason0:
@vznnnnn I appericate your time on this but I think I am going to see what @Voxsylvae has going on with the a more automatic system to update the quest helper plugin on #1442 

Thanks for taking the time to do this!

PR #1446, opened 2025-08-30
feat(AIOFighter): Reanimate ensouled heads

Krulvis:
## Summary by CodeRabbit

* **New Features**
  * Auto-reanimates ensouled heads when the Arceuus spellbook is active, checking rune requirements and using a suitable head from inventory.
  * Temporarily pauses other actions to cast, waits for the reanimated NPC to spawn, then resumes and prioritizes attacking the spawned target.
  * Skips reanimation while in BANKING or WALKING states and supports multiple head tiers for smoother combat loops.


gmason0:
@See1Duck - LMK what you think.

See1Duck (changes requested):
This should be an option in the config and not forced

Krulvis:
> This should be an option in the config and not forced

What makes you say that? My philosophy with scripts/plugins is that if you can determine preferred behavior from reading simple context then the only purpose of a Config is that someone can forget to turn it on. 
More configurable makes plugins more tedious to use.

No one is going to be accidently on arceuus spellbook with the exact runes for reanimating the specific type of ensouled head that is dropped by the monster they are fighting. In addition to this, we'd want to check if the spell can be casted regardless of whether they have the Config enabled so might as well use that check of spellbook + level + runes to determine if the plugin should cast it. I think the same can be said for casting demonic/siniater offering. If a player has brought all the requirements, they're obviously going to want to use it. 

I'm happy to discuss this and hear what other people have to say about maximizing vs. Minimizing configurations. Is there a general consensus on this for example?

chsami:
@See1Duck  and @Krulvis  can you guys find a consensus about this PR so we can merge/close it. Just to avoid having stuff open for to long :)

See1Duck (changes requested):
I have no issues with this, just make it optional and add a config option please

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/magic/reanimate/HeadType.java
```diff
@@ -0,0 +1,84 @@
...
+ * <p>
+ * Each tier specifies a spell from the `Rs2Spells` collection to be used for reanimation and
+ * a set of substring names used to identify the matching item models.
+ * <p>
+ * Features:
+ * - Maps tiers (BASIC, ADEPT, EXPERT, MASTER) to the `Rs2Spells` spell.
+ * - Provides utility to identify and retrieve head items fitting the defined categories
+ * using name substring matching.
+ */
+public enum HeadType {
```
SteffenCarlsen: A lot of this info is already present in net/runelite/client/plugins/microbot/TaF/EnsouledHeadSlayer/EnsouledHeads.java - Could be reuseable/extended instead of having two classes doing the same lifting?

SteffenCarlsen (changes requested):
I agree with Duck - This should be a config option. It would be terrible to be slaying on a Pure that just happens to be on the arceeus spellbook and with a runepouch or looted runes for the spell, to then just start getting prayer xp.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/magic/reanimate/HeadType.java
Krulvis: Oh yea I'll see how I can reuse to avoid redundant code

Krulvis:
> I agree with Duck - This should be a config option. It would be terrible to be slaying on a Pure that just happens to be on the arceeus spellbook and with a runepouch or looted runes for the spell, to then just start getting prayer xp.

Fair enough, I'll make it a Config option

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/magic/reanimate/HeadType.java
Krulvis: I've merged what I have and needed with I found there. Merged it both into re-usable Rs2EnsouledHead into the util/item/ package. Now either plugin can be moved to hub without causing breaking changes. Hopefully this works for you.

chsami (approved):

PR #1456, opened 2025-08-31
Poh teleportation support for WebWalker

Krulvis:
- Adds Transports towards PoH in appropriate TSV files
- Adds Rs2PohCache that keeps track of transport methods available in PoH
- Adds teleportation_poh.tsv and teleportation_portal_poh.tsv with all teleportation data from PoH
- Adds TransportationType.POH with acompanying handlePohTransport(Transport transport) for handling any transport from inside PoH.

Please go ahead and test the added web functionality to see if it breaks anything.

Currently added PoH teleportation systems:
- Portal Nexus
- Portal rooms
- Jewellery box
- Mounted glory
- Mounted digsite
- Mounted Xerics
- Myths cape
- Fairy rings

Missing:
- Tree Spirit
- Wilderness obelisk


.run/Runelite.run.xml
gmason0, line 1: Revert this file , we should see no diff here.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/poh/PohConfig.java
gmason0, line 1: I don't think we should store it in this manner, but more of using caching strategies  such as Rs2SpirtTreeCache.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/poh/data/JewelleryBox.java
gmason0, line 1: Couldn't we use data loaded from the TSV files here? Why create these java files when all the teleport data is available there?

This creates a disconnect between what the shortest-path plugin is already using.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/poh/ui/JewelleryBoxPanel.java
```diff
@@ -0,0 +1,50 @@
...
+import net.runelite.client.plugins.microbot.util.poh.data.JewelleryBox;
+
+import javax.swing.*;
+import javax.swing.border.TitledBorder;
+import java.awt.*;
+
+import static net.runelite.client.plugins.microbot.util.poh.PohConfig.CONFIG_GROUP;
+import static net.runelite.client.plugins.microbot.util.poh.PohConfig.JEWELLERY_BOX;
+
+public class JewelleryBoxPanel extends JPanel {
```
gmason0: Like i was saying before, we shouldn't even provide the option to manually say "i have xyz teleport" this can create false positives due to user error.

gmason0 (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/poh/ui/JewelleryBoxPanel.java
Krulvis: There is no manual option. User can only parse when they're in PoH. Interaction with the configs directly is disabled.

.run/Runelite.run.xml
Krulvis: Yea will do, shouldn't be here.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/poh/data/JewelleryBox.java
Krulvis: I didn't create all of them. There was already something in existence for PoH teleportation for Nexus and jewellery box so I wanted to expand on that. This way the teleportation is accessible for scripters as well as the web.

gmason0: I don't believe this would prevent or make it anymore difficult for scripters to use the POH by solely relying off the TSV data. No one should be using a transport object directly to interact with walking - all walking logic should be encapsulated into the walkTo method

Voxsylvae:
Hey Krulvix,
I noticed the PR for PoH teleport support - first, I want to thank you for taking initiative on this feature that the community definitely needs.
I wanted to reach out because I've been working on a similar implementation in the background and thought it might be valuable to share what I have implemented. I think your PR requires a bit too much configuration from the user side and doesn't directly integrate the PoH into our Rs2Walker implementation.
My Current Implementation Approach:
Architecture Integration:

Direct integration into Rs2Walker core rather than separate plugin dependency
Native support within the walkTo() method for seamless pathfinding
No external plugin required to track available teleports
Smart Routing & Optimization:

Automatic cost-benefit analysis - PoH routes are only selected when they provide meaningful time savings
Dynamic threshold configuration for when PoH should be considered
Full integration with existing optimal routing algorithms
Automatic State Management:

Real-time caching of available PoH teleports without manual user input
Automatic detection and updating of player's house configuration
Support for house advertisements and world-hopping scenarios
User Experience:

Zero configuration required from users (only if the user wants to use PoH at all)
Transparent operation - works behind the scenes without toggles or manual setup
Fallback handling when PoH access is unavailable
I'm not trying to diminish your work at all - I genuinely appreciate that you've tackled this complex feature! However, I wanted to share my perspective on why I believe my approach might offer some architectural advantages for the long-term maintainability of the project.
Key Benefits of My Implementation:
Direct State Caching:

The system maintains a live cache of PoH teleport availability, eliminating the need for users to manually configure or update their teleport options
This reduces user error and ensures routing decisions are always based on current game state
Rs2Walker Integration:

By embedding PoH logic directly into the core pathfinding algorithms, we avoid the overhead and complexity of inter-plugin communication
This creates a more robust and performant solution that's less prone to compatibility issues
That said, I completely understand if the Geroge and the community prefers to move forward with your implementation - I just don't think it makes sense for us to have two parallel solutions addressing the same feature, as this could fragment the base and create maintenance overhead for the project.
Moving Forward:

Would you be open to discussing the best path forward?
I'm flexible about either:
Abandoning my implementation if yours meets the needs
Exploring ways to incorporate some of these architectural concepts into your existing work
Having a technical discussion about the trade-offs between our approaches
I'm happy to discuss more technical details of my implementation if that would be helpful for making this decision. You can reach out to me on Discord if you'd prefer to chat there.
Looking forward to your thoughts on how we can best serve the community while avoiding duplicate effort!

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/poh/data/JewelleryBox.java
Krulvis: It does make it more difficult as a scripter to pick a teleport of your liking. Parsing the TSV's is done explicitly for the pathfinder and there is no easy way to find, filter and use them. 
"No one should be using a 'Transport' object directly to interact with walking" The way I've implemented PoH for the web using the existing PoH Teleports, they don't have to be used as Transport object. They are simply there to allow scripters to teleport at will AND can be used by web as Transport objects.
"all walking logic should be encapsulated into the walkTo method" This is a fallacy that a lot of clients abide by. But why? What is the reason to limit scripters and take away their possibility to use this. For example, the web is slow, takes time to fire up calculate a path and actually go somewhere. In the heat of the moment it is better to straight up call a preferred teleport. Second, it might be nice to let users select a preferred teleport to bank or escape with. When you walk around with a Construct. cape there are tons of options to go to a bank with, and why should a client force a specific way?

Krulvis:
> Hey Krulvix, I noticed the PR for PoH teleport support - first, I want to thank you for taking initiative on this feature that the community definitely needs. I wanted to reach out because I've been working on a similar implementation in the background and thought it might be valuable to share what I have implemented. I think your PR requires a bit too much configuration from the user side and doesn't directly integrate the PoH into our Rs2Walker implementation. My Current Implementation Approach: Architecture Integration:
> 
> Direct integration into Rs2Walker core rather than separate plugin dependency Native support within the walkTo() method for seamless pathfinding No external plugin required to track available teleports Smart Routing & Optimization:
> 
> Automatic cost-benefit analysis - PoH routes are only selected when they provide meaningful time savings Dynamic threshold configuration for when PoH should be considered Full integration with existing optimal routing algorithms Automatic State Management:
> 
> Real-time caching of available PoH teleports without manual user input Automatic detection and updating of player's house configuration Support for house advertisements and world-hopping scenarios User Experience:
> 
> Zero configuration required from users (only if the user wants to use PoH at all) Transparent operation - works behind the scenes without toggles or manual setup Fallback handling when PoH access is unavailable I'm not trying to diminish your work at all - I genuinely appreciate that you've tackled this complex feature! However, I wanted to share my perspective on why I believe my approach might offer some architectural advantages for the long-term maintainability of the project. Key Benefits of My Implementation: Direct State Caching:
> 
> The system maintains a live cache of PoH teleport availability, eliminating the need for users to manually configure or update their teleport options This reduces user error and ensures routing decisions are always based on current game state Rs2Walker Integration:
> 
> By embedding PoH logic directly into the core pathfinding algorithms, we avoid the overhead and complexity of inter-plugin communication This creates a more robust and performant solution that's less prone to compatibility issues That said, I completely understand if the Geroge and the community prefers to move forward with your implementation - I just don't think it makes sense for us to have two parallel solutions addressing the same feature, as this could fragment the base and create maintenance overhead for the project. Moving Forward:
> 
> Would you be open to discussing the best path forward? I'm flexible about either: Abandoning my implementation if yours meets the needs Exploring ways to incorporate some of these architectural concepts into your existing work Having a technical discussion about the trade-offs between our approaches I'm happy to discuss more technical details of my implementation if that would be helpful for making this decision. You can reach out to me on Discord if you'd prefer to chat there. Looking forward to your thoughts on how we can best serve the community while avoiding duplicate effort!

image
There is no 'user configuration' apart from going to your PoH once and pressing 'Detect available POH teleports'. As mentioned before by George I will probably rework it to use Rs2Cache. This will get rid of the PohPlugin and make it a more 'seamless integration into the Rs2Walker'.
I am definitely open to discuss and have noted down the feedback regarding the annoyance with the additional plugin and preferably using Rs2Cache. 
In addition to that, I am super open to discuss the best path forward. Hit me up on discord: Krulvis

Voxsylvae:
I just want to mention that I already have an implementation which is supposed to utilize the caching system, also support player advertisements for player-owned houses, with optimal route planning integrated into the walker. It includes automatic detection of the available teleports without the need for an additional plugin/panel. As in your implementation, the teleports and the interactions can also be freely used by any scriptwriter within the plugins with the web walker itself. My implementation directly uses the "teleportation" class and the teleports defined in the .tsv files, matching against whether a teleport "could" be a PoH teleport. I just want to know if I should abandon my current implementation because, as I said, I don't see the benefit of having duplicate functionality for this feature. I've been working on this feature for some time, and therefore I want to let you know what I have already built.

Krulvis:
> I just want to mention that I already have an implementation which is supposed to utilize the caching system, also support player advertisements for player-owned houses, with optimal route planning integrated into the walker. It includes automatic detection of the available teleports without the need for an additional plugin/panel. As in your implementation, the teleports and the interactions can also be freely used by any scriptwriter within the plugins with the web walker itself. My implementation directly uses the "teleportation" class and the teleports defined in the .tsv files, matching against whether a teleport "could" be a PoH teleport. I just want to know if I should abandon my current implementation because, as I said, I don't see the benefit of having duplicate functionality for this feature. I've been working on this feature for some time, and therefore I want to let you know what I have already built.

You're talking about this project, right? https://github.com/chsami/Microbot/compare/development...Voxsylvae:Microbot:PR_development_PathFinder_POH It was so big and I was already pretty far I didn't want to move to that code (sorry)
Let's go over it in-depth on discord so we can see what the best path forward is. There's definitely some good stuff in there. I am curious about a couple of things:
 - What do you mean by "the caching system"?  
 - How have you utilized said caching system?
 - What do you mean with "optimal route planning integrated into the walker"?
 - You say you directly use the "teleporation" class and teleports defined in the .tsv. I have not come across a "Teleportation" class myself so I am curious to learn about that.
 - What's the advantage of the TSV files? I found them frustrating to work with (no Java, external file-editor, requires parsing)

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/poh/data/JewelleryBox.java
gmason0: There shouldn't be a reason to filter & use them outside of the walker utility. 

The calculation for a path is typically about 60-120ms in most situations, you can see that inside of the debug options of the web walker plugin. 

The web walker is an abstraction layer that scripters can trust to find the fastest way possible to get from point a to point b. In 99.9% of scenarios - scripters or users would expect the shortest-path/walker utility to determine which method of transportation is the most viable & use that, not some pre-defined method that the scripter imposes. if scripters want to manually interact with teleports, i'd say let's just keep it more primitive such as direct interaction with objects widgets or items etc - not creating these pre-populated enums or classes. 

If we can't align with that the shortest-path plugin/walker utility should be the only place where we create these interaction methods instead of starting to segregate this stuff, we may have difficulties with merging this. This is an ideology that I personally have as a maintainer of this tool that the walker utility should be the single source of truth when it comes to traversal / transportation

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/poh/PohConfig.java
Krulvis: I've reworked the caching to use Rs2Cache and gotten rid of the PohPlugin. I have to be honest that I'm not sure if im a fan of this. The plugin provided a very transparent overview to other developers and users that is now lost. The implementation through Rs2Cache requires much more maintenance and is rather obfuscated with the additional requirement of Serializers, deserializers, multiple event listeners etc. I'd like to hear what others think

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/poh/data/JewelleryBox.java
Krulvis: It sounds to me like an arbitrary limitation on developers ability to use teleportation systems all for forcing them into using Rs2Walker.

I understand your 'single source of truth' but hey man, all this data is available already in Runelite and I am just linking that to specific teleportation systems. You're literally making it more difficult for yourself if you want to copy and paste all those values into tsv files.

gmason0: I think the correct term here is consistency - not limitation. Having this walker utility as the only place where developer are executing walking & transport interactions keeps things consolidated & in an area where  everything is tied together 

If we cant agree on this simple design aspect, i think we can close this.

Krulvis: Forcing plugins to use the same function for a different use-case by making it the only method available is limiting, not consistent. The code is consistent in the sense that regardless of how you call said transportation/teleportation method, it is always executing the same code under-the-hood. There will only ever be one place where the code has to be fixed if there's a problem.

If we can't agree on it through this discussion let me propose an alternative solution. I can't seem to think of a good way to even materialize the TSV methodology for Poh, even if I wanted to. How about you help me come up with a structure that suits your ideological needs while at the same time staying practical and not becoming a giant mess?

These are the requirements for Poh:
- Teleport methods' availability depend on **varps**, **gameObjects** and **tileObjects**.
- The varps determine which Nexus portal teleports are available
- The tileObjects are used to find **wall-mounted** teleportation systems (glory, xerics, digsite) which all provide a pre-dtermined list of teleportations
- The gameobjects are **portal** (id can be linked to teleportation), **jewellery box** (id is used to find the jewellery box type: ornate, fancy, basic), **fairy rings**, **spirit tree** and **wilderness obelisk**.
- The PoH is build up of a sprite map taken from the world-map as per: https://mejrs.github.io/osrs?m=-1&z=1&p=0&x=1931&y=5656. This means that the borders between rooms cause your player to be placed on a completely different place on the world map and your WorldPoint jumps all over the place. This means the web walker will not work inside the house.
- Caching teleport availability has to be in the form of: these objects exists and certain varbits have these values from which a list of available teleportations can be extracted.

How do you link objects and varps to the cache and to transportations/teleportations from TSV in a meaningful sensible way?

chsami:
Alright, I took an extensive look at both changes to the webwalker regarding POH

@Voxsylvae changes are a lot more in depth and cover a lot more scenario'. 

Route analysis and optimization -> Adds logic to compare POH/Banking/teleports, walking, and resource costs (runes, jewelry charges) to choose the most efficient travel method. Although this is really nice, it adds a big chunk of complexity. Something I really try to avoid in this project. In a way The POH changes are somewhat hidden by the larger routing analysis updates in the web walker. I also notice that a lot of the code still needs refactoring/cleaning.

On the other hand

The changes made by @Krulvis are more simplified and focussed on actually integrating POH without adding all the whistle and bells, keeping things simple and consistent with the existing patterns in the project. Although it is missing some features like the analysis whether to use the POH vs Banking/teleport, this change seems to be more focussed on the essence of what we are trying to achieve.

This isn’t about putting one approach against the other. The focus here is on the core goal of the feature within this PR, keeping attention on what we are trying to achieve rather than broad structural changes.

@Voxsylvae I really encourage you to make smaller bits of structural changes, so people can catch up and contribute as well to those changes. Large PRs hide smaller but important details (like your POH work being buried under routing analysis). A reviewer can’t mentally hold all of it, so subtle bugs slip through. 

We will go with the more simplistic approach here to reduce the risk of regressions.

**Changes required** 
@Krulvis 
Important to note: I would like to remove the current Java-to-TSV parser that you added. That should be handled as a separate project. At that point we can decide whether to replace TSV files with Java models, but we will also need a parser flexible enough to handle TSV replacement across all updates when we merge new changes of the shortestpath plugin in our repo.

Thanks, and I hope this explains my viewpoint clearly. I also hope this doesn’t take away from your motivation, since both contributions are valuable and needed.

chsami (approved):

PR #1464, opened 2025-09-03
chore(Rs2Walker): isInArea

Sunny-P:
The logic for the isInArea() methods are backwards.
This change doesn't change the actual logic, but changes the method description to properly reflect the logic that's applied.

The overloaded isInArea() method that takes in a centre point and a range previously created a rectangle with larger offsets on the x value, rather than an equally sided square with the calculated corners. This is now changed to create an equally sided square around the centre point. Was there a reason we were creating horizontally favoured rectangle areas?

These pedantic changes better reflect the applied logic without changing the original logic (besides creating an equal-sided area, instead of a horizontally favoured rectangle), so this shouldn't have any affect on the current usages of these methods in other scripts/plugins.

I originally became aware of this flawed logic when I tried to use the `isInArea(WorldPoint... worldPoints)` method to determine if I had the player in a certain area using custom points, instead of the simpler to use overloaded method `isInArea(WorldPoint centerOfArea, int range)`, and of course I ran into problems before looking into how the methods were operating.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/walker/Rs2Walker.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/walker/Rs2Walker.java
@@ -1951,15 +1951,15 @@ public class Rs2Walker {
     /**
      * Checks if the player's current location is within the specified area defined by the given world points.
      *
-     * @param worldPoints an array of two world points of the NW and SE corners of the area
+     * @param worldPoints an array of two world points of the SE and NW corners of the area
      * @return true if the player's current location is within the specified area, false otherwise
      */
     public static boolean isInArea(WorldPoint... worldPoints) {
         WorldPoint playerLocation = Rs2Player.getWorldLocation();
-        return playerLocation.getX() <= worldPoints[0].getX() &&   // NW corner x
-                playerLocation.getY() >= worldPoints[0].getY() &&   // NW corner y
-                playerLocation.getX() >= worldPoints[1].getX() &&   // SE corner x
-                playerLocation.getY() <= worldPoints[1].getY();     // SE corner Y
+        return playerLocation.getX() <= worldPoints[0].getX() &&   // SE corner x
+                playerLocation.getY() >= worldPoints[0].getY() &&   // SE corner y
+                playerLocation.getX() >= worldPoints[1].getX() &&   // NW corner x
+                playerLocation.getY() <= worldPoints[1].getY();     // NW corner Y
         // draws box from 2 points to check against all variations of player X,Y from said points.
     }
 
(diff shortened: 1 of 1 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/walker/Rs2Walker.java
```diff
@@ -1972,9 +1972,9 @@ public static boolean isInArea(WorldPoint... worldPoints) {
      */
     @Deprecated(since = "1.5.5", forRemoval = true)
     public static boolean isInArea(WorldPoint centerOfArea, int range) {
-        WorldPoint nwCorner = new WorldPoint(centerOfArea.getX() + range + range, centerOfArea.getY() - range, centerOfArea.getPlane());
```
Voxsylvae, line 1975: should we not change it here to follow the logic of WorldArea, calculating the 
northeast corner
```Java
ne = new WorldPoint(centerOfArea.getX() + radius, centerOfArea.getY() + radius, centerOfArea.getPlane()); 
```
and 
southwest corner 
```Java
sw  =  new WorldPoint(centerOfArea.getX() - radius, centerOfArea.getY() - radius,  centerOfArea.getPlane()); 
```
range -> radius ? 
or better directly use  use  WorldArea

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/walker/Rs2Walker.java
```diff
@@ -1951,15 +1951,15 @@ private static boolean handleWearableTeleports(Transport transport, int itemId)
     /**
      * Checks if the player's current location is within the specified area defined by the given world points.
      *
-     * @param worldPoints an array of two world points of the NW and SE corners of the area
+     * @param worldPoints an array of two world points of the SE and NW corners of the area
      * @return true if the player's current location is within the specified area, false otherwise
      */
     public static boolean isInArea(WorldPoint... worldPoints) {
         WorldPoint playerLocation = Rs2Player.getWorldLocation();
-        return playerLocation.getX() <= worldPoints[0].getX() &&   // NW corner x
```
Voxsylvae, line 1959: Change the logic here too.
```Java
playerLocation.getX() >= worldPoints[0].getX() &&   // SW corner x
playerLocation.getY() >= worldPoints[0].getY() &&   // SW corner y
playerLocation.getX() <= worldPoints[1].getX() &&   // NE corner x
playerLocation.getY() <= worldPoints[1].getY();     // NE corner Y
```

Sunny-P:
New commit should make isInArea(WorldPoint... worldPoints) more robust as it no longer matters which order we pass through our points.
Cache values from getter methods to keep calls to this method as performant as we can.

Krulvis:
> New commit should make isInArea(WorldPoint... worldPoints) more robust as it no longer matters which order we pass through our points. Cache values from getter methods to keep calls to this method as performant as we can.

damn nice coderabbitai fix <3 that really makes this PR golden

Sunny-P:
> damn nice coderabbitai fix <3 that really makes this PR golden

Haha yeah, wasn't a bad suggestion. Tweaked it a little, but it was reasonable

Krulvis:
> > damn nice coderabbitai fix <3 that really makes this PR golden
> 
> Haha yeah, wasn't a bad suggestion. Tweaked it a little, but it was reasonable

Reasonable is putting it very lightly

chsami (approved):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/walker/Rs2Walker.java
Sunny-P: > Change the logic here too.
> 
> ```java
> playerLocation.getX() >= worldPoints[0].getX() &&   // SW corner x
> playerLocation.getY() >= worldPoints[0].getY() &&   // SW corner y
> playerLocation.getX() <= worldPoints[1].getX() &&   // NE corner x
> playerLocation.getY() <= worldPoints[1].getY();     // NE corner Y
> ```

Im confused as to what you're referring to here. This was changed to not be so static in its approach.

PR #1476, opened 2025-09-05
feat(aiofighter): add wilderness improvements - looting bag support, blighted foods, ferox pool restoration, combat potion optimization

runsonmypc:
## Description
Adds several wilderness-focused improvements to the AIO Fighter plugin to enhance the combat experience in wilderness areas.

## Changes
- **Looting Bag Support**: Automatically empties looting bag when banking
- **Blighted Food Recognition**: Added support for blighted manta ray, anglerfish, and karambwan to the food system
- **Pool of Restoration**: New optional feature to use the Pool of Restoration at Ferox Enclave after banking (configurable in banking settings)
- **Combat Potion Optimization**: Combat potions now only consumed during active combat to prevent waste
- **Always empty containers when banking**: Previously, containers such as Herb sack were only being emptied when using inventory setups, now they will emptied whenever we bank

## Testing
- Tested looting bag emptying with various items
- Verified blighted foods are properly recognized by the banking system
- Confirmed Pool of Restoration usage at Ferox Enclave
- Verified combat potions are only consumed during combat

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/aiofighter/bank/BankerScript.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/aiofighter/bank/BankerScript.java
@@ -1,7 +1,9 @@
 package net.runelite.client.plugins.microbot.aiofighter.bank;
 
 import lombok.extern.slf4j.Slf4j;
+import net.runelite.api.GameObject;
 import net.runelite.api.ItemComposition;
+import net.runelite.api.Skill;
 import net.runelite.api.coords.WorldPoint;
 import net.runelite.api.gameval.ItemID;
 import net.runelite.client.plugins.microbot.Microbot;
@@ -16,7 +18,9 @@ import net.runelite.client.plugins.microbot.inventorysetups.InventorySetup;
 import net.runelite.client.plugins.microbot.inventorysetups.MInventorySetupsPlugin;
 import net.runelite.client.plugins.microbot.util.Rs2InventorySetup;
 import net.runelite.client.plugins.microbot.util.bank.Rs2Bank;
+import net.runelite.client.plugins.microbot.util.bank.enums.BankLocation;
 import net.runelite.client.plugins.microbot.util.equipment.Rs2Equipment;
+import net.runelite.client.plugins.microbot.util.gameobject.Rs2GameObject;
 import net.runelite.client.plugins.microbot.util.inventory.Rs2Inventory;
 import net.runelite.client.plugins.microbot.util.inventory.Rs2ItemModel;
 import net.runelite.client.plugins.microbot.util.math.Rs2Random;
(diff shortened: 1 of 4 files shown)


chsami:
@See1Duck Can you take a look at the changes and see if they align with yours.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/aiofighter/bank/BankerScript.java
```diff
@@ -473,9 +478,40 @@ public boolean handleBanking() {
         Rs2Prayer.disableAllPrayers();
         if (Rs2Bank.walkToBankAndUseBank()) {
             withdrawUpkeepItems(config);
+            
+            // Use Pool of Restoration at Ferox if configured
+            if (config.usePoolAtFerox() && Rs2Bank.getNearestBank() == BankLocation.FEROX_ENCLAVE) {
```
See1Duck, line 483: Was it intended that it should only use the pool if the nearest bank is ferox? Wouldn't it make more sense to force it to select ferox if this option is on?

See1Duck (changes requested):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/aiofighter/bank/BankerScript.java
runsonmypc: This could potentially force a player through the wilderness so I did not consider it. An option could be added to force Ferox if we have a dueling ring, but I think a forced bank location feature should be its own PR.

See1Duck: Yeah i see your point, im just thinking that the setting could be somewhat misleading for some users expecting it will teleport to ferox and just ends up going somewhere else

Bolado: In my opinion if the bank location is Ferox, it should use the pool by default and not as an option (who banks at Ferox if not because of the pool? Castle Wars chest is closer if chosen by distance), and maybe in the place of that option, replace with "Use Ferox Bank" , which would set Ferox as bank location (consequentially using the pool), just an suggestion tho  💯

See1Duck: I agree

runsonmypc: Actually you don't always want to use the pool because you will lose any boosted stats, that's why I made it an option. It's also why I changed combat pots to only be used in combat, to avoid pot -> pool -> pot.

chsami (approved):

PR #1493, opened 2025-09-08
Spirit Tree cache for PoH

Krulvis:
- Removed method of detecting PoH Spirit trees on GameState changes as it didn't work
- Added method for handling GameObjectSpawned and GameObjectDespawned to detect spirit tree in PoH. The GameObjectSpawned event is fired after having entered the PoH. The GameObjectDespawned event is only fired when the spirit tree is removed.
- Omitting the SpiritTree's level check for PoH spirit tree's in `isAvailableForTravel()` as it is not a requirement to use the spirit tree
- Check all requirements when setting default value loader for non-existing cache

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/cache/strategy/farming/SpiritTreeUpdateStrategy.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/cache/strategy/farming/SpiritTreeUpdateStrategy.java
@@ -6,16 +6,13 @@ import net.runelite.api.GameState;
 import net.runelite.api.GameObject;
 import net.runelite.api.Skill;
 import net.runelite.api.coords.WorldPoint;
-import net.runelite.api.events.GameStateChanged;
-import net.runelite.api.events.VarbitChanged;
-import net.runelite.api.events.WidgetLoaded;
+import net.runelite.api.events.*;
 import net.runelite.api.gameval.ObjectID;
 import net.runelite.api.gameval.VarbitID;
 import net.runelite.api.widgets.Widget;
 import net.runelite.client.plugins.microbot.Microbot;
 import net.runelite.client.plugins.microbot.questhelper.helpers.mischelpers.farmruns.CropState;
 import net.runelite.client.plugins.microbot.util.cache.Rs2Cache;
-import net.runelite.client.plugins.microbot.util.cache.Rs2ObjectCache;
 import net.runelite.client.plugins.microbot.util.cache.model.SpiritTreeData;
 import net.runelite.client.plugins.microbot.util.cache.strategy.CacheOperations;
 import net.runelite.client.plugins.microbot.util.cache.strategy.CacheUpdateStrategy;
@@ -63,13 +60,18 @@ public class SpiritTreeUpdateStrategy implements CacheUpdateStrategy<SpiritTree,
     @Override
     public void handleEvent(Object event, CacheOperations<SpiritTree, SpiritTreeData> cache) {
         try {
-            
             if (event instanceof WidgetLoaded) {
                 handleWidgetLoaded((WidgetLoaded) event, cache);
             } else if (event instanceof VarbitChanged) {
                 handleVarbitChanged((VarbitChanged) event, cache);
             } else if (event instanceof GameStateChanged) {
                 handleGameStateChanged((GameStateChanged) event, cache);
+            } else if (event instanceof GameObjectSpawned) {
+                GameObject go = ((GameObjectSpawned) event).getGameObject();
+                handleGameObjectChange(go, true, cache);
+            } else if (event instanceof GameObjectDespawned) {
+                GameObject go = ((GameObjectDespawned) event).getGameObject();
+                handleGameObjectChange(go, false, cache);
             }
         } catch (Exception e) {
             log.error("Error handling event in SpiritTreeUpdateStrategy: {}", e.getMessage(), e);
@@ -78,7 +80,7 @@ public class SpiritTreeUpdateStrategy implements CacheUpdateStrategy<SpiritTree,
     
     @Override
     public Class<?>[] getHandledEventTypes() {
-        return new Class<?>[]{WidgetLoaded.class, VarbitChanged.class, GameStateChanged.class};
+        return new Class<?>[]{WidgetLoaded.class, VarbitChanged.class, GameStateChanged.class, GameObjectSpawned.class, GameObjectDespawned.class};
     }    
     /**
      * Handle widget loaded events to detect spirit tree widget opening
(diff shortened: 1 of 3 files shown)


runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/farming/SpiritTree.java
```diff
@@ -249,16 +250,6 @@ public boolean isAvailableForTravel() {
             return isPatchHealthyAndGrown();
         }
 
-        // Check construction level for POH trees
```
chsami, line 252: Can you elaborate why you removed this? I'm not fully understanding this change.

Krulvis: I removed it from the function `isAvailableForTravel()`. Your construction or farming levels have nothing to do with whether you can use a tree for travel. It either exists or it doesn't. And when it does, you can use it. 
There are many people that boost their farming/construction levels to plant the tree. As such, this lead to unintended blocking behavior. 


**Additional rant:**
I would even go as far as saying that all the level requirements in this file should be removed as planting spirit trees comes with higher level requirements the more trees you plant. "At level 83 Farming, players can only have one spirit tree planted at a time, not including the one in their player-owned house. Level 91 Farming allows for two spirit trees to be planted at a time, whilst 99 Farming allows for unlimited spirit trees." But -to make things more complex- the tree in your PoH does not count towards this. "Players can plant their own spirit tree from a spirit sapling in their Superior Garden at level 75 Construction. Doing so also requires level 83 in Farming. Boosts can be used. This tree does not count towards the number of spirit trees that may be planted at one time."

There is also a `hasLevelRequirement()` function where the level is being looked at, but that doesn't consider the amount of trees you have and uses `Rs2Player.getRealSkillLevel(Skill.FARMING) >= requiredSkillLevel` instead of getBoostedSkillLevel(). None of this code is used anywhere. If it would be used for actual farming, it wouldn't work.

chsami: > I removed it from the function `isAvailableForTravel()`. Your construction or farming levels have nothing to do with whether you can use a tree for travel. It either exists or it doesn't. And when it does, you can use it. There are many people that boost their farming/construction levels to plant the tree. As such, this lead to unintended blocking behavior.
> 
> **Additional rant:** I would even go as far as saying that all the level requirements in this file should be removed as planting spirit trees comes with higher level requirements the more trees you plant. "At level 83 Farming, players can only have one spirit tree planted at a time, not including the one in their player-owned house. Level 91 Farming allows for two spirit trees to be planted at a time, whilst 99 Farming allows for unlimited spirit trees." But -to make things more complex- the tree in your PoH does not count towards this. "Players can plant their own spirit tree from a spirit sapling in their Superior Garden at level 75 Construction. Doing so also requires level 83 in Farming. Boosts can be used. This tree does not count towards the number of spirit trees that may be planted at one time."
> 
> There is also a `hasLevelRequirement()` function where the level is being looked at, but that doesn't consider the amount of trees you have and uses `Rs2Player.getRealSkillLevel(Skill.FARMING) >= requiredSkillLevel` instead of getBoostedSkillLevel(). None of this code is used anywhere. If it would be used for actual farming, it wouldn't work.

Makes sense

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/farming/SpiritTree.java
```diff
@@ -199,9 +200,9 @@ public enum SpiritTree {
             null, // Location is dynamic based on player's house
             SpiritTreeType.POH,
             List.of(Quest.TREE_GNOME_VILLAGE, Quest.THE_GRAND_TREE, Quest.FAIRYTALE_II__CURE_A_QUEEN),
-            75, // Requires 75 Construction
+            95, // Requires 95 Construction
             VarbitID.POH_SPIRIT_TREE_UPROOTED, // TODO Must be checkedPOH Spirit Tree varbit (combined tree/ring uses same varbit)
-            List.of(ObjectID.POH_SPIRIT_RING), // TODO Must be checked Spirit Ring object ID // we must update it. here are also variations for it.. leauge skins
```
Voxsylvae, line 204: Why use Rs2GameObject here and not the actual IDs from ObjectID? Not using the variations and also only the spirit tree variation with the fairy ring.
only using "poh_spirit_ring"  is the wrong approach here.  -  is it not only the spirit tree with the fairy ring? What about the spirit tree without the fairy ring? Here we must consider not only https://oldschool.runescape.wiki/w/Spirit_tree_%26_fairy_ring, but we should add all variations for https://oldschool.runescape.wiki/w/Spirit_tree_(Construction) and the Spirit tree & fairy ring.https://oldschool.runescape.wiki/w/Spirit_tree_(Construction) -> why is the construction level changed to 95?
https://oldschool.runescape.wiki/w/Superior_garden

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/cache/strategy/farming/SpiritTreeUpdateStrategy.java
```diff
@@ -148,53 +184,11 @@ private void handleGameStateChanged(GameStateChanged event, CacheOperations<Spir
...
-            if (Rs2Cache.isInPOH()) {
-                log.debug("Player in POH, checking for spirit tree objects");
-                updatePOHSpiritTreeCache(cache);
-            }
         } catch (Exception e) {
             log.error("Error handling GameStateChanged in SpiritTreeUpdateStrategy: {}", e.getMessage(), e);
         }
     }
     
-    /**
```
Voxsylvae, line 162: Why remove these functions entirely, and why can't we update the state based on the varbit for the spirit tree?We must wait for the object cache check to complete the scan and have populated the update. In the handleGameStateChanged, you cannot use the cache because it's not updated, so we would need to wait for the object cache to be updated on the object spawn and despawn events, also on a gamestate change event.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/farming/SpiritTree.java
Krulvis: > Why use Rs2GameObject here and not the actual IDs from ObjectID? Not using the variations and also only the spirit tree variation with the fairy ring.

Well, the Rs2GameObjects.getObjectIdsByName function returns a list of all IDS we're interested in, which is exactly what you also want I believe? https://github.com/chsami/Microbot/blob/2239281943273591fa45dda21bd337a23fbc22af/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/gameobject/Rs2GameObject.java#L1868

Krulvis: The requirement for Spirit tree with fairy rings is 95 construction: https://oldschool.runescape.wiki/w/Spirit_tree_%26_fairy_ring

Voxsylvae: i see why it better to use the reflection approach here, because of the different IDs per configuration, great job on the changes otherwise.
Regarding the level. Sorry, my bad, have not seen the other one.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/cache/strategy/farming/SpiritTreeUpdateStrategy.java
Krulvis: You can use the varbit still. I didn't touch the part for POH spirit tree varbits that you left with //TODO to still be implemented:
```java
if(varbitId == VarbitID.POH_SPIRIT_TREE_UPROOTED){
    log.debug("TODO update cache for POH spirit tree uprooted varbit change,currently in POH? {} ",PohTeleports.isInHouse());
}
```        
What I removed from handleVarbitChange was just a proxy for scanning for gameobjects:
```java
if (Rs2Cache.isInPOH()) {
       log.debug("Player in POH, checking for spirit tree objects,changed varbit {}", varbitId);
       updatePOHSpiritTreeCache(cache);
}
```

> In the handleGameStateChanged, you cannot use the cache because it's not updated, so we would need to wait for the object cache to be updated on the object spawn and despawn events, also on a gamestate change event.

This is what I gave as a reason to remove scanning for objects in POH from the handleGameStateChanged function. I'm not sure what you're saying here exaclty.

Voxsylvae: You have to delay the scan, of course, until the data are stable. This is what I'm referring to.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/farming/SpiritTree.java
Krulvis: > i see why it better to use the reflection approach here, because of the different IDs per configuration, great job on the changes otherwise. Regarding the level. Sorry, my bad, have not seen the other one.

Aah I understand your confusion. It is unclear from the diff that I am changing only the Spirit Rings, not Spirit Tree enum value.

runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/cache/strategy/farming/SpiritTreeUpdateStrategy.java
Krulvis: > You have to delay the scan, of course, until the data are stable. This is what I'm referring to.

Yes that would be an option. How long are you going to delay it for? Until the game object spawns preferably, right? So how about we just look at the gameobjectspawns event instead and skip the delaying setup altogether?

PR #1500, opened 2025-09-10
Tempoross null fix

T-edit:
This will fix the NullpointException error for Tempoross plugin. Will stop all the crashes/freezes that users are experiencing


runelite-client/src/main/java/net/runelite/client/plugins/microbot/tempoross/TemporossScript.java
```diff
@@ -671,10 +671,12 @@ private void handleMainLoop() {
...
-                    if (Rs2Player.isInteracting() && Rs2Player.getInteracting() != null) {
-                        Actor interactingNpc = Microbot.getClient().getLocalPlayer().getInteracting();
-                        if (interactingNpc.equals(fishSpot.getActor())) {
-                            return;
+                    if (Rs2Player.isInteracting()) {
+                        Actor currentTarget = Rs2Player.getInteracting();
+                        if (currentTarget != null && currentTarget instanceof NPC) {
+                            NPC targetNpc = (NPC) currentTarget;
+                          if (targetNpc.getId() == fishSpot.getId()) {
+                                return;
```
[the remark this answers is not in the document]
T-edit: Rs2NpcModel is an NPC, instanceof NPC will be true for Rs2NpcModel.

chsami (approved):

runelite-client/src/main/java/net/runelite/client/plugins/microbot/tempoross/TemporossScript.java
```diff
@@ -671,10 +671,13 @@ private void handleMainLoop() {
...
                     }
-                    if (Rs2Player.isInteracting() && Rs2Player.getInteracting() != null) {
-                        Actor interactingNpc = Microbot.getClient().getLocalPlayer().getInteracting();
-                        if (interactingNpc.equals(fishSpot.getActor())) {
-                            return;
+                    if (Rs2Player.isInteracting()) {
+                        Actor currentTarget = Rs2Player.getInteracting();
+                        if (currentTarget != null && currentTarget instanceof NPC) {
+                            NPC targetNpc = (NPC) currentTarget;
+                          if (targetNpc.getId() == fishSpot.getId()) {
```
chsami, line 678: doesn't this have to be the index of the npc instead of id?

two of the same npcs can have for example ID: 1

But the index will be different for both of the npcs

T-edit: targetNpc.getId() == fishSpot.getId() is enough it only needs to determine if the interacting NPC is a fishing spot (by type), not a specific instance. There are only two possible fishing spot IDs, and the code prioritizes the special fishing spot (double fishing spawn) when it appears. 

There will always only ever be 2 fishing spots available and it will interact with closest one available, doesn't matter which one. If special fishing spot (double fishing) appears it will always prioritize that. Works exactly like it did previously without null issues

PR #1515, opened 2025-09-12
Added Offensive Combat Prayers and Max Random Delays for Equipping Combat Gear

SamBunker:
Added offensive prayers hotkeys with dropdown menus for people to choose which offensive prayers for melee, range, and mage to use.

Added a max equipment delay setting with a default delay of 500 ms to randomize each equipped item. Originally, it seems each item equipped in the list of items had the same timing with little to no variation in timing.


chsami:
Please make a PR to development and not main.

PR #1516, opened 2025-09-12
Added Royal Titans Prayers to Microbot

SamBunker:
Added the Deadeye and Mystic Vigour prayers to Microbot. These prayers are obtained through royal titans, and they replace the current position of both Eagle Eye and Mystic Might prayers in the prayer book on members' worlds.

Any mid-game or late-game accounts that don't have rigour and or augury unlocked may use these prayers.


chsami (approved):

chsami:
Please make a PR to development and not main.

PR #1517, opened 2025-09-13
Implement MicrobotScript and update WelcomeScreenEvent

Bolado:
Without a plugin initializing a class that extends `Script`, the `run()` method in `Script.java` would never be invoked, preventing the events queued in `BlockingEvent` from being executed. Creating a `Script` class for the `MicrobotPlugin` ensures it is always executed, effectively resolving the core issue of `BlockingEvents` not being processed.

Also updated the Welcome Screen blocking event to use ID from the InterfaceID class instead of using hardcoded ID.

## Github Copilot Summary

This pull request introduces a new `MicrobotScript` class and integrates it into the `MicrobotPlugin`, as well as refactors the `WelcomeScreenEvent` to use a more robust widget identification approach. The main changes improve script execution management and widget handling reliability.

Script execution improvements:
* Added a new `MicrobotScript` class that extends `Script`, implementing a scheduled task to run the script logic periodically with exception handling. (`runelite-client/src/main/java/net/runelite/client/plugins/microbot/MicrobotScript.java`)
* Injected `MicrobotScript` into `MicrobotPlugin` and updated the `startUp` method to run the script on plugin startup. (`MicrobotPlugin.java`) [[1]](diffhunk://#diff-2061831859136e1410676ea04b9dbef1e2fb913a5ffc85421323d4b911373fdeR91-R93) [[2]](diffhunk://#diff-2061831859136e1410676ea04b9dbef1e2fb913a5ffc85421323d4b911373fdeL180-R183)

Widget handling and event refactoring:
* Refactored `WelcomeScreenEvent` to use `InterfaceID.WelcomeScreen.PLAY` for widget identification instead of a hardcoded component ID, improving maintainability and reliability. Also added null checks and logging for better debugging. (`WelcomeScreenEvent.java`)

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/MicrobotPlugin.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/MicrobotPlugin.java
@@ -88,6 +88,9 @@ public class MicrobotPlugin extends Plugin
 	@Inject
 	private MicrobotConfig microbotConfig;
 
+	@Inject
+	private MicrobotScript microbotScript;
+
 	private MicrobotTopLevelConfigPanel topLevelConfigPanel;
 
 	private NavigationButton navButton;
@@ -177,7 +180,7 @@ public class MicrobotPlugin extends Plugin
 		clientToolbar.addNavigation(navButton);
 
 		new InputSelector(clientToolbar);
-
+		microbotScript.run();
 		Microbot.getPouchScript().startUp();
 
 		// Initialize the cache system
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/MicrobotScript.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/MicrobotScript.java
@@ -6,9 +6,11 @@ import lombok.extern.slf4j.Slf4j;
 public class MicrobotScript extends  Script{
     @Override
     public boolean run() {
+        if (isRunning()) {
+            return true;
+        }
         mainScheduledFuture = scheduledExecutorService.scheduleWithFixedDelay(() -> {
             try {
-                if (!Microbot.isLoggedIn()) return;
                 if (!super.run()) return;
             } catch (Exception ex) {
                 Microbot.logStackTrace(this.getClass().getSimpleName(), ex);
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/MicrobotScript.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/MicrobotScript.java
@@ -1,0 +1,19 @@
+package net.runelite.client.plugins.microbot;
+
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class MicrobotScript extends  Script{
+    @Override
+    public boolean run() {
+        mainScheduledFuture = scheduledExecutorService.scheduleWithFixedDelay(() -> {
+            try {
+                if (!Microbot.isLoggedIn()) return;
+                if (!super.run()) return;
+            } catch (Exception ex) {
+                Microbot.logStackTrace(this.getClass().getSimpleName(), ex);
+            }
+        }, 0, 600, java.util.concurrent.TimeUnit.MILLISECONDS);
+        return true;
+    }
+}
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/events/WelcomeScreenEvent.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/events/WelcomeScreenEvent.java
@@ -1,6 +1,7 @@
 package net.runelite.client.plugins.microbot.util.events;
 
 import net.runelite.api.annotations.Component;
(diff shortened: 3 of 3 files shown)


chsami:
@Bolado I'm not sure if I understand this change correctly but If no plugin is enabled, no automation should run. With this change, automation will run regardless of whether the user has a plugin enabled.

Is this correct?

Bolado:
> @Bolado I'm not sure if I understand this change correctly but If no plugin is enabled, no automation should run. With this change, automation will run regardless of whether the user has a plugin enabled.
> 
> Is this correct?

The only "automation" that runs is the ones added to the `BlockingEventManager`, such as the WelcomeScreen clicking event, which used to be ran before v2. There is also the Pause blocking event, which will still hold the script. However, there is an underlying issue with pause, if a plugin pauses all other scripts and then get stuck, into an infinite loop or shut down without unpausing, re-enabling it or any other plugin wont work because all classes which extends on Script is now paused. This requires investigation into the plugin itself, which I couldnt do yet.

George suggested that the solution could be added to the `ShortestPathScript` instead, which always runs with the client, it's another option to have the WelcomeScreen and other utility BlockingEvents being done with no other plugins running.

If it's not ideal, I can just remove as well, to only have the fix to WelcomeScreen's ID, let me know.

chsami:
> > @Bolado I'm not sure if I understand this change correctly but If no plugin is enabled, no automation should run. With this change, automation will run regardless of whether the user has a plugin enabled.
> > Is this correct?
> 
> The only "automation" that runs is the ones added to the `BlockingEventManager`, such as the WelcomeScreen clicking event, which used to be ran before v2. There is also the Pause blocking event, which will still hold the script. However, there is an underlying issue with pause, if a plugin pauses all other scripts and then get stuck, into an infinite loop or shut down without unpausing, re-enabling it or any other plugin wont work because all classes which extends on Script is now paused. This requires investigation into the plugin itself, which I couldnt do yet.
> 
> George suggested that the solution could be added to the `ShortestPathScript` instead, which always runs with the client, it's another option to have the WelcomeScreen and other utility BlockingEvents being done with no other plugins running.
> 
> If it's not ideal, I can just remove as well, to only have the fix to WelcomeScreen's ID, let me know.

Thanks for the clarification, I'll need some time and investigation to think about this one.

chsami:
> > @Bolado I'm not sure if I understand this change correctly but If no plugin is enabled, no automation should run. With this change, automation will run regardless of whether the user has a plugin enabled.
> > Is this correct?
> 
> The only "automation" that runs is the ones added to the `BlockingEventManager`, such as the WelcomeScreen clicking event, which used to be ran before v2. There is also the Pause blocking event, which will still hold the script. However, there is an underlying issue with pause, if a plugin pauses all other scripts and then get stuck, into an infinite loop or shut down without unpausing, re-enabling it or any other plugin wont work because all classes which extends on Script is now paused. This requires investigation into the plugin itself, which I couldnt do yet.
> 
> George suggested that the solution could be added to the `ShortestPathScript` instead, which always runs with the client, it's another option to have the WelcomeScreen and other utility BlockingEvents being done with no other plugins running.
> 
> If it's not ideal, I can just remove as well, to only have the fix to WelcomeScreen's ID, let me know.

I had some time to think about the problem.

I think the issue you mentioned that others script that set pauseAllScript to true and for some reason crash and never reset is a valid concern and this issue mostly comes from the way we handle the pausing and unpausing.

I will use this code from tormenteddemon script as an example, but there are many cases that do the same:

```java
Microbot.pauseAllScripts.compareAndSet(false, true);
sleep(5000);
attemptLooting(config); // <----- If this code crashes, we never reset pauseAllScripts
lootAttempted = true;
Microbot.pauseAllScripts.compareAndSet(true, false);
```

The idea of pauseAllScripts is that you want to pause all the scripts, do one ore more actions and reset it afterwards. This means that the one or multiple actions made between setting pauseAllScripts to true and then setting it back to false should be transactional. If anything in between the state change fails, it should reset to it's original state.


So I think that we should move to something more like:

A manager class that handles the transactional logic with a method

```java
public static void runPaused(Runnable action) {
    pauseLock.lock();
    try {
        action.run();
    } finally {
        pauseLock.unlock();
    }
}

// impl

PauseManager.runPaused(() -> {
    Thread.sleep(5000);
    attemptLooting(config);
    lootAttempted = true;
});
```

Wdyt?

Bolado:
> A manager class that handles the transactional logic with a method
> 
> ```java
> public static void runPaused(Runnable action) {
>     pauseLock.lock();
>     try {
>         action.run();
>     } finally {
>         pauseLock.unlock();
>     }
> }
> 
> // impl
> 
> PauseManager.runPaused(() -> {
>     Thread.sleep(5000);
>     attemptLooting(config);
>     lootAttempted = true;
> });
> ```
> 
> Wdyt?

Looks very straightforward, especially on the implementation side of things!  

I'm going to close this PR for now, considering the overall scope of things. The script idea wouldn't have fixed much, so I'll submit a PR later with only the WelcomeScreenEvent, which I'm currently adding more things to it. 💯

PR #1535, opened 2025-09-23
feat(WebWalker): Teleport Tab to Inside for poh

Krulvis:
Teleport to inside the poh rather than outside with the teleport tab

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/poh/PohTeleports.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/poh/PohTeleports.java
@@ -247,14 +247,13 @@ public class PohTeleports {
         WorldPoint outsidePoint = location.getPortalLocation();
 
         transportMap.put(null, Set.of(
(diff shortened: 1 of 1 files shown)


Krulvis:
> **Actionable comments posted: 0**
> 🧹 Nitpick comments (1)
> 
> > runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/poh/PohTeleports.java (1)> `251-252`: **POH arrival toggle not accounted for; inside-point may be wrong for many users**
> > > Both the house spell and the tablet obey the player's “Teleport inside house” setting. Hardcoding the destination to insidePoint will misroute pathfinding when that toggle is off (tablets can’t force inside via menu). Please resolve by detecting the current house arrival setting and choosing insidePoint vs outsidePoint accordingly, or gate behind a config with correct runtime detection.
> 
> 📜 Review details

House Tablets have 'break', 'inside' and 'outside' as options. Teleport to House spell has 'cast' and 'outside' where cast always teleports the player inside the poh

chsami (approved):

PR #1538, opened 2025-09-24
Construction 2, Rs2GroundItem,  and Rs2GrandExchange improvements.

Gage307:
Construction 2: Added support for break handler, added support for both tile objects and game objects as both are needed depending on what we're building.

Rs2GroundItem: Fixed a bug that would cause Rev Killer to click loot on the ground over and over despite a full inventory.

Rs2GrandExchange: Fixed a bug that would cause the script to loop buying an item unsuccessfully; if we've previously searched the item within the Grand Exchange.


chsami:
the construction2 scripts have been moved to the hub. plese take out these changes and apply them in the hub if you like.

Thx

Gage307:
> the construction2 scripts have been moved to the hub. plese take out these changes and apply them in the hub if you like.
> 
> Thx

I'm on it!

PR #1579, opened 2025-10-20
feat(Rs2Cannon): Added pickup and start for cannon fix(Rs2Equipment): doInvoke with option and target

Sorries:
- Added pickup and start for cannon ( the start should be incorporated within qolCannonScript before refill() )
- Fix the NewMenuEntry when doInvoke as old one would set target as ""


chsami:
Does not meet coding standard

PR #1583, opened 2025-10-22
ReturnTab boolean function in Rs2Prayer.toggle

Jamdrizzle:
Added a new optional boolean function in Rs2Prayer.toggle that has "returnTab" of which the purpose is to do the original function and then return to the previous tab after the prayer is turned on. This will be useful for PvM purposes where staying on another tab is ideal for manual play, or otherwise.


Krulvis:
I would advice against adding behavioral logic like that to this function. The added functionality of going back to the previous tab has nothing to do with the original function and only adds more code to maintain. I believe it is better to make sure that this behavior is added when necessary to a script that actually requires it. If you want to reduce duplicate code you can create a wrapper function inside the script that requires this.

PR #1663, opened 2026-01-28
fix(walker): handle dialogue for inventory teleport items

Chillibloke:
Add dialogue handling for multi-destination teleport items when used from inventory (not equipped). Previously, items like Slayer Ring would interact successfully but fail to handle the destination selection dialogue that appears afterward.

Supported items: Slayer ring, Games necklace, Skills necklace, Ring of dueling, Ring of wealth, Amulet of glory, Combat bracelet, Digsite pendant, Necklace of passage, Giantsoul amulet.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/walker/Rs2Walker.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/walker/Rs2Walker.java
@@ -1925,6 +1925,18 @@ public class Rs2Walker {
 
             if (itemAction.equalsIgnoreCase("open") && itemId == ItemID.BOOKOFSCROLLS_CHARGED) {
                 return handleMasterScrollBook(destination);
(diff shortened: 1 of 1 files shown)


chsami:
I'm  not sure if I understand this change I don't think this is needed, we use the immediate action. I just tested the walker with games necklace and it teleports correctly.

Image

Chillibloke:
It was definitely not working specifically with the slayer ring, was opening the dialogue and not going through, was it was not using the right click options

chsami:
> It was definitely not working specifically with the slayer ring, was opening the dialogue and not going through, was it was not using the right click options

Maybe this is only for slayer ring specific?

Chillibloke:
> > It was definitely not working specifically with the slayer ring, was opening the dialogue and not going through, was it was not using the right click options
> 
> Maybe this is only for slayer ring specific?

Potentially, either way this has fixed it for me!

PR #1675, opened 2026-02-04
feat: Added bank IDs and container emptying utility

irkedMATT:
Added bank IDs for locations:
- SUNBLEAK_ISLAND
- DEEPFIN_POINT
- DEEPFIN_MINE_MID
- DEEPFIN_MINE_EAST
- CHARRED_ISLAND

Rs2Bank API Enhancement:
- Added `emptyContainers()` method to empty containers in inventory via the bank interface
- Uses widget ID 786471 to trigger the empty containers action


Sunny-P:
Nice increased functionality.
If I could add a suggestion; I would propose that instead of using an interface ID to find the empty container Widget, use a sprite ID - the same way the empty inventory and equipment buttons are found. This is a more more resilient way so it is less likely to breaking in future, as any changes/updates made to the bank in future will more likely change interface IDs around, but still use the same sprite ID.

chsami (approved):

PR #1751, opened 2026-04-19
fix(shortestpath): TransportNode cost double-count + overflow; remove MoA audit temp plugin

runsonmypc:
## Summary

Three bundled changes on one branch to avoid back-to-back Rs2Walker baseline-regen conflicts:

### 1. `TransportNode` cost fix (main)

- `TransportNode` was stamping chained-transport costs at `2 * previous.cost + travelTime` instead of `previous.cost + travelTime`, *tripling* the effective cost of any agility shortcut (or other non-teleport transport) reached via a teleport. Invisible on first-hop TransportNodes (since `previous.cost == 0` at the start node), which is why it was latent for a long time.
- **Concrete symptom:** `getNearestBank()` picked Shilo Village (actual cost 55) over Nemus Retreat (reported cost 74, true cost 40 = MoA teleport 34 + east broken-wall climb 3 + 3-tile walk). The shortcut's TransportNode was getting `g=71` instead of `g=37`.
- **Secondary fix:** the new path also sidesteps an overflow bug for plane-crossing transports with `travelTime=0` (e.g. Slayer Tower chains). `Node.cost` falls back to `WorldPointUtil.distanceBetween`, which returns `Integer.MAX_VALUE` across planes, and `previousCost + Integer.MAX_VALUE` wraps into `Integer.MIN_VALUE`-territory, making those transports look free. Passing the absolute cost directly through `Node(int packed, Node prev, int cost)` never consults `distanceBetween`.

### 2. Remove MoA audit temp debug plugin

`MoaAuditPlugin` + its `runMoaAudit` / `closeMoaWidgetIfOpen` helpers in `Rs2Walker` were supposed to stay on a side branch and got merged by accident via #1750. Pure deletion — plugin was only ever enabled manually and wrote to logs. Supersedes #1752.

### 3. Leftover /simplify findings from PR #1750 review

- `CollisionMap` MoA debug log now reports actual per-edge costs instead of hardcoded `+ 4`.
- `Rs2Walker.findMoaWidget` uses token-set membership instead of substring match — avoids false positives like token `log` hitting a hypothetical `logstrum` label.
- Demoted two `[MoA]` `log.info` calls to `log.debug` (entry + destination selection) — noisy in normal operation.
- `Rs2Walker.collectMoaChildren` helper inlined into its two callers so the client-thread guardrail scanner sees the `runOnClientThreadOptional` wrapper.
- Regenerated `client-thread-guardrail-baseline.txt` — Rs2Walker edits shifted pre-existing lambda indices; no new offenders introduced.

### Diagnosis trail for the TransportNode bug

1. Instrumented `Pathfinder`/`CollisionMap` with per-pop logs → shortcut dest `(1389, 3309, 0)` was appearing in pending with `g=71`, not the expected `g=37`.
2. Traced `TransportNode` constructor chain → static `cost(previous, travelTime)` computes `previous.cost + travelTime`, passes that as `wait` to `Node(WorldPoint, Node, int)`, which adds `previous.cost` again via its instance `cost()` method.
3. Initial fix (just `super(point, previous, travelTime)`) compiled fine but blew up on plane-crossing transports with duration 0 — the log showed `g=-2147483615` for the Slayer Tower chain tile, which led to the absolute-cost constructor swap.

### Verification (limited — only on a League 6 account)
- Before: `go to nearest bank` from Ardougne picks Shilo Village, Nemus bank target shows cost 74 when reachable via MoA east + broken-wall shortcut.
- After: same click picks Nemus Retreat, pathfinder reaches `(1386, 3309, 0)` at `cost=40`, walker actually selects `Map of Alacrity: Varlamore - Nemus retreat (east)` and teleports.
- Plane-crossing transports (Slayer Tower chain `(3421, 3550, 0)↔(3421, 3550, 1)`) no longer produce overflowed `Integer.MIN_VALUE`-ish costs.

Only manually tested on one League 6 account (all MoA regions unlocked). Haven't confirmed main-game behavior; the TransportNode change is generic enough that other transport->transport chains (boat-after-teleport, shortcut-after-teleport) should now cost correctly too — would appreciate a second set of eyes on that.

## Test plan
- [ ] `./gradlew :client:runTest`, log in, MoA in inventory, click "go to nearest bank" from mainland → expect Nemus Retreat to be chosen and the walker to fire `Map of Alacrity: Varlamore - Nemus retreat (east)`.
- [ ] Walk to a destination that requires a non-MoA teleport followed by an agility shortcut (any region) → path cost should reflect teleport + shortcut, not teleport + 2x shortcut.
- [ ] Walk across Slayer Tower or similar plane-crossing chain → cost remains positive and sensible (no overflow).
- [ ] Smoke test non-teleport pathfinding (no transports involved) → unchanged, first-hop TransportNodes also unchanged since `previous.cost == 0`.
- [ ] Confirm `MoaAuditPlugin` is gone from the plugin list.


chsami:
@copilot resolve the merge conflicts in this pull request

chsami:
is this still relevant?

runsonmypc:
@chsami  Yes:
- The TransportNode cost fix is general: any transport→transport chain (e.g. teleport→agility shortcut, teleport→boat) is currently mispriced as 2*prev.cost + travelTime.
- MoaAuditPlugin and its runMoaAudit/closeMoaWidgetIfOpen helpers in Rs2Walker are temp debug code that slipped in via #1750 (my bad) and should be removed.
- The Rs2Keyboard focus-steal fix is independent also. Tested this manually. The fix allows keyboard inputs to not steal focus. Curious to hear your thoughts on this, is there a reason we were avoiding keyboard inputs besides this?

chsami (approved):

PR #1759, opened 2026-05-05
fix(walker): convert unbounded recursion to iterative loop

seppulcro:
Bug: Walking long paths via world map (e.g. to Lighthouse for Horror from the Deep) causes the client to freeze with cascading TimeoutExceptions when the
 walker hits a stuck condition near obstacles.
 
 Cause: processWalk() recurses at 3 call sites with no depth limit, flooding the client thread queue.
 
 Fix: Iterative while(true) loop with MAX_RECALC_ATTEMPTS=15. Returns UNREACHABLE cleanly instead of crashing.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/walker/Rs2Walker.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/walker/Rs2Walker.java
@@ -265,10 +265,39 @@ public class Rs2Walker {
         return processWalk(target, distance, 0);
     }
 
+    private static final int MAX_RECALC_ATTEMPTS = 15;
+
+    /**
+     * Checks whether the walker has exceeded its maximum allowed path recalculation attempts.
+     * If the limit is reached, clears the target and returns true to signal the caller to give up.
+     *
+     * @param attempts the current number of recalculation attempts
+     * @return true if the limit has been exceeded, false otherwise
+     */
+    private static boolean hasExceededRecalcLimit(int attempts) {
+        if (attempts >= MAX_RECALC_ATTEMPTS) {
+            log.warn("[Walker] Exceeded max recalc attempts ({}), giving up", MAX_RECALC_ATTEMPTS);
+            setTarget(null);
+            return true;
+        }
+        return false;
+    }
+
+    /**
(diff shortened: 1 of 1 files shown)


seppulcro:
Added an additional with a couple more fixes I noticed while trying to use walker

1. Early-arrive at unwalkable tiles

When the player is adjacent (dist ≤ 1) to a target tile that isn't walkable (e.g. NPC spawn tile, object, wall), return ARRIVED immediately instead of burning all 15 recalc attempts. This was the root cause of the Exceeded max recalc attempts spam during quest helper walks — the pathfinder finds a route but the
exact destination tile has a collision flag, so the walker keeps recalculating despite the player being as close as possible.

2. Preserve minimap zoom on walkMiniMap()

walkMiniMap(WorldPoint) was delegating to walkMiniMap(wp, 5) which force-resets the minimap zoom to 5.0 on every single walk click. This is detectable bot behaviour — real players don't snap their zoom level 50 times a minute. The no-arg overload now reads the current minimap position directly without touching
zoom.

seppulcro:
Closing this — all three changes in this PR have been independently implemented on `development`:
 
 1. **Recursion → iteration** — landed via [`f9e1b9de04`](https://github.com/chsami/Microbot/commit/f9e1b9de04608ff6ed3f797e733aa20f842b85b6) (Sami's 
`refactor(walker): improve pathfinding and cancellation handling`) using `MAX_PROCESS_WALK_TAIL_ITERATIONS` + `for` loop, plus Leagues transport guards and
 detailed cancellation/logging.
 
 2. **Adjacent-to-unwalkable-target → arrived** — exists as `nearUnwalkableGoal` check in the arrival logic (`!Rs2Tile.isWalkable(targetLocal) && 
distanceTo <= distance → ARRIVED`).
 
 3. **Preserve minimap zoom** — `walkMiniMap(WorldPoint)` still delegates to `walkMiniMap(wp, 5)` on `development`, so this one could arguably still be 
contributed separately if it's wanted. That said, the upstream click cadence refactor 
([`dfb26e056`](https://github.com/chsami/Microbot/commit/dfb26e0569c463ab6257e91e9afcc183fd153192)) reworked minimap clicking significantly, so the 
zoom-snap may be less of an issue now.
 
 Nothing left to merge here.

PR #1779, opened 2026-05-18
Introduce Rs2Player.isIdle() and waitForXpDrop() overloads

afss0:
This PR introduces:

1. Skill unspecific waitForXpDrop overloads, that triggers when the player receives experience in any skill.
2. Rs2Player.isIdle(), to check if the player is idle (not performing any action).

The player is considered idle if all of the following are true:

- Not currently animating (animation is IDLE and no animation within the last 3000ms)
- Not in combat
- Not moving
- Not interacting with a bank, deposit box, grand exchange, or chat dialog


afss0:
ClientThreadGuardrailTest claims net.runelite.api.Client#getOverallExperience() requires to be called in the client thread, however I didn't detect any exceptions during runtime. I'll review this further.

PR #1791, opened 2026-06-03
fix(walker): four robustness fixes in Rs2Walker

mikefallen:
## Summary

- **Bound `processWalk` recursion** — the final `else` branch recursed with no counter, risking a `StackOverflow` on routes that never converge. Now retries up to 3 times then returns `WalkerState.UNREACHABLE`, matching the existing `partialPath` retry pattern directly above it.

- **Guard invalid canvas in `walkFastLocal`** — if `Perspective.localToCanvas` returns `null` the coords fall back to `-1/-1`, producing a bogus `WALK` menu entry. Now aborts early with a warning log instead of dispatching the bad click.

- **Match pickaxes by item ID in motherlode rockfall check** — substring matching on `"pickaxe"` is fragile: false positives on handle/head items, breaks if display names gain suffixes. Replaced with a `PICKAXE_IDS` constant covering all wieldable pickaxes via `ItemID`, using the existing `hasItem(int...)`/`isWearing(int...)` overloads.

- **Skip stuck-detection while in dialogue or bank** — standing still during an NPC dialogue or open bank incremented `stuckCount` and the 10-second timer, causing the walker to fire a random minimap click or recalculate the path mid-interaction. Both counters are now reset while `Rs2Dialogue.isInDialogue()` or `Rs2Bank.isOpen()` is true.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/walker/Rs2Walker.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/walker/Rs2Walker.java
@@ -86,6 +86,14 @@ public class Rs2Walker {
 
     public static boolean disableTeleports = false;
 
+    private static final int[] PICKAXE_IDS = {
+        ItemID.BRONZE_PICKAXE, ItemID.IRON_PICKAXE, ItemID.STEEL_PICKAXE, ItemID.BLACK_PICKAXE,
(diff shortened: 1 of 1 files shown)


mikefallen:
Closing — this work is superseded by the walker rewrite already in `development`. `processWalk` was converted from recursion to a for-loop upstream (the stuck/partial-retry paths now `continue` instead of `return processWalk(...)`), which eliminates the unbounded-recursion path raised in review and covers the same robustness themes (stall recalc, partial-retry, off-path, unreachable) more thoroughly via the new `Telemetry` and Leagues-aware handling. Rebasing onto `development` conflicts in every region with the upstream version taking precedence, leaving nothing unique to salvage.

PR #1792, opened 2026-06-05
fix(webwalker): traverse far gates/doors and recover from smoothed-path dead-ends

oversized93:
## Summary

Two related robustness fixes for the web walker (`Rs2Walker`), both reproduced and verified live in-game on a fenced gate route, plus unit tests. Both bugs share one root: the walker trusted a geometric shortcut over the actual route.

## Bug 1 — stuck behind a far gate/door

Clicking a gate/door several tiles away makes the OSRS server walk the player all the way there and open it in one action. The walker was (a) firing recovery/minimap clicks that **cancelled** that walk-to-open, and (b) calling `markStationaryDoorOpened(...)` on a **non-traversal**, suppressing re-interaction for `STATIONARY_DOOR_SUPPRESS_MS` (10s) — so the player parked against a still-closed gate and never retried.

**Fix:** a distance-scaled *walking-to-door* state. While active it keeps that door interactable (overrides the stationary-opened suppression for it) and yields recovery clicks until the player arrives or a deadline / no-progress timeout elapses. The door is only suppressed on actual traversal; the state self-clears on traversal, timeout, or route change.

## Bug 2 — "player just doesn't move" → idle logout

The smoothed walkable path can dead-end against a barrier that must be crossed from the far side (e.g. a closed gate). The unreachable-tile recovery only scanned 5 tiles for a tile *Euclidean-closer to the target*, which can't follow the winding route to the crossing — so the walker livelocked (`tile-unreachable-near-player` + `stall_recalc` indefinitely) until the character idle-logged-out.

**Fix:** raw-path recovery. When the smoothed next tile is unreachable and there's no openable obstacle nearby, fall back to the **raw** path (`findFurthestReachableRawPathPoint`) and walk to the furthest reachable raw-path tile — advancing the player to the actual crossing tile, where existing door handling opens it.

## Verification

- **Live** on a closed fenced gate: the walk now logs `unreachable-raw-path-recovery` ×2 → gate opens → arrives, with `tile-unreachable-near-player = 0` and `stall_recalc = 0` (previously ~16s of thrash / livelock to logout).
- **Unit:** 65 walker tests pass, including new tests for the pure helpers `walkToDoorDeadlineMs`, `walkToDoorWindowOpen`, `doorTileNearSegment`, `doorApproachDistance`.

## Scope note

This does not add a generic "give up → UNREACHABLE" loop-breaker, so a *genuinely* unreachable destination (no valid route at all) could still livelock — a separate, narrower follow-up.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/walker/Rs2Walker.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/walker/Rs2Walker.java
@@ -141,6 +141,16 @@ public class Rs2Walker {
     private static final long DOOR_EDGE_SKIP_COOLDOWN_MS = 700L;
     private static final long RECOVERY_MOVEMENT_IN_FLIGHT_MS = 1400L;
     private static final long DOOR_TRAVERSAL_RECOVERY_BLOCK_MS = 2_200L;
+    // "Walking to door" state. Clicking a gate/door several tiles away makes the OSRS server walk
+    // the player all the way to it and open it in one action. While that approach is in flight the
+    // walker must NOT fire recovery/minimap clicks (they cancel the walk-to-open and strand the
+    // player against a closed gate) and must keep the door interactable instead of suppressing it
+    // as "opened". The deadline scales with the click distance so a far gate gets time to arrive.
+    private static final int WALK_TO_DOOR_ADJACENT_TILES = 2;
+    private static final long WALK_TO_DOOR_BASE_MS = 2_000L;
+    private static final long WALK_TO_DOOR_PER_TILE_MS = 700L;
+    private static final long WALK_TO_DOOR_MAX_MS = 12_000L;
+    private static final long WALK_TO_DOOR_NO_PROGRESS_MS = 3_000L;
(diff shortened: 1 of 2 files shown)


chsami:
@oversized93 this PR is currently not mergeable. GitHub reports it as `CONFLICTING` / `DIRTY` against `development`, and the `Build` check is failing.

Please update the branch with the latest `development`, resolve the conflicts, and rerun/fix the build so it can be checked again.

chsami:
Closing this as superseded by #1817 and #1827.

The original far-door cancellation and raw-path dead-end reports were valuable, but current development now contains the newer modular door probing/await flow, route-state ownership, forward raw-path recovery, and regression coverage merged in #1827 (merge commit e555b91da730e7d7f8154b79d0c89ae4ecb8bc94). That implementation also avoids carrying the unresolved stale walkingToDoor state called out in this PR review.

This branch now conflicts with development and its CI fails the client-thread guardrail, so merging or rebasing this older implementation would reintroduce overlapping walker logic. Thanks for the live reproduction and analysis that helped shape the later fixes.

PR #1795, opened 2026-06-09
fix(antiban): prevent action-cooldown stall when play-style tick intervals cross

pjmarz:
## Summary

With "Use Non-Linear Intervals" enabled (which the per-skill antiban setup templates turn on by default), `PlayStyle.evolvePlayStyle()` random-walks `primaryTickInterval` and `secondaryTickInterval` independently, so over time they can cross (primary greater than secondary). Once that happens, `PlayStyle.getRandomTickInterval()` calls `ThreadLocalRandom.nextInt(origin, bound)` with origin above bound and throws on every stat change:

```
java.lang.IllegalArgumentException: bound must be greater than origin
    at java.base/java.util.concurrent.ThreadLocalRandom.nextInt(ThreadLocalRandom.java:337)
    at net.runelite.client.plugins.microbot.util.antiban.enums.PlayStyle.getRandomTickInterval(PlayStyle.java:114)
    at net.runelite.client.plugins.microbot.util.antiban.Rs2Antiban.performActionCooldown(Rs2Antiban.java:297)
    at net.runelite.client.plugins.microbot.util.antiban.Rs2Antiban.actionCooldown(Rs2Antiban.java:277)
    at net.runelite.client.plugins.microbot.util.antiban.AntibanPlugin.onStatChanged(AntibanPlugin.java:345)
```

Captured live on 2.6.5 while high alching.

The impact is worse than the exception: `performActionCooldown()` sets `Microbot.pauseAllScripts = true` BEFORE computing the interval, and `actionCooldownActive = true` only after. The throw lands between the two, so the cooldown clearer (`AntibanPlugin.performActionBreak`, gated on `actionCooldownActive`) never runs and every running script stays frozen until the interval random-walk happens to uncross (observed multi-minute stalls). Any script using a templated antiban setup is exposed.

## Fix

1. `PlayStyle.getRandomTickInterval()`: order the bounds with min/max before calling `nextInt(lo, hi + 1)`, so crossed intervals can never throw.
2. `Rs2Antiban.performActionCooldown()`: compute the interval first, then set `pauseAllScripts` and `actionCooldownActive` together, so no exception path can leave scripts paused with nothing to release them.
3. New regression test `PlayStyleRandomTickIntervalTest` forces the crossed-interval state (which deterministically threw before the fix) plus the equal-bounds edge case.

## Testing

- `./gradlew :client:runUnitTests --tests "*PlayStyleRandomTickIntervalTest"` passes (failed before the fix).
- `./gradlew :client:compileJava` green.
- Patched client boots and runs clean; normal action-cooldown pause and resume behavior unchanged.
- Workaround for affected users until this lands: disable "Use Non-Linear Intervals" in the Antiban panel.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/antiban/Rs2Antiban.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/antiban/Rs2Antiban.java
@@ -295,8 +295,9 @@ public class Rs2Antiban {
         else
             TIMEOUT = playStyle.getPrimaryTickInterval();
 
-        // Pause and set the active flag together, only after TIMEOUT is computed. If computing the
-        // interval ever throws, scripts must not be left paused with no countdown to clear them.
+        // The pause (universal antiban only) and the always-set active flag both happen only after
+        // TIMEOUT is computed: if computing the interval ever throws, scripts must not be left
+        // paused with no countdown to clear them.
         if (Rs2AntibanSettings.universalAntiban)
 			Microbot.pauseAllScripts.compareAndSet(false, true);
 
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/antiban/Rs2Antiban.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/antiban/Rs2Antiban.java
@@ -287,9 +287,6 @@ public class Rs2Antiban {
     }
 
     private static void performActionCooldown() {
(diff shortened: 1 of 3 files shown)


pjmarz:
Addressed: the inclusive upper bound is computed in long so Integer.MAX_VALUE cannot overflow, the test pins PlayStyle.MODERATE instead of values()[0], and the reorder comment now describes the conditional pause plus unconditional flag precisely. On the enum-state mutation concern: the intervals are restored in a finally block and the unit-test suite runs single-threaded; noted in the test.

PR #1813, opened 2026-07-02
fix: block Al Kharid mine in pathfinder collision data

Zimada1:
## Problem
The shortest-path A* pathfinder routes through the Al Kharid mine open pit because the cache-derived collision data doesn't mark the terrain as blocking. This causes the walker to get stuck — the path goes green then recalculates to red but the character never moves.

## Fix
Add 88 blocked edges around the mine perimeter (x=3281-3300, y=3151-3178) to locked_edges.tsv. These edge-level blocks prevent the pathfinder from routing through the mine area, forcing it to find an alternate surface route.

## Testing
Tested by walking through the Al Kharid area — pathfinder now correctly routes around the mine instead of through it.


chsami:
@Zimada1 this PR is currently not mergeable. GitHub reports it as `CONFLICTING` / `DIRTY` against `main`.

Please update the branch with the latest `main`, resolve the conflicts, and push the fixed branch so it can be checked again.

chsami:
Thanks for identifying and live-testing the Al Kharid mine collision gap. I preserved the intended blocked-edge fix in focused replacement PR #1829, added exhaustive bidirectional perimeter coverage and an end-to-end route regression, and based it on the current development branch. This PR also includes an obsolete RuneLite snapshot/generated-file update, targets main, and now conflicts with current development, so I’m closing it in favor of #1829.

PR #1833, opened 2026-08-05
Add optional long and mega breaks to Break Handler V2

MuffinKid23:
## Summary
- add optional long-break and mega-break timing configuration to Break Handler V2
- keep normal breaks independent from pending long/mega timers
- merge due long/mega timers into an already-active break instead of starting competing break cycles
- prioritize mega breaks when long and mega timers coincide
- show runtime, break count, long-break countdown, mega-break countdown, and active break type in the overlay

## Test
- ./gradlew.bat :client:compileJava -x test

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/breakhandler/breakhandlerv2/BreakHandlerV2Config.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/breakhandler/breakhandlerv2/BreakHandlerV2Config.java
@@ -66,6 +66,124 @@ public interface BreakHandlerV2Config extends Config {
         return 15;
     }
 
+    @ConfigItem(
+        keyName = "enableLongBreaks",
+        name = "Enable Long Breaks",
+        description = "Enable a second independent long-break timer",
+        position = 4,
+        section = breakTimingSettings
+    )
+    default boolean enableLongBreaks() {
+        return false;
+    }
+
+    @ConfigItem(
+        keyName = "minLongBreakInterval",
+        name = "Min Long Break Interval (minutes)",
+        description = "Minimum time to play before a long break can trigger",
+        position = 5,
+        section = breakTimingSettings
+    )
+    @Range(min = 1, max = 600)
+    default int minLongBreakInterval() {
+        return 120;
+    }
+
+    @ConfigItem(
+        keyName = "maxLongBreakInterval",
+        name = "Max Long Break Interval (minutes)",
+        description = "Maximum time to play before a long break can trigger",
+        position = 6,
+        section = breakTimingSettings
+    )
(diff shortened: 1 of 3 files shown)


chsami:
Thanks for the contribution. Before this is ready to merge, please address the following:

1. Retarget the PR from `main` to `development` so it goes through the repository's integration branch and CI.
2. Resolve the two open review findings:
   - The default long-break interval currently occurs before the normal playtime window, so enabling long breaks can continually preempt normal breaks. Adjust the defaults or preserve the normal-break deadline.
   - `currentBreakIsLong` is written by the script/scheduler path and read by the overlay thread without guaranteed visibility. Make it `volatile` or synchronize both access paths.
3. Re-run CI after updating the branch and confirm the normal build passes.

Once those are addressed, we can review it again for merge.

MuffinKid23:
Addressed the review feedback:

- Retargeted the PR from main to development.
- Updated Break Handler V2 defaults so normal breaks run before long breaks by default: normal playtime 7-15 minutes, normal duration 2-3 minutes, long interval 20-30 minutes, long duration 8-11 minutes, mega interval 120-180 minutes, mega duration 20-30 minutes.
- Made the current break-type flags read by the overlay olatile for cross-thread visibility.
- Re-ran ./gradlew.bat :client:compileJava -x test locally and it passed.

chsami:
Thanks for the update. The `volatile` finding is fixed, but this still is not merge-ready for two reasons:

1. The timing fix changes the existing normal defaults from **45–90 minutes / 5–15 minutes** to **7–15 minutes / 2–3 minutes** just to make the 20–30 minute long-break timer occur later. That is a large behavior change for every existing default user. Please preserve the current normal defaults and either move the long-break default interval beyond the normal window or preserve the normal-break deadline when a long break completes.
2. The PR currently contains 31 unrelated release/generated files (`gradle.properties`, legacy/gameval IDs, and many `rs2asm`/hash files). Please rebase onto current `development` or cherry-pick/squash only the three Break Handler V2 files onto a fresh branch from `development`.

After the scope is clean and the timing behavior is corrected, please let the repository Build check run; the current PR still has no Build result.

MuffinKid23:
Addressed the latest review feedback:

- Rebuilt/force-pushed the branch from current development with a clean scope: only the three Break Handler V2 files are changed.
- Preserved the existing normal defaults: 45-90 minutes playtime and 5-15 minutes break duration.
- Made long/mega breaks opt-in by default, and moved their default intervals beyond the normal window:
  - long: 120-180 minutes, 8-11 minutes duration
  - mega: 240-360 minutes, 20-30 minutes duration
- Kept the break type flags volatile.
- Ran ./gradlew.bat :client:compileJava -x test --no-daemon locally and it passed.

PR #1837, opened 2026-08-12
Hotfix/webwalker fixes

dylanreniers:
Applied fixes:

1. Persistent Single-Thread Walk Task Execution (ShortestPathScript.java):
- Eliminates thread restarts and pathfinder cancellations, keeping movement active on a single dedicated background thread.

2. Fixing False Route Failures in Banked Walking (Rs2Walker.java):
- Preserves active movement states (MOVING) when walking with banked transports, preventing self-cancellation.

3. Humanized Run Energy Recovery (Rs2Player.java):
- Waits for run energy to recover to a randomized threshold (15%–35%) before toggling run back on, eliminating robotic 1% toggles.

4. Camera Settling for Door Interaction Accuracy (Rs2Walker.java):
- Waits for camera movement to settle after turning towards off-screen doors before computing 3D canvas clickboxes, ensuring 100% click accuracy.

5. Fast Miss Recovery & Reduced Door Wait (Rs2WalkerAwaits.java):
- Detects stationary state after 1200ms (2 game ticks) on missed clicks and releases immediately for a fast retry, eliminating 5-second idle stalls.

6. Realistic Physical Mouse Click Hold Duration (VirtualMouse.java):
- Incorporates a 40–90ms hold duration between button press and release.

7. Short-Distance 3D Scene Clicks (Rs2Walker.java):
- Clicks directly on the 3D ground canvas when the destination is within 4 tiles (a couple of steps) and visible on screen, while preserving minimap navigation for longer journeys.


chsami:
I’m holding this hotfix from merge. The changes are broad enough (walker task lifecycle, pathfinder fallback, door timing, canvas/mouse input, run-energy behavior, Stronghold answers, and transport data) that they need the normal integration path and runtime evidence.

Please address the following:

1. Retarget from `main` to `development`; the PR currently has no repository Build check.
2. In `ShortestPathScript`, replace the hardcoded `10` passed to `walkWithBankedTransportsAndState` with the same configured/default reached-distance used by the non-banked branch.
3. Update `Rs2Player.toggleRunEnergy`’s return contract: it now also returns `false` when energy is below the randomized activation threshold, not only when the widget is missing.
4. Remove the trailing whitespace on the three new `transports.tsv` rows (`git diff --check` currently fails).
5. Add or report focused tests/live walks for target replacement/cancellation, unreachable-plane recovery, short canvas clicks, door misses, and the Stronghold/transport additions. The current body also says the idle threshold is 600 ms, while the latest code is back at 1200 ms, so please make the description match the intended behavior.

These are substantial walker/input changes, so I’m leaving them with the author rather than editing and merging them without runtime validation.

chsami (changes requested):
Re-reviewed current head `31b263b` after the follow-up fixes. The reached-distance and TSV-whitespace items from the earlier comment are addressed, but this is still blocked by current integration and correctness issues:

1. The branch is 18 commits behind `development`, has 15 branch-only commits, and conflicts in both `VirtualMouse` and `Rs2Walker`. The diff still carries generated RuneLite API/rs2asm/version updates unrelated to the seven walker/input changes in the PR body. Please rebuild/rebase the feature commits onto current `development` and resolve them against the newer input-arbiter and walker implementations; this is not a safe mechanical maintainer merge.

2. CI is red on `ClientThreadGuardrailTest`. The log reports a new direct `TileObject#getWorldLocation()` read in `Rs2Walker.handleStrongholdOfSecurityAnswer`, plus 14 baseline entries whose lambda identifiers shifted. Please put the new object read on the client thread, then regenerate/verify the baseline only after the rebase so it records real call sites rather than masking the new violation.

3. This head changes `UNREACHABLE_DOOR_RECOVERY_BACKTRACK_EDGES` from 2 to 0 and changes the raw door scans to begin at `rawStart`. That contradicts the current movement invariant: when a player is slightly off-path near a gate, the closest raw index can already be beyond the blocking edge, so scanners must include one or two preceding raw edges. Please restore the backtrack or provide focused regression/live evidence proving the Lumbridge-farm/nearby-gate class of route remains handled.

The global mouse press delay and run-energy behavior also need to be reconciled and runtime-tested on the current input/walker stack after the conflicts are resolved. I have not attempted to rewrite this broad branch from the maintainer side.

PR #1838, opened 2026-08-14
Walker overhaul: decomposed executor, door ledger, planner-selection gate

infuse21:
## What

The walker as it runs on my fork today, delivered as **one clean commit cut directly from the current \`development\` tip** (post-revert of #1832), so it auto-merges with no unrelated history attached.

- **Decomposed executor** — the former 14k-line \`Rs2Walker\` split into a route loop + public API (\`Rs2Walker\`), transport dispatch/handlers (\`Rs2WalkerTransports\`), the door cascade (\`Rs2WalkerDoors\`), and movement/click issuance (\`Rs2WalkerMovement\`). Every move verified behaviour-identical (full suite + live walks at each step).
- **Door handling rework** — a single \`DoorAttemptLedger\` owning all door state (attempt cooldowns, strikes, settle windows, per-pass claims), a unified route-door classifier replacing four contradictory rules, crossed-face release (door wait releases when the collision edge opens, not when the player is through), goal-object and double-gate-wing guards.
- **Route execution fixes** — fold-stall fix (multi-gate corridors no longer stall 4–26s per gate), WalkExit taxonomy with correct progress classification, oscillation-proof no-progress budgets, stuck-recovery that respects route order.
- **Planner-selection gate** — vendored upstream shortest-path planner as a pinned source set (\`src/upstreamPlanner\`) with comparison harness tasks, for convergence work against Skretzo's planner.
- **Refreshed transport data** (\`transports.tsv\`, \`blocked_edges.tsv\`, and friends) and the walker test suite: decision tables, route corpus, session-reset pins, and the client-thread guardrail baseline regenerated for this tree.

Support-file changes are limited to what the walker compiles against: \`Rs2Tile\` (edge passability), \`Rs2Magic\`/\`Rs2Staff\`/\`Rs2Tome\` (cast + rune providers), \`Rs2LeaguesTransport\` injection, and the client build script (source set + verification tasks).

## Why

The walker's runtime executor was the failure source (path generation was healthy). This is several weeks of log-driven fixes, each verified on live walks — multi-gate corridors that took minutes now run in ~90–120s; long cross-map walks (e.g. Lumbridge→Varlamore, 2 gates + 2 NPC travels) complete clean with zero recovery interventions.

## Notes for review

[...]

PR #1839, opened 2026-08-14
Transport data refresh (walker split 1/4)

infuse21:
## What

Part **1 of 4** splitting #1838 into reviewable PRs, per review feedback there. This part is the transport **data refresh only** — 15 files, zero code.

- Coverage and coordinate corrections found by live walks: `agility_shortcuts`, `canoes`, `minecarts`, `quetzals`, `ships`, `transports`, `teleportation_items/spells/minigames`.
- **Duration corrections**: spirit tree rows had a blank Duration column, and non-teleport transports get no minimum — so every spirit tree edge cost the planner exactly **zero**, and it picked them even when a short walk was faster (user-reported). Spirit trees blank→10, gnome gliders 8→12, magic mushtrees blank→6, proportionate to the existing convention (fairy ring 5, quetzal 6, charter 6, boats 9, ships 10).
- `blocked_edges`/`restrictions`: doors that never open and edges the live game refuses.
- `collision-map.zip` regenerated for newer map areas.

## Review notes

- Every file parses with the loader **already on `development`** — unknown header columns are ignored by the parser, so no code rides along.
- Full unit suite (`:client:runUnitTests`) green on this branch.
- The remaining parts: (2) pinned upstream-planner source set + gradle wiring, (3) shortestpath engine, (4) walker executor — opening sequentially as agreed.


infuse21:
Note on the CodeRabbit skip above: it's the repo's own path filters (`!**/*.tsv`, `!**/*.zip`) — data files are excluded from automated review by configuration, so this PR will never receive a CodeRabbit pass regardless of size. The review surface here is the 15-file data diff itself plus the green unit suite, which exercises the loader over every changed file. The code-bearing parts of the split (2–4) are within CodeRabbit's scope.

infuse21:
i have found some issues that need to be addressed before the pr happens . i have set to draft for now.

PR #1842, opened 2026-08-20
fix(shortestpath): refresh transport data and reopen Al Kharid Palace

infuse21:
## Summary

- refresh shortest-path transport and collision data
- remove the erroneous blocked-edge perimeter that sealed Al Kharid Palace
- add regressions proving routes enter the palace from the north and finish at the requested interior tiles

This is PR 1 of 6. Each PR is kept below 100 changed files for CodeRabbit. The remaining slices will be rebased and opened sequentially after their predecessor merges so earlier files are not counted again.

## Validation

- `:client:runUnitTests --tests net.runelite.client.plugins.microbot.shortestpath.ShortestPathCoreTest`
- verified in game: the walker now enters Al Kharid Palace

## Scope

16 changed files relative to `development`.


chsami (changes requested):
Blocking current head `bfa0fe4` after testing a clean synthetic merge with the latest `development`:

1. The refreshed TSVs introduce unsupported `Item IDs` syntax, and the loader silently skips the affected transports. `Transport` currently parses numeric IDs separated by the existing delimiters, but this head adds values such as `13280=1||13342=1`, `29271=1||29273=1||29275=1`, `33120=1`, `CROSSBOW=1&MITH_GRAPPLE=1`, `ROPE=1`, `2887=1`, and `952=1`. Running `ShortestPathCoreTest` logs `Skipping transport row ... For input string ...` for the new max/quest cape, quetzal whistle, grapple/rope, Elemental Workshop, and Barrows rows, so those advertised additions never enter the graph. Please either translate the data to the supported numeric grammar or extend the parser with explicit AND/OR/quantity semantics, and add assertions that the newly added transports actually load; `testTransportLoadingDoesNotThrow` does not catch this because row parse failures are swallowed and logged.

2. The PR removes the Al Kharid mine perimeter edges and deletes the regression that required routes to avoid the open pit. I restored an equivalent test locally against this head's collision map (start `3280,3164,0`, target `3305,3164,0`, pit bounds `3281..3300 x 3151..3178`); it fails because the computed route enters the pit. Please restore/model the mine boundary without resealing the palace, and retain both regressions: palace access and mine avoidance.

Validation performed on the synthetic merge: `:client:compileJava` and the submitted `ShortestPathCoreTest` pass, `collision-map.zip` passes `unzip -t`, but the restored mine-avoidance probe fails as described. These are data-model/path-correctness issues across several files, so I have not attempted a maintainer-side patch.

infuse21:
Addressed the requested transport-data compatibility issue in `545b3a3cc1`:

- Converted every new Item IDs expression to the legacy numeric grammar the current parser supports (including the OR alternatives).
- Added `testReviewedItemRequirementTransportsActuallyLoad`, which asserts the reviewed grapple, rope, cape, whistle, key, and spade rows are present with their expected item IDs.
- The loader now reports 6,353 transports with no row-skip warnings in `ShortestPathCoreTest`.
- Verified the PR merges cleanly onto current `development` (`dcbe7dca0a`), the focused test class passes there, and `collision-map.zip` is readable.

On the requested “Al Kharid mine perimeter” restoration: the removed rectangle (`x=3281..3300, y=3151..3178`) contains both palace destinations used by the regression (`3293,3163` and `3298,3163`). Restoring it would therefore reintroduce the palace bug. RuneLite’s own clue data places Al Kharid mine around `(3300,3291)`, `(3299,3289)`, and `(3303,3271)`, roughly 100+ tiles north:

- https://github.com/runelite/runelite/blob/master/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/MapClue.java
- https://github.com/runelite/runelite/blob/master/runelite-client/src/main/java/net/runelite/client/plugins/cluescrolls/clues/EmoteClue.java

I added `testRemovedPerimeterCoordinatesDescribeThePalaceNotTheMine` to lock down that coordinate distinction alongside the existing palace-entry path regression. Because the old rectangle necessarily blocks valid palace targets and does not cover the canonical mine coordinates, I have not restored it.

chsami (approved):
Re-reviewed head 32a4f6ae59 after applying the four transport compatibility/execution fixes. Canoe axe gates and Quetzal varplayer gates now load, Dougne canoes select their own interface, and exact Barrows mound entries execute via spade. Focused walker/shortest-path tests, the client-thread guardrail, the full 1,320-test client unit suite, and a login-screen startup/transport-loader smoke test all pass. Fresh CI and CodeRabbit are green.

PR #1852, opened 2026-09-03
fix(grandexchange): bound quantity entry retries

RLitePlus:
### Problem

Grand Exchange offer processing could stall indefinitely when the custom quantity button was temporarily unavailable because that retry path bypassed the existing attempt limit.

### Solution

Limit quantity entry to three attempts, stop immediately once the requested quantity is observed, and return failure cleanly when the attempts are exhausted.

### Additional notes

- Local regression tests passed for unavailable-button exhaustion and successful quantity entry.
- Compilation and checkstyle passed.
- Test files are not included in this pull request.
- No version bump is required for this core utility fix.


RLitePlus:
no

PR #1855, opened 2026-09-05
fix(grandexchange): bound quantity entry retries

RLitePlus:
### Problem

Grand Exchange offer processing could stall indefinitely when the custom quantity button was unavailable because that retry path bypassed the attempt limit. Immediate retries could also exhaust the limit before a temporarily missing button appeared.

### Solution

Limit quantity entry to three attempts and wait up to two seconds for the quantity button on each attempt. Fetch and use the button after the wait succeeds, including on the final attempt. Stop once the requested quantity is observed; close the exchange and return failure when attempts are exhausted.

### Additional notes

- Replaces #1852, which could not be reopened after the original fork visibility changed.
- Compilation and Checkstyle passed (`:client:compileJava :client:checkstyleMain`); `git diff --check` passed.
- Five local regression scenarios passed using the extracted production methods with simulated widget timing: delayed availability, final-attempt recovery, bounded exhaustion, immediate availability, and already-correct quantity.
- The regression harness remains local. Live game behavior was not tested.
- No version bump is required for this core utility fix.


chsami:
Pushed a narrow guardrail-baseline correction: the 12 existing GE entries now follow their moved method/lambda locations. The API calls and number of exemptions are unchanged. Client compilation passed; the full local unit run had 1,502 tests with only this guardrail failure (2 skipped), and the corrected guardrail passes. Waiting for CI on the new head before considering merge.

PR #1857, opened 2026-09-06
Fix world hopper TCP ping fallback

infuse21:
World Hopper currently treats TCP port 43594 as the only endpoint and does not bound the socket connect operation. When that endpoint is unavailable, reachable worlds are reported as failed pings.

This change adds a bounded connect timeout and tries port 443 when 43594 fails. `Rs2WorldUtil` now uses the same shared ping logic so world selection and World Hopper report reachability consistently. A regression test covers falling back when the first endpoint refuses the connection.


infuse21:
@coderabbitai review

PR #1859, opened 2026-09-07
Prefer scene clicks for webwalker routes

MuffinKid23:
## Summary
- prefer scene/canvas walk clicks for route-backed webwalker movement when the target tile is visible
- keep minimap clicks as fallback for off-screen or clipped route targets
- reduce active-route idle nudge timing so stalled routes resume sooner
- update first movement telemetry from `first_minimap_click` to `first_route_click`

## Testing
- `./gradlew.bat :client:compileJava --no-daemon --console=plain`
- manually verified locally with a custom jar build

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/walker/Rs2Walker.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/util/walker/Rs2Walker.java
@@ -259,8 +259,8 @@ public class Rs2Walker {
      * genuinely still (not moving/animating/interacting) on the same tile, and
      * {@link #ACTIVE_ROUTE_IDLE_NUDGE_COOLDOWN_MS} still prevents click spam.
      */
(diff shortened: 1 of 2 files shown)


chsami (changes requested):
P1 — CI fails on the new `Rs2WalkerMovement.isSceneCanvasClickable` client/world-view reads (run 34168386578). These execute on the walker worker, so viewport/projection state needs a client-thread snapshot before deciding whether to click.

I tried the localized client-thread wrapper in an isolated worktree: compilation and all 183 walker tests passed, but `ClientThreadGuardrailTest` then detected six additional `Perspective.localToCanvas` callers in Rs2Camera, Rs2UiHelper, Rs2Walker, and Rs2WalkerMovement. I have not pushed that incomplete fix. Please make the projection boundary consistently client-thread-safe without merely adding baseline exemptions, then run `buildAll` and `:client:runUnitTests`.

PR #1861, opened 2026-09-08
Improve webwalker camera and route continuity

MuffinKid23:
This replaces #1859 with the complete scene-click, camera-following, and route-continuation changes.

The design goal is less static movement: prefer the farthest visible, reachable tile along the planned route to cover more ground with fewer clicks. Door, transport, collision, and arrival handling still take priority.

- Prefer forward scene targets within a bounded 32-tile scan. Stop at blocked edges, doors, and planned transports; require the click point to fit inside the viewport.
- Occasionally use the minimap after 2-7 successful scene clicks. Failed attempts do not advance the counter, and successful minimap fallbacks reset it. Scene clicks remain the primary choice.
- Follow the direction of travel using the existing arrow-key helpers at the client's configured camera speed, without snapping the camera.
- Vary pitch across 210-325 legacy camera units, selecting a target at least 24 units from the current inclination. This includes intermediate views without the previous high overhead angle. Refresh variation every 5-10 seconds while walking, with lateral offsets bounded to 12 degrees.
- Keep vertical adjustment active even when yaw is already aligned. Release held keys on completion, cancellation, logout, plane changes, errors, or timeout.
- Prepare the next route click with 6 tiles remaining while walking or 8 while running, retain the 900 ms retarget cooldown, and reduce movement polling to 200 ms. Track the actual selected destination after issuing a click.
- Batch bounded scene selection on the client thread. Protect viewport projection reads without changing the thread-safety baseline or suppressing checks.

The interaction-handoff update also includes Rs2Bank, a reusable read-only Rs2InteractionApproach helper, and focused tests. Banking can stop coordinate walking when a nearby visible bank object is reachable, then interact after the walker returns. Same-plane, screen visibility, and collision-aware approach-distance checks prevent handing off through closed doors. Banks without an eligible object retain the normal opening fallback.

Scene clicks may choose among the final three eligible forward tiles, while retaining the existing collision, door, and transport boundaries. The pathfinder's existing equal-cost route tie-breaking is unchanged; this does not force longer alternative routes.

No generated files, client version changes, logs, account details, or unrelated plugins are included.

Portal Nexus control: add an enabled-by-default `Use Portal Nexus` setting, subordinate to `Use Player-owned-house Teleports`. Exclude only Nexus destinations when disabled, keep other POH transports and saved selections intact, and include the setting in the transport-refresh cache key. The existing config-change handler invalidates and recalculates routes for this `use...` option. Tests cover default compatibility and cache-key invalidation.

The contributor has manually tested and approved the equivalent compatible client build, including the latest camera tuning and minimap cadence. Local validation results are recorded in the update comment; GitHub CI remains authoritative for the Linux build.

### DIFF
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/shortestpath/PohPanel.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/shortestpath/PohPanel.java
@@ -204,6 +204,11 @@ public class PohPanel extends PluginPanel {
      * @return allTransports map with all cached PoH transports added in
      */
     public static Map<WorldPoint, Set<Transport>> getAvailableTransports(Map<WorldPoint, Set<Transport>> allTransports) {
+        return getAvailableTransports(allTransports, true);
+    }
+
+    public static Map<WorldPoint, Set<Transport>> getAvailableTransports(
+            Map<WorldPoint, Set<Transport>> allTransports, boolean usePortalNexus) {
         if (instance == null) return allTransports;
         Set<PohTeleport> pohTeleports = new HashSet<>();
         Map<WorldPoint, Set<Transport>> pohTransports = new HashMap<>();
@@ -215,7 +220,9 @@ public class PohPanel extends PluginPanel {
 
         pohTeleports.addAll(instance.checkboxPanel.getTeleports());
         pohTeleports.addAll(instance.portalPanel.getTeleports());
-        pohTeleports.addAll(instance.nexusPanel.getTeleports());
+        if (usePortalNexus) {
+            pohTeleports.addAll(instance.nexusPanel.getTeleports());
+        }
         pohTeleports.addAll(instance.jewelleryBoxPanel.getTeleports());
 
         if (instance.checkboxPanel.fairyRingCb.isSelected()) {
--- a/runelite-client/src/main/java/net/runelite/client/plugins/microbot/shortestpath/ShortestPathConfig.java
+++ b/runelite-client/src/main/java/net/runelite/client/plugins/microbot/shortestpath/ShortestPathConfig.java
@@ -408,6 +408,17 @@ public interface ShortestPathConfig extends Config {
     }
 
     @ConfigItem(
+            keyName = "usePortalNexus",
+            name = "Use Portal Nexus",
+            description = "Include Portal Nexus destinations saved in PoH Web Config. Requires Player-owned-house Teleports.",
+            position = 24,
+            section = sectionSettings
+    )
+    default boolean usePortalNexus() {
+        return true;
+    }
+
+    @ConfigItem(
             keyName = "cancelInstead",
             name = "Cancel instead of recalculating",
             description = "Whether the path should be cancelled rather than recalculated " +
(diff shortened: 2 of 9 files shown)


MuffinKid23:
Follow-up: route clicks now alternate occasionally between direct scene tiles and the minimap. Direct scene clicks remain the default; every 4-7 route clicks the walker attempts the minimap, then falls back to the scene route if the point is outside the minimap or unavailable. The cadence resets for each active route, so it does not switch methods on every click or introduce artificial pauses. This supports the PR's goal of less static, more human-like route interaction while preserving planned path, door, transport, and collision handling.

MuffinKid23:
Adjustment: the alternation is now probabilistic rather than cadence-based. Each route click has a 30% chance to try the minimap first and a 70% chance to use the direct scene tile first. If the minimap click is unavailable, the direct scene route remains the immediate fallback; no fixed 4-7 click cadence is used.

MuffinKid23:
Final adjustment: restored the cadence-based behavior. The walker now attempts the minimap after a randomized 4-7 direct route clicks, then resets the cadence. Direct scene tiles remain the normal priority and immediate fallback.

MuffinKid23:
Updated in 93e44db87d with the manually tested camera tuning and 2-7 successful scene-click minimap cadence. Pitch now includes intermediate inclinations (210-325 legacy units), without the previous overhead angle, and vertical adjustment continues when yaw is already aligned. Full local :client:runUnitTests suite passed with English JVM locale, including ClientThreadGuardrailTest; no baseline changes or test suppressions. GitHub Build is currently running. Only the two walker implementation files are included.

MuffinKid23:
Added accessible-bank handoff, the read-only Rs2InteractionApproach helper, and bounded choice among the last three eligible scene tiles. Existing equal-cost pathfinder tie-breaking is unchanged. Full local suite initially reported only the guardrail failure; after using the camera visibility helper and preserving existing lambda identities, the guardrail, bank, and new approach tests pass on the final commit. No baseline changes or suppressions. Companion Mahogany Homes draft: https://github.com/chsami/Microbot-Hub/pull/552 (depends on publication of this client API). GitHub CI should validate the final revision.

MuffinKid23:
Updated with the independent Use Portal Nexus option (enabled by default), POH transport filtering, and transport-cache invalidation. Focused config/cache tests and ClientThreadGuardrailTest passed locally. POH detection crash handling is kept in a separate branch/PR and is not bundled into this update.

chsami:
Pushed two localized fixes for the existing review findings: the main moving-checkpoint branch now uses the same cooldown/expiry policy as the start-of-pass check, and direct/fallback scene clicks reuse a viewport-validated canvas point captured on the client thread. Removed two obsolete guardrail exemptions after moving the reads; no new exemptions were added. Added regression coverage for the 900 ms checkpoint boundary and viewport edges, and documented the movement invariant. Client compilation and the full local unit suite pass (1770 tests, 2 skipped, zero failures/errors). Waiting for fresh expected CI before merge.

chsami:
Final review tightened the same checkpoint guard for a player who stops inside the preclick distance: keep any still-active interim until the shared close/expiry policy permits clearing it. This preserves the existing far/stationary stale-target recovery. Client compilation, all 185 walker unit tests, and the client-thread guardrail pass after the one-line adjustment. The preceding head passed full local tests and CI; waiting for CI on this final head before merging.

PR #1868, opened 2026-09-19
Fix bank cache persistence and refresh

infuse21:
Restores profile-specific bank snapshots across client restarts and updates saved contents from live bank container events. Bank openings update the last-opened timestamp, and route planning can use a restored snapshot while banking actions still require a fresh live update.

Cache reconstruction uses saved values without dispatching to or waiting for the client thread. Snapshot publication remains synchronized, with live-update counters advancing only after the item list is published.

Validation:
- Client compilation passed.
- Bank cache and banked-transport planning tests passed.
- Client-thread and queryable guardrails passed with unchanged baselines.

This PR contains only the banking-cache fix in one commit, based directly on development.


infuse21:
addressed code rabbit findings @coderabbitai review

PR #1872, opened 2026-09-20
Title: Fix looting selection and validate ground-item pickup

infuse21:
Description:
Fix looting paths that could use stale selections, report success without confirming progress, or leave scripts paused after an exception.
- Select fresh ground items when looting executes, with consistent filters, stack-value bounds, and reserved inventory slots.
- Route legacy and tile-item pickup through shared validation of item identity, world view, action, and click bounds.
- Reject unresolved actions and active item/spell selections instead of silently changing Take into another interaction.
- Wait for menu preparation and acknowledgement, clear pending actions on failure, and distinguish confirmed collection from ground-item changes.
- Preserve manual pause state using a scoped looting gate.
Validation: Compilation and focused regression checks passed. Live testing on the upstream-based branch confirmed successful pickup, restored inventory quantity, and pending-menu cleanup.


infuse21:
@coderabbitai review

